PR #5, opened 2021-01-18
Removes aws sdk

willfish:
### Jira link

HOTT-<TODO>

### What?

- [x] Removes unused gems

### Why?

This gem brings in a ridiculous number of libraries 99% of which we
don't use.

AWS migrated to separate gems for each of their apis and the one we're
using for sending mail is ses

See https://github.com/aws/aws-sdk-rails

- I'll validate that this doesn't break sending forum emails in the frontend


mafonso (approved):
LGTM

PR #9, opened 2021-01-19
Add GA server side tracking service

tavy88:
### Jira link

HOTT-262

### What?

Using the GA Measurements Protocol,
this service allows server-side event tracking.

The advantage of this approach is that JS can be
disabled on the device and still work.

### Why?

We want to be able to start various events to understand our users' behaviour better.

Resources: https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide

### DIFF
--- a/app/helpers/ga_tracking_helper.rb
+++ b/app/helpers/ga_tracking_helper.rb
@@ -1,0 +1,31 @@
+module GaTrackingHelper
+  def track_event(event_key, event_value, scope = 'events')
+    event_label = I18n.t(event_key, scope: scope)
+
+    track_events(
+      [
+        {
+          key: event_key,
+          label: event_label,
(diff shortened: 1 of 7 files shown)


app/helpers/ga_tracking_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  private
+
+  def client_tracking_data
+    {
+      ga_cookie: cookies['_ga']&.gsub(/^(.*?\..*?\.)/, ''),
+      ip_address: request.remote_ip,
+      user_agent: request.env['HTTP_USER_AGENT']
+    }
```
tavy88, line 29: This information is needed so we can correlate the events correctly in GA and know where they are coming from.

PR #13, opened 2021-01-25
Remove useless blocking condition

tavy88:
### Jira link

HOTT-NA

### What?

Files should be looked for inside the S3 bucket on production.

### Why?

If we want to enable the CDS flag, without this change the importer would never check the S3 bucket, where the increments are being downloaded.


willfish (approved):
Talked about this previously. Makes total sense. Well, it never made sense that they did this in the first place :facepalm:

PR #17, opened 2021-01-26
Extend the list of licensed quotas

tavy88:
### Jira link

HOTT-368

### What?

Extend the list of licensed quotas

### Why?

Since HMRC wants to start sending 07* in as prefixes
for licensed quotas, we better add all possible
cases.


willfish (approved):
LGTM

PR #23, opened 2021-01-27
[HOTT-359] Adds Clearable concern to all Sequel::Model classes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-359

### What?

If a client of a model wants to clear the association cache there is
currently no public api that enables this behaviour.

We typically want to clear it in the tests because we run the same
queries (at least as they appear to Sequel) multiple times and expect different results to be returned from the cache.

This has been specifically highlighted in the tests around the
TimeMachine which doesn't update the cache key and therefore looks like
the same query.

We can get around this by adding behaviour to Sequel::Model classes
which clears the cache after every example and by making sure that we
don't have multiple TimeMachine expectations in the same example (see
subsequent commit for this change).

### Why?

I am doing this because:

- Our TimeMachine implementation doesn't affect the Sequel::Associations::AssociationReflection cache key and therefore looks like the same query.
- This is a precursor to proving/removing the reload behaviour from all files which has been added because of artefacts of our testing approach mixed with the TimeMachine idiosyncracies.

### DIFF
--- a/app/models/concerns/clearable.rb
+++ b/app/models/concerns/clearable.rb
@@ -1,0 +1,13 @@
+module Clearable
+  extend ActiveSupport::Concern
+
+  included do
+    def self.clear_association_cache
+      association_reflections.map do |_association, association_state|
+        association_state[:cache] = {} if association_state[:cache].present?
+      end
+    end
+  end
+end
+
+Sequel::Model.include Clearable
--- a/spec/models/concerns/clearable_spec.rb
+++ b/spec/models/concerns/clearable_spec.rb
@@ -1,0 +1,31 @@
+require 'rails_helper'
+
+RSpec.describe Clearable do
+  let(:mocked_model) do
+    Struct.new('Commodity') do
+      include Clearable
+
+      def self.association_reflections
+        @association_reflections ||=
+          begin
+            {
+              foo: { cache: { rows: ['some stuff'] } },
+              bar: { cache: {} },
+              baz: { flibble: { some_option: true } },
+
+            }
+          end
+      end
+    end
+  end
+
+  it 'clears association reflection caches' do
+    mocked_model.clear_association_cache
+
+    expect(mocked_model.association_reflections).to eq(
+      foo: { cache: {} },
+      bar: { cache: {} },
+      baz: { flibble: { some_option: true } },
+    )
+  end
+end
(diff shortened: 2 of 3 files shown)


app/models/concerns/clearable.rb
```diff
@@ -0,0 +1,13 @@
+module Clearable
+  extend ActiveSupport::Concern
+
+  included do
+    def self.clear_association_cache
+      association_reflections.map do |_association, association_state|
+        association_state[:cache] = {} if association_state[:cache].present?
+      end
```
tavy88, line 8: 👍

spec/models/concerns/clearable_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      def self.association_reflections
+        @association_reflections ||=
+          begin
+            {
+              foo: { cache: { rows: ['some stuff'] } },
+              bar: { cache: {} },
+              baz: { flibble: { some_option: true } },
+
+            }
+          end
```
tavy88, line 17: Do we need `begin..end` here?

spec/models/concerns/clearable_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+  end
+
+  it 'clears association reflection caches' do
+    mocked_model.clear_association_cache
+
+    expect(mocked_model.association_reflections).to eq(
+      foo: { cache: {} },
+      bar: { cache: {} },
+      baz: { flibble: { some_option: true } },
```
tavy88, line 28: Rubocop is weird here when asking comma after the last key.

spec/rails_helper.rb
```diff
@@ -55,5 +55,17 @@
...
     Sidekiq::Worker.clear_all
+
+    clearable_models = [
+      Certificate,
+      Chief::Tamf,
+      ExportRefundNomenclature,
+      Footnote,
+      GoodsNomenclature,
+      QuotaOrderNumber,
+    ]
```
tavy88, line 66: Nice!

tavy88 (approved):

spec/models/concerns/clearable_spec.rb
willfish: I could move to instantiating the variable outside of the getter and remove the begin/end, here.

spec/models/concerns/clearable_spec.rb
willfish: Yeah, I reckon we could globally change this and autofix

PR #26, opened 2021-01-27
[HOTT-359] Removes reload: true from associations

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-359

### What?

I have added/removed/altered:

- [x] Removes `reload: true` from association methods
- [x] Removes eager_block override to remove false negative issues with the TimeMachine functionality

### Why?

I am doing this because:

- These shouldn't be required by the production application and are causing significant slowness in our api

### DIFF
--- a/app/models/footnote.rb
+++ b/app/models/footnote.rb
@@ -13,14 +13,13 @@ class Footnote < Sequel::Model
                                                      footnote_id],
                                        right_primary_key: %i[footnote_description_period_sid
                                                              footnote_type_id
-                                                             footnote_id],
-                                       eager_block: (->(ds) { ds }) do |ds|
+                                                             footnote_id] do |ds|
     ds.with_actual(FootnoteDescriptionPeriod)
       .order(Sequel.desc(:footnote_description_periods__validity_start_date))
   end
 
   def footnote_description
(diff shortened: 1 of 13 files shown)


app/models/footnote.rb
```diff
@@ -13,14 +13,13 @@ class Footnote < Sequel::Model
                                                      footnote_id],
                                        right_primary_key: %i[footnote_description_period_sid
                                                              footnote_type_id
-                                                             footnote_id],
-                                       eager_block: (->(ds) { ds }) do |ds|
```
willfish, line 17: When `reload: true` basically turns off eager loading and the corresponding eager_block. When we removed the reload we ended up calling the eager_block instead of the relation block on line 18/19 below. This was breaking some tests so this change fixes those expectations.

willfish: I don't know why we'd want to execute something different when eager loading (for example here we were skipping the TimeMachine and the ordering functionality.

willfish: Probably worth a review

willfish: I suspect this might account for a decent number of the forced reloads/presumed TimeMachine issues

PR #30, opened 2021-02-03
Adjust ETL logic as follows

tavy88:
### Jira link

HOTT-416
### What?

- when we have Measures coming through, we delete all
corresponding geographical areas exclusions
- if the XML increment file contains any exclusions they
will be recreated

### Why?

CDS no longer sends exclusions with operation type D, instead, the records will be completely missing.

### DIFF
--- a/lib/cds_importer/entity_mapper.rb
+++ b/lib/cds_importer/entity_mapper.rb
@@ -62,7 +62,7 @@ class CdsImporter
     def remove_excluded_geographical_areas!
       return if xml_node['sid'].blank?
 
-      MeasureExcludedGeographicalArea.where(measure_sid: xml_node['sid']).delete
+      MeasureExcludedGeographicalArea.operation_klass.where(measure_sid: xml_node['sid']).delete
     end
 
     def transform!
--- a/lib/cds_importer/entity_mapper.rb
+++ b/lib/cds_importer/entity_mapper.rb
@@ -21,6 +21,8 @@ class CdsImporter
                            .sort_by { |m| m.mapping_path ? m.mapping_path.length : 0 }
 
       mappers.each do |mapper|
+        remove_excluded_geographical_areas! if mapper == CdsImporter::EntityMapper::MeasureMapper
+
         transform! if mapper == CdsImporter::EntityMapper::GeographicalAreaMembershipMapper
 
         instances = mapper.new(xml_node).parse
@@ -57,8 +59,14 @@ class CdsImporter
       nil
     end
 
+    def remove_excluded_geographical_areas!
+      return if xml_node['sid'].blank?
+
+      MeasureExcludedGeographicalArea.where(measure_sid: xml_node['sid']).delete
+    end
+
     def transform!
-      return unless xml_node.has_key?('geographicalAreaMembership')
+      return unless xml_node.key?('geographicalAreaMembership')
 
       mutate_geographical_area_membership_node!
     end
@@ -67,8 +75,7 @@ class CdsImporter
       convert_single_geo_area_member_to_array!
 
       xml_node['geographicalAreaMembership'] = xml_node['geographicalAreaMembership'].each_with_object([]) do |geographical_area_membership, array|
-
-        next unless geographical_area_membership.has_key?('geographicalAreaGroupSid')
+        next unless geographical_area_membership.key?('geographicalAreaGroupSid')
 
         geographical_area = GeographicalArea[hjid: geographical_area_membership['geographicalAreaGroupSid']]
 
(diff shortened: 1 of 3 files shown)


lib/sequel/plugins/oplog.rb
```diff
@@ -127,10 +127,6 @@ def update(*_attr)
         def insert
           raise NotImplementedError.new('You should be inserting model instances.')
         end
-
-        def delete
-          raise NotImplementedError.new('You should be *destroying* model instances.')
-        end
```
tavy88, line 133: We need unfortunately to do hard deletes.

willfish: Could we use the same hard delete mechanism of the rollbacks rather than removing this protective layer?

lib/cds_importer/entity_mapper.rb
```diff
@@ -57,8 +59,14 @@ def save_record(record)
       nil
     end
 
+    def remove_excluded_geographical_areas!
+      return if xml_node['sid'].blank?
```
willfish, line 63: Is this defensive or have you actually seen a blank sid for the Measure?

willfish: Just curious

lib/cds_importer/entity_mapper.rb
```diff
@@ -57,8 +59,14 @@ def save_record(record)
...
+
+      MeasureExcludedGeographicalArea.where(measure_sid: xml_node['sid']).delete
```
willfish, line 65: So I think that you can do

```suggestion
      MeasureExcludedGeographicalArea.operation_klass.where(measure_sid: xml_node['sid']).delete
```

And not need the change to the overridden delete method in the oplog plugin.

spec/integration/cds_importer/entity_mapper_spec.rb
```diff
@@ -90,5 +92,72 @@
       mapper.import
       expect(AdditionalCode.last.filename).to eq 'test.gzip'
     end
+
+    context 'when measureExcludedGeographicalArea changes are present' do
```
willfish, line 96: It think we're missing one spec for deletion

```ruby
context 'when there are existing exclusions for this measure' do
  it 'does hard delete the exclusions' do
  end
end
```

spec/integration/cds_importer/entity_mapper_spec.rb
```diff
@@ -90,5 +92,72 @@
...
+
+      it 'does recreate the excluded geographical areas contained within the XML increment' do
+        create(:geographical_area, geographical_area_sid: '439')
+
+        expect {
+          mapper.import
+        }.to change(MeasureExcludedGeographicalArea, :count).from(0).to(1)
+      end
+
+      it 'does persists the correct excluded geographical area from the XML increment' do
```
willfish, line 150:
```suggestion
      it 'does persist the correct excluded geographical area from the XML increment' do
```

willfish:
Looks good man. I've walked the code and am happy with it, really. I guess our options re recreating the list of exclusions everytime or doing something else are minimal?

Like the only other alternative to this is to make it so that the CDS api doesn't completely change the mechanism for exclusions? Which I guess they're just not going to do

lib/sequel/plugins/oplog.rb
tavy88: yep!

spec/integration/cds_importer/entity_mapper_spec.rb
tavy88: yep

lib/cds_importer/entity_mapper.rb
tavy88: exactly right

spec/integration/cds_importer/entity_mapper_spec.rb
```diff
@@ -90,5 +92,84 @@
...
+      it 'does not remove excluded geographical areas that belong to measures not present within the XML increment' do
+        create(:geographical_area, geographical_area_sid: '439')
+        other_exclusion = create(:measure_excluded_geographical_area)
+
+        mapper.import
+
+        expect(MeasureExcludedGeographicalArea[measure_sid: other_exclusion.measure_sid]).to be_present
+      end
+
+      context 'when there is an existing exclusion for this measure' do # rubocop:disable RSpec/NestedGroups
```
willfish, line 142: Should we chuck this in the config file? It would be significant amount of rearchitecting to correct tis nested groups violation and I can read your tests fine.

tavy88: Hmm, we could do that.

willfish (approved):

PR #31, opened 2021-02-05
Force ttl of headings, chemical search and commodities to 24 hours

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Fix headings, chemical search and commodities in v2 to be a static 24 hours
- [x] Set the schedule for the recache to be after the import

### Why?

I am doing this because:

- We invalidate the redis cache on a schedule way before this time so this is completely redundant.
- The logic and naming of this method were just wrong, too.

- The combination of performance enhancements and the fact that these caches get invalidated every night anyway diminish the risk around this change

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -25,7 +25,7 @@
     cron: "0 14 * * 6" # 14:00 every Saturday
     description: "TaricSequenceCheckWorker will run every Saturday at 14:00."
   ClearCacheWorker:
-    cron: "0 0 * * *" # 00:00 every day
+    cron: "30 2 * * *" # 02:30 every day
     description: "Clear Rails cache at midnight"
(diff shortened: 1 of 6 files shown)


config/sidekiq.yml
```diff
@@ -25,7 +25,7 @@
     cron: "0 14 * * 6" # 14:00 every Saturday
     description: "TaricSequenceCheckWorker will run every Saturday at 14:00."
   ClearCacheWorker:
-    cron: "0 0 * * *" # 00:00 every day
+    cron: "30 2 * * *" # 00:00 every day
```
tavy88, line 28: Can you update the description as well, please?

tavy88 (approved):
:shipit:

PR #33, opened 2021-02-05
[HOTT-351] Warn on guard clauses in cds importer flow

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-351

### What?

I have added/removed/altered:

- [x] Adds warnings about missing references to email report that comes in after a successful apply

### Why?

I am doing this because:

- We want to have some way of knowing when we've skipped a transform or
exclusion clobbering due to missing references.
- This should be super rare but we want to know about it because it could
indicate data issues upstream

### DIFF
--- a/lib/cds_importer/entity_mapper.rb
+++ b/lib/cds_importer/entity_mapper.rb
@@ -60,7 +60,13 @@ class CdsImporter
     end
 
     def remove_excluded_geographical_areas!
-      return if xml_node['sid'].blank?
+      if xml_node['sid'].blank?
+        message = 'Skipping removal of measure geographical exclusions due to missing measure sid.'
+
+        instrument_warning(message, xml_node)
+
+        return
+      end
 
       MeasureExcludedGeographicalArea.operation_klass.where(measure_sid: xml_node['sid']).delete
     end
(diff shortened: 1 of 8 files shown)


lib/cds_importer/entity_mapper.rb
```diff
@@ -75,7 +81,12 @@ def mutate_geographical_area_membership_node!
       convert_single_geo_area_member_to_array!
 
       xml_node['geographicalAreaMembership'] = xml_node['geographicalAreaMembership'].each_with_object([]) do |geographical_area_membership, array|
-        next unless geographical_area_membership.key?('geographicalAreaGroupSid')
+        unless geographical_area_membership.key?('geographicalAreaGroupSid')
+          message = "Skipping membership import due to missing geographical area sid\n\n#{JSON.pretty_generate(geographical_membership)}\n"
+
+          instrument_warning(message, xml_node)
```
tavy88, line 87: Is it possible to have some unit tests here to check that the right message and the right xml_node is being passed?

lib/cds_importer/entity_mapper.rb
```diff
@@ -60,7 +60,13 @@ def save_record(record)
     end
 
     def remove_excluded_geographical_areas!
-      return if xml_node['sid'].blank?
+      if xml_node['sid'].blank?
+        message = 'Skipping removal of measure geographical exclusions due to missing measure sid.'
+
+        instrument_warning(message, xml_node)
```
tavy88, line 66: Is it possible to have some unit tests here to check that the right message and the right xml_node is being passed?

lib/tariff_synchronizer/mailer.rb
```diff
@@ -54,9 +54,10 @@ def file_write_error(path, reason)
       mail subject: "#{subject_prefix(:error)} Update fetch failed: cannot write update file to file system"
     end
 
-    def applied(update_names, conformance_errors = [])
+    def applied(update_names, conformance_errors, import_warnings)
```
tavy88, line 57: Hopefully `applied` is not being called anywhere else with just the first argument? Wonder why they had to use a default for the second argument?

willfish: This is literally just for the importers (hence the apply/applied naming). I've made sure both pathways conform to the new signature

tavy88 (approved):

PR #36, opened 2021-02-09
[HOTT-426] Clears association cache on all known problem models before a request

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-426

### What?

I have added/removed/altered:

- [x] Clear the class level Sequel association cache of known problem models before each request

### Why?

We know that association queries get cached and don't recognise that the
time machine should represent a change in the stored query.

Weirdly this affects only a subset of models that are invoked with the
TimeMachine.

I'd need to further dig into the specifics to understand why this is but
I know for a fact that the class level cache needs invalidating when we
want to change the actual date.

It's important to know that data gets stored in lexical scope under an
instance of a Sequel query result not in the class cache that we're
clearing here. What we're clearing is explicitly the query itself which
is reused between runs to (I suspect) save Sequel a smidge of time in
generating the output SQL as driven by invoking the sequel model and
dataset methods.

I am doing this because:

- We know the class level association cache doesn't recognise TimeMachine changing the cache key and needs clearing
- We don't want to revert to using reload and forcing our application to grind to a halt/propagate timeout exceptions and introduce lots of n+1s

We can revert this change very quickly and it is non-data-affecting - well it should reduce the issue of the association cache causing increasing data malforms.

### DIFF
--- a/lib/trade_tariff_backend.rb
+++ b/lib/trade_tariff_backend.rb
@@ -169,6 +169,19 @@ module TradeTariffBackend
       [Heading, Certificate, AdditionalCode, Footnote]
     end
 
+    def clearable_models
+      [
+        Certificate,
+        Chief::Tamf,
+        ExportRefundNomenclature,
+        Footnote,
+        GeographicalArea,
+        GoodsNomenclature,
+        Measure,
+        QuotaOrderNumber,
+      ]
+    end
+
     def search_indexes
       indexed_models.map { |model|
         "::Search::#{model}Index".constantize.new(search_namespace)
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -58,17 +58,6 @@ RSpec.configure do |config|
     Rails.cache.clear
     Sidekiq::Worker.clear_all
 
-    clearable_models = [
-      Certificate,
-      Chief::Tamf,
-      ExportRefundNomenclature,
-      Footnote,
-      GeographicalArea,
-      GoodsNomenclature,
-      Measure,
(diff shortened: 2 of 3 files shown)


lib/trade_tariff_backend.rb
```diff
@@ -169,6 +169,19 @@ def cached_models
       [Heading, Certificate, AdditionalCode, Footnote]
     end
 
+    def clearable_models
```
willfish, line 172: At some stage - not now - I should clearly understand the mechanical differences between all of the other models that don't need their caches clearing and the ones that do. Mainly because we may have missed the more obscure models that need clearing and haven't been identified by the tests and because if we understand exactly what the difference is we can document and implement the TimeMachine in a way that means that Sequel recognises it needs to update it's SQL when the actual date is different from the one used to generate the previously cached SQL.

spec/rails_helper.rb
```diff
@@ -58,17 +58,6 @@
...
-      ExportRefundNomenclature,
-      Footnote,
-      GeographicalArea,
-      GoodsNomenclature,
-      Measure,
-      QuotaOrderNumber,
-    ]
-
-    clearable_models.map(&:clear_association_cache)
+    TradeTariffBackend.clearable_models.map(&:clear_association_cache)
```
tavy88, line 61: So they already had this approach on the specs.

tavy88 (approved):
In the light of the current events, I'll approve. It just feels wrong that we have to do this cache clear every time. We need to review this implementation.

PR #40, opened 2021-02-23
[HOTT-473] Adds filter for geographical_area_id

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-473

### What?

I have added/removed/altered:

- [x] Extracts commodity caching and loading into a service
- [x] Adds a filter to the commodity loading that only loads measures for a given geographical area (defaulting to no filter)

### Why?

I am doing this because:

- We want the backend to do the heavy loading when filtering relevant measures so that the duty calculator can just get the measures that it's interested in.
- This will also mean we can automatically get much smaller fixtures


app/services/cached_commodity_service.rb
```diff
@@ -115,10 +118,29 @@ def options
   end
 
   def measures
-    @commodity.measures_dataset.eager(*MEASURES_EAGER_LOAD_GRAPH).all
+    measures = commodity.measures_dataset.eager(*MEASURES_EAGER_LOAD_GRAPH)
```
willfish, line 121: Condense me

app/models/measure.rb
```diff
@@ -410,6 +410,14 @@ def quota_definition_or_nil
     end
   end
 
+  def relevant_for_country?(country_id)
+    return false if measure_excluded_geographical_areas.map(&:excluded_geographical_area).include?(country_id)
+    return true if geographical_area_id == GeographicalArea::ERGA_OMNES_ID && national?
+    return true if geographical_area_id.blank? || geographical_area_id == country_id
+
```
tavy88, line 417: I know you had had enough of this. I am happy if you refactor this in the next PR, but ideally, we should return true only once and link those two conditions with OR. I do appreciate there probably is a need for a bit of refactoring for these conditions.

app/services/cached_commodity_service.rb
```diff
@@ -0,0 +1,148 @@
...
+  attr_reader :commodity, :actual_date, :filter_params
+
+  def presented_commodity
+    Api::V2::Commodities::CommodityPresenter.new(commodity, presented_measures)
+  end
+
+  def presented_measures
+    # TODO: This should be Api::V2::Measures::MeasurePresenter. It works currently because the real presenter used by the CommodityPresenter uses the correct V2 measures presenter.
+    #       Also, why are we validating in a presenter?
+    MeasurePresenter.new(measures, commodity).validate!
```
tavy88, line 112: 👍

app/services/cached_commodity_service.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  def measures
+    measures = commodity.measures_dataset.eager(*MEASURES_EAGER_LOAD_GRAPH).all
+    if filter_by_country_id? && filtering_country.present?
+      measures.select do |measure|
+        measure.relevant_for_country?(filtering_country.geographical_area_id)
+      end
+    else
+      measures
+    end
```
tavy88, line 130: This could be in a separate method maybe?
```ruby
def apply_filter(measures, filtering_country)
return measures unless filtering_country.present? || filter_by_country_id?

measures.select do |measure|
    measure.relevant_for_country?(filtering_country.geographical_area_id)
  end
end 
```

tavy88 (approved):
🚢 🇮🇹

PR #43, opened 2021-02-26
[HOTT-356] Decorate zero duties for third countries on commodity response

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-356

### What?

I have added/removed/altered:

- [x] Adds `zero_mfn_duty` key to metadata of commodity response
- [x] Fixes broken bitzesty logic

### Why?

I am doing this because:

- We need this question answered for the duty calculator flow

### DIFF
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -392,7 +392,7 @@ class Measure < Sequel::Model
   def zero_mfn?
     third_country? &&
       measure_components.count == 1 &&
-      measure_components.first.zero_duty
+      measure_components.first.zero_duty?
   end
 
   def order_number
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -389,6 +389,12 @@ class Measure < Sequel::Model
     measure_components.any?(&:meursing?)
   end
 
+  def zero_mfn?
+    third_country? &&
+      measure_components.count == 1 &&
+      measure_components.any?(&:zero_duty?)
+  end
+
   def order_number
     if quota_order_number.present?
       quota_order_number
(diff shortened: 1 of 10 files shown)


app/models/measure.rb
```diff
@@ -389,6 +389,12 @@ def meursing?
     measure_components.any?(&:meursing?)
   end
 
+  def zero_mfn?
```
tavy88, line 392: What does mfn stand for?

app/models/measure.rb
```diff
@@ -389,6 +389,12 @@ def meursing?
...
+    third_country? &&
+      measure_components.count == 1 &&
+      measure_components.any?(&:zero_duty?)
```
tavy88, line 395: Hmm because the measure_components has to be an array of one element Line#394, can we not just say:

```ruby
measure_components.first.zero_duty?
```

tavy88 (approved):

app/models/measure.rb
willfish: We could

willfish, line 395:
```suggestion
      measure_components.first.zero_duty
```

app/models/measure.rb
willfish: most favoured nation

app/models/measure.rb
```diff
@@ -389,6 +389,12 @@ def meursing?
     measure_components.any?(&:meursing?)
   end
 
+  def zero_mfn?
+    third_country? &&
+      measure_components.count == 1 &&
+      measure_components.first.zero_duty
```
willfish, line 395:
```suggestion
      measure_components.first.zero_duty?
```

PR #44, opened 2021-02-26
[HOTT-355] Surface trade remedies in commodity response meta

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-355

### What?

I have added/removed/altered:

- [x] Adds whether there are any trade remedies in place to the commodity response

### Why?

I am doing this because:

- This is needed to make decisions in the duty calculator

### DIFF
--- a/app/presenters/api/v2/commodities/commodity_presenter.rb
+++ b/app/presenters/api/v2/commodities/commodity_presenter.rb
@@ -45,7 +45,7 @@ module Api
         end
 
         def meursing_code?
-          import_measures.any?(&:meursing?) || export_measures.any?(&:meursing?)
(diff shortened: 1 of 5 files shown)


app/presenters/api/v2/commodities/commodity_presenter.rb
```diff
@@ -45,7 +45,7 @@ def ancestor_ids
         end
 
         def meursing_code?
-          import_measures.any?(&:meursing?) || export_measures.any?(&:meursing?)
```
willfish, line 48: Meursing measure components only apply to import measure. Reviewed this with @mattlavis-transform

tavy88 (approved):
Nice and simple!

PR #46, opened 2021-03-04
[HOTT-410] Surface measurement units for the duty calculator

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-410

### What?

I have added/removed/altered:

- [x] Surface all of the applicable measure units for the duty calculator
- [x] Adds business logic for determining whether a measure component or measure condition component is an ad valorum component
- [x] Backfills on a missing spec for the zero duty question
- [x] Surface meursing code for a future iteration of the duty calc

### Why?

I am doing this because:

- In order to more simply iterate through a given set of measure units and present them as simplified components in the duty calculator
- To make it possible for any user of the api to be able to use this information

### DIFF
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -432,9 +432,10 @@ class Measure < Sequel::Model
   end
 
   def zero_mfn?
-    third_country? &&
-      measure_components.count == 1 &&
-      measure_components.first.zero_duty?
+    return false unless third_country?
+    return false unless measure_components.count == 1
+
+    measure_components.first.zero_duty?
   end
 
   def expresses_unit?
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -442,36 +442,29 @@ class Measure < Sequel::Model
   end
 
   def ad_valorem?
-    ad_valorem_measure_components? || ad_valorem_measure_conditions?
+    ad_valorem_resource?(:measure_components) || ad_valorem_resource?(:measure_conditions)
   end
 
-  def ad_valorem_measure_components?
-    measure_components.count == 1 &&
-      measure_components.first.ad_valorem?
+  def ad_valorem_resource?(resource)
+    public_send(resource).count == 1 &&
+      public_send(resource).first.ad_valorem?
   end
 
-  def ad_valorem_measure_conditions?
-    measure_conditions.count == 1 &&
-      measure_conditions.first.ad_valorem?
+  def units
+    component_units + condition_units
   end
 
-  def measure_component_units
-    measure_components.map do |component|
-      {
-        'measure_sid' => component.measure_sid,
-        'measurement_unit_code' => component.measurement_unit_code,
-        'measurement_unit_qualifier_code' => component.measurement_unit_qualifier_code,
-      }
+  private
+
+  def component_units
+    measure_components.map(&:values).map do |component|
+      component.values.slice(:measure_sid, :measurement_unit_code, :measurement_unit_qualifier_code)
     end
   end
 
-  def measure_condition_units
-    measure_conditions.map do |condition|
-      {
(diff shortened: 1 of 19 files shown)


app/models/measure_component.rb
```diff
@@ -1,4 +1,6 @@
 class MeasureComponent < Sequel::Model
+  AD_VALORUM_DUTY_EXPRESSION = '%'.freeze
```
willfish, line 2: Remove me

app/models/measure_type.rb
```diff
@@ -24,6 +24,12 @@ class MeasureType < Sequel::Model
     '696', # Additional duties (safeguard)
   ].freeze
 
+  UNIT_EXPRESSABLE_MEASURES = [
```
willfish, line 27: The plan is to expand these to include VAT and Excise measures at a later date.

app/models/measure.rb
```diff
@@ -436,4 +426,61 @@ def self.changes_for(depth = 1, conditions = {})
...
+    ad_valorem_resource?(:measure_components) || ad_valorem_resource?(:measure_conditions)
+  end
+
+  def units
+    component_units + condition_units
+  end
+
+  private
+
+  def ad_valorem_resource?(resource)
```
willfish, line 454: Revert public send - this is actually clearer/nice with a smidge of duplication

app/models/measure.rb
```diff
@@ -436,4 +426,61 @@ def self.changes_for(depth = 1, conditions = {})
...
   end
+
+  def meursing?
+    measure_components.any?(&:meursing?)
+  end
+
+  def zero_mfn?
+    third_country? &&
+      measure_components.count == 1 &&
+      measure_components.first.zero_duty?
```
tavy88, line 437: Wow-what a crazy condition...

app/models/measure.rb
```diff
@@ -436,4 +426,61 @@ def self.changes_for(depth = 1, conditions = {})
...
+  end
```
tavy88, line 438:
```suggestion
def zero_mfn?
    return false if measure_components.count > 1
    
    third_country? && measure_components.first.zero_duty?
  end
``` ?

File: app/models/measure.rb [hunk 6]
```diff
@@ -436,4 +426,61 @@ def self.changes_for(depth = 1, conditions = {})
...
+
+  def expresses_unit?
+    measure_type.expresses_unit? && !ad_valorem?
+  end
+
+  def ad_valorem?
+    ad_valorem_resource?(:measure_components) || ad_valorem_resource?(:measure_conditions)
```
tavy88, line 445:
```suggestion
    %i[measure_components measure_conditions].any? { |condition| ad_valorem_resource?(condition)}
```
Thoughts?

app/services/measure_unit_service.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def call
+    units.each_with_object({}) do |unit, acc|
+      unit_key = "#{unit[:measurement_unit_code]}#{unit[:measurement_unit_qualifier_code]}"
+
+      if acc[unit_key].present?
+        acc[unit_key]['measure_sids'].add(unit[:measure_sid])
+
+        next
```
tavy88, line 13: Do you need next here?

willfish: No, since we're at the tail of this iteration - good spot

app/models/measure.rb
willfish: I don't know about this one. I'd probably lean on doing this for 3 or more resources :confused: . Struggling to prefer either implementation, atm

tavy88: 👍

tavy88 (approved):
:shipit:

PR #47, opened 2021-03-08
[HOTT-503] Add CircleCI config 

mafonso:
### Jira link

HOTT-503

### What?

I have added/removed/altered:

- [ ] Setup CircleCI 
- [ ] Run tests in CI

### Why?

I am doing this because:

- We are porting CI from GitHub Actions to CircleCI


.circleci/config.yml
```diff
@@ -0,0 +1,124 @@
...
+
+jobs:
+  build:
+    docker:
+      - image: cimg/ruby:2.7-node
+      - image: redis
+    resource_class: medium
+    steps:
+      - checkout
+      - ruby/install-deps
```
willfish, line 18: Presume this automatically caches?

mafonso: It does

https://circleci.com/developer/orbs/orb/circleci/ruby#commands-install-deps

.circleci/config.yml
```diff
@@ -0,0 +1,124 @@
...
+          DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/tariff_test"
+      - image: circleci/postgres:10-alpine
+        environment:
+          POSTGRES_USER: postgres
+          POSTGRES_DB: tariff_test
+          POSTGRES_PASSWORD: postgres
+      - image: circleci/redis:4.0.9
+        environment:
+          - REDIS_URL: "redis://localhost:6379/"
+      - image: docker.elastic.co/elasticsearch/elasticsearch:7.9.3
```
willfish, line 49: A big part of me feels like it's outrageous not to mock redis and es in the tests :man_facepalming:

.circleci/config.yml
```diff
@@ -0,0 +1,124 @@
...
+          - http.port: 9200
+          - discovery.type: single-node
+    resource_class: medium
+    steps:
+      - checkout
+      - run:
+          name: "Install OS packages"
+          command: |
+           sudo apt-get update
+           sudo apt-get install libcurl4-openssl-dev postgresql-client
```
willfish, line 64: That we need libcurl is also a bit insane. We don't benefit from the curl api anymore than we do from just using plain ruby core network libs

.circleci/config.yml
```diff
@@ -0,0 +1,124 @@
...
+      - ruby/install-deps
+      - run:
+          name: "Setup DB"
+          command: |
+            RAILS_ENV=test bundle exec rake db:structure:load --trace
+      - run:
+          name: "Check ElasticSearch for availability"
+          command: |
+            curl "http://localhost:9200/_cat/health"
+#      - ruby/rspec-test
```
willfish, line 74: Can remove?

.circleci/config.yml
```diff
@@ -0,0 +1,124 @@
...
+      - run:
+          name: RSpec Tests
+          command: |
+            mkdir -p "/tmp/test-results/rspec"
+            TESTFILES=$(circleci tests glob "spec/**/*_spec.rb" | circleci tests split --split-by=timings)
```
willfish, line 79: So awesome

.circleci/config.yml
```diff
@@ -0,0 +1,124 @@
...
+          context: trade-tariff
+      - build:
+          context: trade-tariff
+          requires:
+            - test
+#          filters:
+#            branches:
+#              only:
+#                - master
+# TODO: Discuss production deployment strategy.
```
willfish, line 124: I think merging to master should automatically deploy to staging

.circleci/config.yml
```diff
@@ -0,0 +1,124 @@
...
+workflows:
+  version: 2
+  build_and_test:
+    jobs:
+# TODO: Fix rubocop issues
+#      - checking:
+#          context: trade-tariff
+      - test:
+          context: trade-tariff
+      - build:
```
willfish, line 116: Should we invoke the deploy_dev step at this point?

willfish:
So if the goal is only to get the tests running in circleci I think this is totally fine. I think for consistency we want to make sure we're deploying to dev? I guess that would mean removing the current test workflow file for github actions in the same PR?

.circleci/config.yml
willfish: Nice

.circleci/config.yml
mafonso: Yeah, it feels a bit too heavy.

.circleci/config.yml
mafonso: A gem was not compiling due to the lack of this lib.  I then noticed we were installing it as a pre-step in BZ's workflow.

.circleci/config.yml
mafonso: We can, This is something I want to check with you guys. I'd rather use the orb call
But have to exclude some tests `--tag '~skip_ci'` so I had to unroll that call.

I'll like to keep this and remove the unrolled call if possible.

.circleci/config.yml
mafonso: I'm implementing deploy in a separate PR

.circleci/config.yml
mafonso: Yes, that would be the plan.

willfish (approved):

PR #51, opened 2021-03-11
[HOTT-463] Surface exchange rates for the duty calculator

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-463

### What?

I have added/removed/altered:

- [x] Adds exchange rate lookup service that expires the cache lazily at 15:15 each day.
- [x] Expose the response from this service behind a v2 json:api endpoint

### Why?

I am doing this because:

- This is needed for the duty calculator and (we expect) other things in the backend.

### AC

- Must be encapsulated and portable
- Must refresh when the api data refreshes

### DIFF
--- a/app/services/exchange_rate_service.rb
+++ b/app/services/exchange_rate_service.rb
@@ -1,0 +1,61 @@
+class ExchangeRateService
+  API_URL = 'https://api.exchangeratesapi.io/latest'.freeze
+  EXCHANGE_RATES_UTC_REFRESH_HOUR = 15
+  EXCHANGE_RATES_UTC_REFRESH_MIN  = 15
+  REDIS_KEY = 'trade-tariff-exchange-rates'.freeze
+
+  def call
+    cached_exchange_rates || fetched_exchange_rates
+  end
+
+  private
+
+  def cached_exchange_rates(ignore_expiry: false)
+    response = TradeTariffBackend.redis.get(REDIS_KEY)
+
+    if response.present?
+      exchange_rates = JSON.parse(response)
+      exchange_rates['expires_at'] = Time.zone.parse(exchange_rates['expires_at'])
+
+      return exchange_rates if ignore_expiry
+      return exchange_rates if exchange_rates['expires_at'] > now
+
+      false
+    end
+  end
+
+  def fetched_exchange_rates
+    response = Faraday.new(API_URL).get
+
+    if response.success?
+      exchange_rates = JSON.parse(response.body)
+      exchange_rates['expires_at'] = expires_at.iso8601
+
+      TradeTariffBackend.redis.set(REDIS_KEY, exchange_rates.to_json)
+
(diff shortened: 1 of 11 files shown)


app/services/exchange_rate_service.rb
```diff
@@ -0,0 +1,61 @@
...
+    response = TradeTariffBackend.redis.get(REDIS_KEY)
+
+    if response.present?
+      exchange_rates = JSON.parse(response)
+      exchange_rates['expires_at'] = Time.zone.parse(exchange_rates['expires_at'])
+
+      return exchange_rates if ignore_expiry
+      return exchange_rates if exchange_rates['expires_at'] > now
+
+      false
```
tavy88, line 23: Should this return a boolean? How about nil maybe?

app/services/exchange_rate_service.rb
```diff
@@ -0,0 +1,61 @@
...
+      min: EXCHANGE_RATES_UTC_REFRESH_MIN,
+    )
+  end
+
+  def relative_day
+    if now.hour > EXCHANGE_RATES_UTC_REFRESH_HOUR
+      now.tomorrow
+    else
+      now
+    end
```
tavy88, line 55:
```suggestion
    return now.tomorrow if now.hour > EXCHANGE_RATES_UTC_REFRESH_HOUR
    
    now
```

spec/rails_helper.rb
```diff
@@ -46,10 +46,12 @@
   RedisLockDb.redis = redis
 
   config.before(:suite) do
+    TradeTariffBackend.redis.flushdb
     redis.flushdb
```
tavy88, line 50: Hmm, so whose instance is this?

app/services/exchange_rate_service.rb
willfish: Yeah, good shout

spec/rails_helper.rb
willfish: A separate instance for creating a mutex I believe

tavy88: Have you tried removing it does it break the tests?

willfish: It's using a different redis database from the default (0) one

willfish: Quite a few tests fail if we remove clearing the lock db

app/services/exchange_rate_service.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    if response.present?
+      exchange_rates = JSON.parse(response)
+      exchange_rates['expires_at'] = Time.zone.parse(exchange_rates['expires_at'])
+
+      return exchange_rates if ignore_expiry
+      return exchange_rates if exchange_rates['expires_at'] > now
+    end
+
+    nil
```
willfish, line 24: This is implicit - remove me

tavy88: 👍

tavy88 (approved):
Very clean an nice!

PR #52, opened 2021-03-15
Removes pdf generation from the backend

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes all PDF functionality from the backend

### Why?

I am doing this because:

- We've removed the functionality from the UKTT gem
- We've removed the functionality from the frontend


tavy88 (approved):
This is amazing!!

PR #55, opened 2021-03-18
Use orb command to run tests

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Use the orb command to run rspec tests
- [x] Only run tests on a feature branch and save credits

### Why?

I am doing this because:

- This is simpler and avoids duplication


mafonso (approved):
LGTM

PR #57, opened 2021-03-19
Exclude geographical areas by service

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Adds cached_geographical_area_service and spec
- [x] Exclude geographical areas depending on the service following the same logic as the frontend change we made

### Why?

I am doing this because:

- These areas aren't needed in each of these services.
- This avoids duplication in both of our frontends


tavy88 (approved):

tavy88 (approved):
:shipit:

PR #61, opened 2021-04-06
Fixes ExchangeRateService

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Adds access key query to `ExchangeRateService`
- [x] Move from https -> http

### Why?

I am doing this because:

- We can't pull out exchange rates in the duty calculator
- The api changed to requiring an access key and disabling requests over
https

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,0 +1,1 @@
+EXCHANGE_RATE_ACCESS_KEY=flibble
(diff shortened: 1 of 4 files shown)


tavy88 (approved):
Nice

PR #73, opened 2021-04-27
Add a note about the PostgreSQL version in README

cacquani:
### Jira link

none

### What?

This PR adds a note in the `README.md` file regarding the required versions for PostgreSQL and ElasticSearch.

### Why?

I am doing this because:

During a first setup of the application, I've been informed that the backend only works with PostgreSQL 10, and the error I was experiencing were due to the fact that I was using a more recent version.

Adding the information in the `README.md`  will hopefully make things easier for the next person that needs to set up the application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,10 @@ Other related projects:
 ### Dependencies
 
   - Ruby
-  - Postgresql 10 (won't work with more recent versions)
-  - ElasticSearch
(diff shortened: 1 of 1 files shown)


tavy88 (approved):
Nice!

willfish (approved):

willfish (approved):

PR #74, opened 2021-04-27
Bumps sso gem

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Bumps gds-sso gem to 15.0.1

### Why?

I am doing this because:

- This is needed to keep us up-to-date with the latest patches from the client and signon service


Gemfile
```diff
@@ -10,6 +10,7 @@ ruby '~> 2.7.1'
 # Server
 gem 'puma', '~> 5.0.4'
 gem 'rails', '>= 6.0.3.4'
+gem 'mimemagic'
```
willfish, line 13: This is now completely removed as a dependency from this project. I've added it here to reflect that we still use the mimetype functionality of the mimemagic gem

Gemfile.lock
```diff
@@ -1,61 +1,65 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    actioncable (6.0.3.4)
-      actionpack (= 6.0.3.4)
+    actioncable (6.1.3.1)
```
willfish, line 4: These get updated when you unlock the dependency graph and update the gds-sso gem

app/models/user.rb
```diff
@@ -50,6 +50,12 @@ def disabled?
     disabled
   end
 
+  def update!(options)
```
willfish, line 53: Add a test of me

willfish: This is actually tested by the shared examples

mafonso (approved):
Pair reviewed.

LGTM 

:shipit:

PR #75, opened 2021-04-29
Surface additional codes for the duty calculator

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-581
https://transformuk.atlassian.net/browse/HOTT-658

### What?

I have added/removed/altered:

- [x] Adds additional code service for constructing applicable_additional_code meta data
- [x] Adds plumbing in the presenter to expose additional codes to the serializer
- [x] Adds fixed annotations in a file that we can easily update with more additional codes
- [x] Adds serializer plumbing to pull these out for the filtered list of measures for a given geographical area
- [x] Adds business logic for deciding on the type of the additional code
- [x] Adds specs for all of this

The structure of the surfaced codes is:

```json
{
  "meta": {
    "applicable_additional_codes": {
      "552": {
        "heading": {
          "overlay": "Describe your goods in more detail",
          "hint": "To trade this commodity, you need to specify an additional 4 digits, known as an additional code"
        },
        "additional_codes": [
          {
            "code": "B999",
            "overlay": "Other",
            "hint": ""
          },
          {
            "code": "B349",
            "overlay": "Hunan Hualian China Industry Co., Ltd; Hunan Hualian Ebillion China Industry Co., Ltd; Hunan Liling Hongguanyao China Industry Co., Ltd; Hunan Hualian Yuxiang China Industry Co., Ltd.",
            "hint": ""
          }
        ]
      }
    }
  }
}
```
Here are four interesting variants of additional codes:

https://dev.trade-tariff.service.gov.uk/commodities/4016999710#import
https://dev.trade-tariff.service.gov.uk/commodities/2903898090#import
https://dev.trade-tariff.service.gov.uk/xi/commodities/1007109000#import
https://dev.trade-tariff.service.gov.uk/xi/commodities/6911100090?country=CN#import

### Why?

I am doing this because:

- We want to surface the additional codes that we need to pick in the duty calculator. These choices will affect which measures get filtered down and converted into options in the duty calculator.
- We may need these questions for other projects later down the line and don't want to duplicate these

### DIFF
--- a/app/models/additional_code.rb
+++ b/app/models/additional_code.rb
@@ -1,4 +1,11 @@
 class AdditionalCode < Sequel::Model
+  PREFERENCE_TYPE = 'preference'.freeze
+  REMEDY_TYPE = 'remedy'.freeze
+  UNKNOWN_TYPE = 'unknown'.freeze
+
+  PREFERENCE_TYPE_IDS = %w[2].freeze
+  REMEDY_TYPE_IDS = %w[8 A B C].freeze
+
   plugin :time_machine
   plugin :oplog, primary_key: :additional_code_sid
   plugin :conformance_validator
@@ -38,5 +45,24 @@ class AdditionalCode < Sequel::Model
 
   def id
     additional_code_sid
+  end
+
+  def applicable?
+    type != UNKNOWN_TYPE
+  end
+
+  def type
+    return PREFERENCE_TYPE if additional_code_type_id.in?(PREFERENCE_TYPE_IDS)
+    return REMEDY_TYPE if  additional_code_type_id.in?(REMEDY_TYPE_IDS)
+
+    UNKNOWN_TYPE
+  end
+
+  def self.additional_codes
+    @additional_codes ||=
+      begin
+        file = File.join(::Rails.root, 'db', 'additional_codes.json').freeze
+        JSON.parse(File.read(file))
+      end
   end
 end
--- a/app/presenters/measure_presenter.rb
+++ b/app/presenters/measure_presenter.rb
@@ -30,6 +30,7 @@ class MeasurePresenter
   end
 
   def third_country_duty_dedup
+    # TODO: Fix measures with additional codes that are duplicate
     if @collection.select { |m| m.measure_type_id == THIRD_COUNTRY_DUTY_ID }.size > 1
       @collection.delete_if do |m|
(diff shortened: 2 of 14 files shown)


spec/services/applicable_additional_code_service_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+              },
+            ],
+          },
+        }
+      end
+
+      it { expect(service.call).to include_json(expected_additional_codes) }
+    end
+
+    context 'when the measures do not have additional codes' do
```
willfish, line 86:
```suggestion
    context 'when the measures do not have applicable additional codes' do
```

app/models/additional_code.rb
```diff
@@ -1,4 +1,11 @@
 class AdditionalCode < Sequel::Model
+  PREFERENCE_TYPE = 'preference'.freeze
+  REMEDY_TYPE = 'remedy'.freeze
+  UNKNOWN_TYPE = 'unknown'.freeze
```
tavy88, line 4: How would you feel about `OTHER_TYPE` here?

app/models/additional_code.rb
```diff
@@ -39,4 +46,23 @@ def code
...
+    return PREFERENCE_TYPE if additional_code_type_id.in?(PREFERENCE_TYPE_IDS)
+    return REMEDY_TYPE if  additional_code_type_id.in?(REMEDY_TYPE_IDS)
+
+    UNKNOWN_TYPE
+  end
+
+  def self.additional_codes
+    @additional_codes ||=
+      begin
+        file = File.join(::Rails.root, 'db', 'additional_codes.json').freeze
```
tavy88, line 64: Can't we use an API endpoint here instead? Seems a bit worrying that we have to resort to reading files for this.

willfish: Yeah, this is another way of saying what we were talking about in that Slack Thread. For now, we're the source of truth for overriding annotations, types and headings. This is the same for the annotations around measure units, sadly

tavy88: Yep as we discussed let's keep this for now due to time constraints. But as a second iteration let's add this section being managed from our Admin Application.

tavy88 (approved):

app/presenters/measure_presenter.rb
```diff
@@ -30,6 +30,7 @@ def method_missing(*args, &block)
   end
 
   def third_country_duty_dedup
+    # TODO: Fix measures with additional codes that are duplicate
```
willfish, line 33: [shakes fist]

PR #76, opened 2021-04-30
Deduplicate third country measures with additional codes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-666

### What?

I have added/removed/altered:

1. Preserve the behaviour to deduplicate third country measures where there are more than one in the list of import measures
2. Consolidate the behaviour to unique additional code measures by the measure_sid for those third country measures with additional codes
3. Remove unused measure deduplication in the ApplicableAdditionalCodeService since this is handled in a consolidated manner
4. Updates specs to reflect removal of uniqueness from the service

### Why?

I am doing this because:

- We currently show duplicate measures in the uk backend and need to hide these.

### DIFF
--- a/app/presenters/measure_presenter.rb
+++ b/app/presenters/measure_presenter.rb
@@ -1,5 +1,4 @@
 class MeasurePresenter
-  THIRD_COUNTRY_DUTY_ID = '103'.freeze
   HIDDEN_MEASURE_TYPE_IDS = %w[430 447].freeze
 
   def initialize(collection, declarable)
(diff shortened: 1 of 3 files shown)


app/presenters/measure_presenter.rb
```diff
@@ -29,13 +28,18 @@ def method_missing(*args, &block)
...
   end
 
-  def third_country_duty_dedup
-    # TODO: Fix measures with additional codes that are duplicate
-    if @collection.select { |m| m.measure_type_id == THIRD_COUNTRY_DUTY_ID }.size > 1
-      @collection.delete_if do |m|
-        m.measure_type_id == THIRD_COUNTRY_DUTY_ID && m.additional_code.blank? && m.goods_nomenclature_sid != @declarable.goods_nomenclature_sid
-      end.compact!
-    end
+  def delete_duplicate_third_country_measures
```
willfish, line 31: I've renamed this for consistent naming between the methods

PR #78, opened 2021-05-04
Hott 428 - change displayed id for quota definitions

cacquani:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-428

### What?

I have added/removed/altered:

- [x] Replace the `quota_order_number_sid` with `quota_definition_sid` as surfaced quota definition ID in the JSON API
- [x] Add a spec to make sure the correct value is used for the `definition_id` function in the `QuotaOrderNumber` model
- [x] Fix some minor linting issues in some of the files touched during the investigation

### Why?

I am doing this because:

- The ID currently surfaced for quota definitions is in fact the `quota_order_number_sid`. 
  This makes some sense as the above is the join key, but the correct ID for quota definition is in fact the `quota_definition_sid`

- This changes the ID used to link the order number with the definition in the JSON API. 
  This doesn't seem to be breaking the frontend as the ID is changed consistently, but a thorough test is still required.

### DIFF
--- a/app/models/quota_blocking_period.rb
+++ b/app/models/quota_blocking_period.rb
@@ -1,9 +1,8 @@
 class QuotaBlockingPeriod < Sequel::Model
   plugin :time_machine, period_start_column: :blocking_start_date,
-         period_end_column: :blocking_end_date
+                        period_end_column: :blocking_end_date
   plugin :oplog, primary_key: :quota_definition_sid
   plugin :conformance_validator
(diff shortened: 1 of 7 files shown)


tavy88:
Can you make sure there are o 3rd party services consuming this API v2? Probably worth checking with Matt. Looks good otherwise!

tavy88 (approved):

PR #79, opened 2021-05-06
Add Brakeman checks

mafonso:
### Jira link

HOTT-367

### What?

I have added/removed/altered:

- [ ] .gitignored `vendor` path
- [ ] Added Brakeman gem and CI jobs

### Why?

I am doing this because:

- We want to run brakeman checks.

- vendor path was removed from the repo. Cypress tests and unit tests pass, but it's not clear at this time if this was needed at some point or it just slipped into the repo.


.gitignore
```diff
@@ -1,5 +1,6 @@
 # Ignore bundler config.
 /.bundle
+vendor/*
```
mafonso, line 3: Not clear if this was needed at some point or it just slipped into the repo.
Unit tests and manual cypress tests still pass after this was removed.

Also we had this path under .cfignore, so it never made it to the PaaS, which build everything from scratch.

willfish:
Also, is it possible to fail on a high brakeman issue, only?

mafonso:
> @mafonso Do we want to fix the brakeman issues separately? Or would you prefer to fix them as part of this branch?

There are two tickets on the backlog to address those.  
Since it might require some code refactoring, it may be best to address them on their own stories and keep the scope of this PR just on the CI runs.

mafonso:
> Also, is it possible to fail on a high brakeman issue, only?

Yes, we can choose have `--no-exit-on-warn` option.

.circleci/config.yml
```diff
@@ -134,10 +134,14 @@ jobs:
...
           format: progress
           label: Inspecting with Rubocop
-      - slack/notify:
-          channel: deployments
-          event: fail
-          template: basic_fail_1
+      - run:
+          name: Inspecting with Brakeman
+          when: always
+          command: 'bundle exec brakeman -o test-results/brakeman/brakeman.junit -o brakeman.html --no-progress --separate-models'
```
willfish, line 140:
```suggestion
          command: 'bundle exec brakeman -o test-results/brakeman/brakeman.junit -o brakeman.html --no-progress --separate-models --no-exit-on-warn'
```

willfish (approved):
We're going to fix violations separately

PR #82, opened 2021-05-13
HOTT-659: Search quotas by full date

cacquani:
Search quotas by full date

### Jira link

https://transformuk.atlassian.net/browse/HOTT-659

### What?

I have added/removed/altered:

- [x] alter the logic to search by full date when provided
- [x] add specs
- [x] clean up some of the many linting issues

### Why?

I am doing this because HOTT-659 changes to the quote search form included moving from a plain year field to a full date search, and the related backend service was lacking this option.

### Have you? (optional)

### Deployment risks

No known risk: the service has been extended by adding a new filter by full date functionality, but the old search by years is still in place.

### DIFF
--- a/app/services/quota_search_service.rb
+++ b/app/services/quota_search_service.rb
@@ -1,16 +1,16 @@
 class QuotaSearchService
-  STATUS_VALUES = %w(blocked exhausted not_blocked not_exhausted).freeze
+  STATUS_VALUES = %w[blocked exhausted not_blocked not_exhausted].freeze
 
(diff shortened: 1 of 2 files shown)


app/services/quota_search_service.rb
```diff
@@ -41,6 +42,13 @@ def perform
 
   private
 
+  def extract_date_or_years(attributes)
```
willfish, line 45: Isn't this really extract_year_month_day?

willfish (approved):

PR #87, opened 2021-05-24
Make quotas conform to time machine functionality

willfish:
### Jira link

HOTT-<TODO>

### What?

The quota search service enables searching for quota definitions for measures that are valid according to the current year. We've recently changed the service to use a specific date (e.g. precise to the day)

We noticed an issue with the service when pulling quotas for certain order numbers not returning the related resources.

We previously were using a custom implementation of a date filter that was not propagating to related resources (e.g. quota definitions) because we were using whole year/years as the filter. We were only filtering in related measures based on, say, the current year that was passed and filtering in the definitions that are valid on today's date.

When we moved to a more specific filter of the full date we were able to rely on the actual time machine functionality which would have fixed the broken year filter implementation - but importantly not the year implementation.

This PR fixes this by:

- [ ] Removing the enforced time machine for today filter (which is actually the default, anyway without an as_of query param)
- [ ] Overriding the time machine actual_date to be the date parsed into the query for the given day, month and year coming from the quotas form in the quota search controller
- [ ] Adding unit tests of the quota order service to make sure that it returns related resources using the same date range filter

### Why?

I am doing this because:

- The quota search service is broken in confusing ways

### DIFF
--- a/app/controllers/api/v2/quotas_controller.rb
+++ b/app/controllers/api/v2/quotas_controller.rb
@@ -1,20 +1,17 @@
 module Api
   module V2
     class QuotasController < ApiController
-      before_action :find_quotas
-
       def search
-        options = {}
-        options[:include] = [:quota_order_number, 'quota_order_number.geographical_areas', :measures, 'measures.geographical_area']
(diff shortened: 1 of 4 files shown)


willfish:
Matt is going to define the exact behaviour under different circumstances

PR #88, opened 2021-05-26
Use 2.7.3 and parameterise buildpacks

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Updates ruby to 2.7.3
- [x] Parameterise build pack

### Why?

I am doing this because:

- We need an easier way of updating ruby


mafonso (approved):
LGTM

PR #91, opened 2021-06-01
Surface quota balance events

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-716

### What?

Mainly this PR enables calls to fetch quota balance events with:

```
GET /api/v2/quotas/search?order_number=058063&include=quota_balance_events,measures,measures.geographical_area_id
```

This doesn't affect the existing functionality in any way but does make
some optimisations to what queries get executed based on what's being
passed to the api endpoint.

Some other things that are happening in this PR are:

- Move to lazy loading quota definition relationships (only load them if
  they're specified)
- Enable people to specify the includes in the quotas controller
- Default to the original includes that are required so we don't affect
  existing functionality
- Surface quota balance events and add corresponding serializer and
  specs on a quota definition
- Adds brakeman ignore file and fixes some issues identified by brakeman
  in the quota search service we're updating
- Ignore rubocop complaints about validations without indexes.
- Tidy up some weird api decisions in the quota search service
- Fixes a spec in the search service which hides an actual failing spec
  around encoding (not really encoding, just replacing plusses with
  underscores).

### Why?

I am doing this because:

- This service pulls out the moon when it needs a rock and hasn't been optimised based on the inputs
- We need to surface quota balance events to reduce the load on the frontend UI where people are walking our quota search service
  and incorrectly thinking that walking through past quota definitions yields that day's balance events.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,6 +30,8 @@ RSpec/NestedGroups:
   Enabled: false
 Rails/SaveBang:
   Enabled: false
+Rails/UniqueValidationWithoutIndex:
+  Enabled: false
 Rails/FilePath:
   Enabled: false
 Style/Documentation:
--- a/app/controllers/api/v2/quotas_controller.rb
+++ b/app/controllers/api/v2/quotas_controller.rb
@@ -2,15 +2,14 @@ module Api
   module V2
     class QuotasController < ApiController
       DEFAULT_INCLUDES = %w[quota_order_number quota_order_number.geographical_areas measures measures.geographical_area].freeze
-      POSSIBLE_INCLUDES = (%w[quota_balance_events] + DEFAULT_INCLUDES).freeze
 
       def search
-        render json: serialized
+        render json: serialized_quota_definitions
       end
 
       private
 
-      def serialized
+      def serialized_quota_definitions
         Api::V2::Quotas::Definition::QuotaDefinitionSerializer.new(
           quotas, serializer_options
         ).serializable_hash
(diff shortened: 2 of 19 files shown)


.rubocop.yml
```diff
@@ -30,6 +30,8 @@ RSpec/NestedGroups:
   Enabled: false
 Rails/SaveBang:
   Enabled: false
+Rails/UniqueValidationWithoutIndex:
```
willfish, line 33: This was causing rubocop to crash since it doesn't understand rails apps that use sequel :face_with_head_bandage:

app/controllers/api/v2/quotas_controller.rb
```diff
@@ -1,22 +1,41 @@
 module Api
   module V2
     class QuotasController < ApiController
-      before_action :find_quotas
+      DEFAULT_INCLUDES = %w[quota_order_number quota_order_number.geographical_areas measures measures.geographical_area].freeze
+      POSSIBLE_INCLUDES = (%w[quota_balance_events] + DEFAULT_INCLUDES).freeze
```
willfish, line 5: Remove me

app/controllers/application_controller.rb
```diff
@@ -7,33 +7,48 @@ class ApplicationController < ActionController::Base
   around_action :configure_time_machine
 
   unless Rails.application.config.consider_all_requests_local
-    rescue_from Exception,                           with: :render_error
-    rescue_from Sequel::RecordNotFound,              with: :render_not_found
-    rescue_from ActionController::RoutingError,      with: :render_not_found
-    rescue_from AbstractController::ActionNotFound,  with: :render_not_found
+    rescue_from Exception,                          with: :render_internal_server_error
+    rescue_from ArgumentError,                      with: :render_bad_request
```
willfish, line 11: Handles includes that do not exist for the serialized entity

cacquani (approved):

PR #96, opened 2021-06-09
Disable chief synchronizer

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-740

### What?

CHIEF endpoints return Excise and VAT measures and related resources in CSV format. We ETL these into the database via thei ChiefUpdate mechanism. This PR disables CHIEF ETL

I have added/removed/altered:

- [x] Disable chief synchronizer

### Why?

I am doing this because:

- The measures a related resources created by this ETL process are no longer used - UK measures are the source of truth for VAT/Excise duties


tavy88 (approved):
Looks good overall. We do need to run the ETL locally and make sure we don't get any weird notifications or failures now that CHIEF is silenced.

PR #101, opened 2021-06-27
HOTT-726: fill changes table

cacquani:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-726

### What?

I have added/removed/altered:

- [ ] Add worker that will generate changes table content overnight and clean up changes older than 3 months
- [ ] Run the worker every night at 4.30
- [ ] Add the logic to extract the changes

### Why?

I am doing this because we need to provide an effective changes API, and this development is a prerequisite for that.


db/migrate/20210610150945_create_deltas.rb
```diff
@@ -0,0 +1,13 @@
+Sequel.migration do
+  change do
+    create_table :deltas do
```
willfish, line 3: I don't think this table name actually describes what's in this table. So technically these are change_events that get created when a certain trigger happens - in our case when we detect that a certain difference occurs between two versions of the same object. Not sure if we can come up with a different name that would flag to a developer more automatically what these are all about?

lib/delta_tables_generator.rb
```diff
@@ -0,0 +1,59 @@
...
+require 'active_support/log_subscriber'
+
+module DeltaTablesGenerator
+  extend self
+
+  delegate :instrument, :subscribe, to: ActiveSupport::Notifications
+
+  # Starts the generation of the deltas data.
+  def generate(day: Date.current)
+    DeltaTablesBackend.with_redis_lock do
```
willfish, line 13: How comes we need to lock the redis table when generating the change events?

lib/delta_tables_generator.rb
```diff
@@ -0,0 +1,59 @@
...
+            CommodityCodeStarted,
+            CommodityCodeDescriptionChanged,
+            MeasureEndDated,
+            MeasureStarted,
+            MeasureDeleted,
+            MeasureCreatedOrUpdated,
+          ].map do |importer|
+            importer.perform_import(day: day)
+          end
+          return true
```
tavy88, line 27: If this returns a boolean value, maybe it should do the same in the rescue for consistency? In this case the ruby method name convention would be `generate?`. Does this need to return a boolean value?

lib/delta_tables_generator.rb
```diff
@@ -0,0 +1,59 @@
...
+            CommodityCodeStarted,
+            CommodityCodeDescriptionChanged,
+            MeasureEndDated,
+            MeasureStarted,
+            MeasureDeleted,
+            MeasureCreatedOrUpdated,
+          ].map do |importer|
+            importer.perform_backlog_import(from: from, to: to)
+          end
+          return true
```
tavy88, line 51: Same comment as: https://github.com/trade-tariff/trade-tariff-backend/pull/101/files#r661256843

lib/delta_tables_generator/importer.rb
```diff
@@ -0,0 +1,191 @@
...
+        %i[goods_nomenclature_item_id
+           goods_nomenclature_sid
+           productline_suffix
+           end_line
+           delta_type
+           delta_date]
+      end
+
+      def where_condition(day: Date.current)
+        'true'
```
tavy88, line 34: Should this not be `raise NotImplementedError, 'Implement this method in the subclasses' ? Looking at the children this method is supposed the return a SQL statement as string.

spec/unit/delta_tables_generator_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+describe DeltaTablesGenerator do
+  let(:db) { Sequel::Model.db }
+
+  describe '#generate' do
+    context 'on empty database' do
+      before do
+        db[:measures].delete
+        db[:goods_nomenclatures].delete
```
tavy88, line 10: Is the DB not empty on the spec runtime?

lib/delta_tables_generator/importer.rb
```diff
@@ -0,0 +1,191 @@
...
+              (element[:goods_nomenclature_item_id] > item_id ||
+               (element[:goods_nomenclature_item_id] == item_id &&
+                suffix > row_suffix))
+          end
+        end
+
+        return children
+      end
+
+      def find_children(row:, day: Date.current)
```
tavy88, line 103: This method is quite heavy. Maybe move each sql statement in separate methods and break it in smaller chunks?

db/migrate/20210610150945_create_deltas.rb
cacquani: The name is technically correct (the "delta", aka changes in the data, between one day and the previous) but not very descriptive.
I'll think about it while I fix the rest

lib/delta_tables_generator.rb
cacquani: It actually doesn't need to return anything at all, and it's definitely not a question mark method, since the return value is not the purpose of the computation.
The `true` was a mean to:
1. signal that the execution went OK, and 
2. override the fact that by default it would return the array, which is exactly as useless anyway.

Would returning `nil` on success (and raising if there are issues) be a good solution in your view?

lib/delta_tables_generator.rb
cacquani: It's a leftover from an initial version, I'll remove it, thanks for pointing it out :)

lib/delta_tables_generator/importer.rb
cacquani: Good point, thanks!

spec/unit/delta_tables_generator_spec.rb
cacquani: It should, but I had some erratic failures in the past due to specs running in random order and db cleaner not working as I expected, so I just wanted to be extra sure here

lib/delta_tables_generator/importer.rb
cacquani: True, that helps making it a bit more readable, thanks :)

spec/unit/delta_tables_generator_spec.rb
tavy88: Yeah fair enough.

lib/delta_tables_generator.rb
```diff
@@ -0,0 +1,51 @@
...
+        MeasureDeleted,
+        MeasureCreatedOrUpdated,
+      ].map do |importer|
+        importer.perform_backlog_import(from: from, to: to)
+      end
+      return nil
+    rescue StandardError => e
+      instrument('failed_generation.delta_tables_generator', exception: e)
+      raise e.original
+    end
```
tavy88, line 49: Yep better now! Thanks.

lib/delta_tables_generator.rb
```diff
@@ -0,0 +1,51 @@
+require_relative './delta_tables_generator/logger'
+
+require 'active_support/notifications'
+require 'active_support/log_subscriber'
+
+module DeltaTablesGenerator
+  extend self # rubocop:disable Style/ModuleFunction
```
willfish, line 7: How do you feel about just prefixing the methods with self and not using this extend?

cacquani: Needed some adjustment for the instrumenting part, but yeah, it's actually a better idea, thanks!

db/migrate/20210628165555_add_unique_constraint_to_deltas.rb
```diff
@@ -0,0 +1,10 @@
+Sequel.migration do
+  change do
```
willfish, line 2: Add a down to enable rollback removing constraint if we can for emergency

db/migrate/20210610150945_create_deltas.rb
```diff
@@ -0,0 +1,13 @@
...
+      primary_key :id
+      String  :goods_nomenclature_item_id
+      String  :goods_nomenclature_sid
```
willfish, line 6: Do we want to make this match the type of the other table to make it easy to setup relationships on the delta model at some later stage/be consistent

spec/factories/delta_factory.rb
```diff
@@ -0,0 +1,14 @@
+FactoryBot.define do
+  factory :delta, class: 'Delta' do
+    sequence(:goods_nomenclature_sid) { |n| n }
+
+    goods_nomenclature_item_id { 10.times.map { Random.rand(1..9) }.join }
```
willfish, line 5: I think we can reuse a sequence from the goods nomenclature factory?

db/migrate/20210628165555_add_unique_constraint_to_changes.rb
```diff
@@ -0,0 +1,16 @@
...
+      add_unique_constraint(
+        %i[goods_nomenclature_sid change_date],
+        name: :changes_upsert_unique
+      )
+    end
+  end
+
+  down do
+    alter_table :changes do
+      drop_constraint(:changes_upsert_unique)
```
willfish, line 13: :+1:

db/structure.sql
```diff
@@ -836,6 +836,35 @@ CREATE SEQUENCE public.certificates_oid_seq
...
 
 
+--
+-- Name: changes; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.changes (
+    id integer NOT NULL,
+    goods_nomenclature_item_id text,
+    goods_nomenclature_sid text,
```
willfish, line 846: Do we want to update the type of this one?

willfish (approved):

db/structure.sql
cacquani: True, I had forgot this, sorry

db/structure.sql
```diff
@@ -843,7 +843,7 @@ ALTER SEQUENCE public.certificates_oid_seq OWNED BY public.certificates_oplog.oi
 CREATE TABLE public.changes (
     id integer NOT NULL,
     goods_nomenclature_item_id text,
-    goods_nomenclature_sid text,
+    goods_nomenclature_sid integer,
```
willfish, line 846: Perfect :smile_cat:

willfish (approved):

PR #103, opened 2021-06-29
Make sure all MFN measures are zero duty

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-784

### What?

In the situation where we have multiple third country duties with additional codes (e.g. covid-related commodities) we want the backend to flag the commodity as having a non-zero amount on the third country duty if and only if _all_ the third country duties have components that are 0.

I have added/removed/altered:

- [x] Make sure MFN duties are _all_ zero for mfn zero check
- [x] Adds a presenter spec to more easily test the behaviour of the presenter without needing to do a full integration

### Why?

I am doing this because:

- We need to make sure we're returning true only when all the mfn measures are zero in the xi service for the duty calculator

### DIFF
--- a/spec/factories/goods_nomenclature_factory.rb
+++ b/spec/factories/goods_nomenclature_factory.rb
@@ -29,14 +29,6 @@ FactoryBot.define do
       validity_end_date   { nil }
     end
 
-    trait :fifteen_years do
-      validity_start_date { Date.current.ago(15.years) }
-    end
-
-    trait :twenty_years do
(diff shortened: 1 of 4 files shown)


spec/factories/goods_nomenclature_factory.rb
```diff
@@ -29,14 +29,6 @@
       validity_end_date   { nil }
     end
 
-    trait :fifteen_years do
```
willfish, line 32: These are unused traits

tavy88 (approved):

PR #104, opened 2021-07-05
HOTT-726: create changes API

cacquani:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-726

### What?

I have added/removed/altered:

- [ ] Add the API endpoint
- [ ] Add specs
- [ ] Builds on https://github.com/trade-tariff/trade-tariff-backend/pull/101, most of the changes will disappear once that is merged

### Why?

I am doing this because we need to provide an effective changes API

### Have you? (optional)

- [ ] Added documentation for new apis (yet to do)


willfish (approved):

app/serializers/api/v2/changes_serializer.rb
```diff
@@ -0,0 +1,14 @@
+module Api
+  module V2
+    class ChangesSerializer
+      include JSONAPI::Serializer
+
+      set_type :change
```
willfish, line 6: I think the type gets inferred from the class name if you don't specify it here? I think json api should serialize included types as the plural form, too - it just so happens that we're not completely compliant with our types for other serializers.

willfish (approved):

PR #111, opened 2021-07-14
Fix integration in the populate_changes_table_worker

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Updated broken reference in change table population worker
- [x] Added a spec to avoid regressions

### Why?

I am doing this because:

- https://sentry.io/organizations/engine-le/issues/2498388253/?project=5557005&referrer=slack


cacquani (approved):
Thanks for spotting it!

tavy88 (approved):

PR #115, opened 2021-07-15
Refactor quota search service

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-689

### What?

I have added/removed/altered:

- [x] Consolidated TimeMachine functionality by overriding the
  actual_date that is used around the action
- [x] Removed the years filtering since all time filtering is managed by
  the time machine - this is more idiomatic with the rest of the
  application - this may break integrations (but we believe we're the
  only ones using this. We need to investigate api usage to see before
  we go live with this change.
- [x] Shake out and refactor the tests
  - Remove duplication
  - Fix naming of contexts and examples
  - Use subject and filter with lets properly
  - Remove redundant specs (especially of day, month, year and years
    filtering)
  - Tests necessary and sufficient behaviour
- [x] Handle join without requiring an exact match on the
  validity_start_date (this is an overlap with the TimeMachine
  functionality and excludes the wrong records). These start dates are
  handled by different processes and needed removing
- [x] Force the service to conform to rails service api

### Why?

I am doing this because:

- The service TimeMachine functionality was just broken and not idiomatic

### AC

- [ ] Review https://github.com/trade-tariff/trade-tariff-frontend/issues/117
- [ ] Madhu does a manual test of the quota search service

### What's not covered

- Handling geographical area filtering in a way that works with exclusions, groups, etc
- Handling goods_nomenclature_item_id filtering properly according to the hierarchy of the tree
- Does not remove filters we suspect we do not use

### DIFF
--- a/app/controllers/api/v2/quotas_controller.rb
+++ b/app/controllers/api/v2/quotas_controller.rb
@@ -52,8 +52,8 @@ module Api
 
       def actual_date
         Date.parse([params['year'], params['month'], params['day']].join('-'))
-      rescue ArgumentError # empty date, month and year params means today
-        Date.current
+      rescue ArgumentError # empty date, default to as_of in ApplicationController
+        super
       end
     end
   end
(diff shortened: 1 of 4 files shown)


tavy88 (approved):
Looks good but needs proper testing.

PR #121, opened 2021-07-20
Adds excise overlays

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-832

### What?

I have added/removed/altered:

- [x] Added additional code overlays for excise additional codes

### Why?

I am doing this because:

- The current descriptions that would automatically come back via the ApplicableAdditionalCodeService class isn't sufficient/doesn't look nice or make any sense


jebw (approved):
:+1:

PR #123, opened 2021-07-21
Hott 854 add fallback for missing measurement units

jebw:
### Jira link

https://transformuk.atlassian.net/jira/software/projects/HOTT/boards/96?selectedIssue=HOTT-854

### What?

I have added a fallback for when a Measurement Unit is requested which is not listed in the YML:

- [x] Added an `#measurement_unit` method to access a specific unit, rather than going to the units hash directly
- [x] `#measurement_unit` will now query the DB and create 'stubbed measurement unit' in the units hash if found in the DB
- [x] `#measurement_unit` will log to Sentry with the unit code of the missing measurement unit
- [x] made `#measurement_units` private - any public access should go via the new singular method with appropriate fallback handling

### Why?

I am doing this because:

- The commodities listed in ticket `HOTT-825` were triggering a key error because whilst they were valid and present in the database, they had no corresponding entries in the YML file of questions (`db/measurement_units.yml`)

### Points for discussion? (optional)

- I've left the existing `#measurement_units` method public _but_ there are no consumers of this method now and it would make sense to make it private since that will force consumers down the path with fallback handling (ie `MeasurementUnit#measurement_unit`)

Screenshot from 2021-07-21 12-21-37

### DIFF
--- a/app/models/measurement_unit.rb
+++ b/app/models/measurement_unit.rb
@@ -13,6 +13,20 @@ class MeasurementUnit < Sequel::Model
 
   delegate :description, to: :measurement_unit_description
 
+  class << self
+    def measurement_units
+      @measurement_units ||=
+        begin
+          file = File.join(::Rails.root, 'db', 'measurement_units.json').freeze
+          JSON.parse(File.read(file))
+        end
+    end
+
+    def measurement_unit(unit_key)
+      measurement_units[unit_key]
+    end
+  end
+
   def to_s
     description
   end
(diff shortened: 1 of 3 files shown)


tavy88 (approved):
Super nice and clean!

willfish (approved):

willfish:
Looks good man. I wish these labels could just get updated in the database :)

willfish:
And not by us

PR #124, opened 2021-07-21
Remove exchange rates

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-855

### What?

I have added/removed/altered:

- [x] Removes the exchange rate endpoint

### Why?

I am doing this because:

- This changes too frequently to be useful for traders
- We need to follow the standards set by the Taric files


config/routes.rb
```diff
@@ -103,8 +103,6 @@
         collection { get :search }
       end
 
-      resources :exchange_rates, only: [:index]
```
jebw, line 106: This removes an API endpoint - should this be returning a 410 instead?

jebw (commented):
Looks simple, only question I have is the removal of an API endpoint? Is that going to surprise any users? Should the route return a 410 instead?

jebw (approved):

PR #127, opened 2021-07-23
Hott 720 eager loading fixes

jebw:
### Jira link

[HOTT-720](https://transformuk.atlassian.net/jira/software/projects/HOTT/boards/96?selectedIssue=HOTT-720)

### What?

I have added/removed/altered:

- [x] Changed the log level in Dev - this enables SQL query logging and will help surface N+1 queries
- [x] Fixed an N+1 on `/api/v2/geographical_areas/countries` - see `CachedGeograpicalAreaService`
- [x] Fixed an N+1 for `FootnoteDescription`'s on `/api/v2/commodities`

### Why?

I am doing this because:

- We were generating a lot of N+1 queries on the `/commodities/:id` endpoint on the frontend - this is a common end point and this code path will be used frequently

### DIFF
--- a/app/services/cached_commodity_service.rb
+++ b/app/services/cached_commodity_service.rb
@@ -77,7 +77,6 @@ class CachedCommodityService
     { geographical_area: [:geographical_area_descriptions,
                           { contained_geographical_areas: :geographical_area_descriptions }] },
     { additional_code: :additional_code_descriptions },
-    :footnotes,
     :base_regulation,
     :modification_regulation,
     :full_temporary_stop_regulations,
--- a/app/services/cached_geographical_area_service.rb
+++ b/app/services/cached_geographical_area_service.rb
@@ -24,7 +24,10 @@ class CachedGeographicalAreaService
   ].freeze
 
   DEFAULT_INCLUDES = [:contained_geographical_areas].freeze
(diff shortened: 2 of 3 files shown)


app/services/cached_commodity_service.rb
```diff
@@ -64,7 +64,7 @@ class CachedCommodityService
         {
           measure_condition_components: [
             { measurement_unit: %i[measurement_unit_description measurement_unit_abbreviations] },
-            :measure_condition,
+            { measure_condition: { measure: :measure_type } },
```
tavy88, line 67: What is your plan in terms of regression testing this? Making sure nothing has broken. Pretty tough to test all cases - but I guess if our smoke tests + unit tests - should be ok?

jebw: My (assumption) was that there's no behavioural change so no change to specs. its an internal detail (does it eager load or not?)

I'm guessing there's not any existing mechanism to test for N+1s or these would already be getting picked up? I think it should be possible to include some kind of logging of query counts but actual N+1 detection I'm not sure about?

tavy88: Makes sense!

tavy88 (approved):

willfish (approved):

PR #130, opened 2021-07-29
Handle units that belong to measure condition components

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Units can come from either the measure conditions or the measure
condition components

### Why?

I am doing this because:

- This is needed when generating duty calculations against measure condition components


willfish:
Approved on call by @tavy88

PR #131, opened 2021-07-30
Fix n + 1 issue due to inverse load of measure via the measure condition

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-879

### What?

I have added/removed/altered:

- [x] Fixes inverse load n + 1 issue where the measure condition component was loading the measure condition and the measure to pull out the measure type over and over just to avoid doing currency conversion on excise measure types.
- [x] Removed unused/disabled currency conversion feature from the DutyExpressionFormatter

### Why?

I am doing this because:

- The currency conversion feature isn't being used.
- The n + 1 inverse load was creating a ton of SQL queries that are slowing down calls to the commodity api

### DIFF
--- a/app/models/measure_condition_component.rb
+++ b/app/models/measure_condition_component.rb
@@ -50,7 +50,6 @@ class MeasureConditionComponent < Sequel::Model
       measurement_unit_qualifier: measurement_unit_qualifier,
       currency: TradeTariffBackend.currency,
       formatted: true,
-      excise: measure_condition.measure.excise?,
     )
   end
 end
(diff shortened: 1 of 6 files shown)


app/models/measure_condition_component.rb
```diff
@@ -50,7 +50,6 @@ def formatted_duty_expression
       measurement_unit_qualifier: measurement_unit_qualifier,
       currency: TradeTariffBackend.currency,
       formatted: true,
-      excise: measure_condition.measure.excise?,
```
willfish, line 53: Fixes the n + 1

jebw (approved):

PR #132, opened 2021-07-30
HOTT-866 extend legal act node

jebw:
### Jira link

[HOTT-866](https://transformuk.atlassian.net/jira/software/projects/HOTT/boards/96?selectedIssue=HOTT-866)

### What?

I have added/removed/altered:

- [x] Added a spec for `MeasuresLegalActSerializer` to check for regressions to existing behaviour
- [x] Added a `MeasureLegalActPresenter` to allow manipulating the legal act data before serializing it
- [x] Changed the `MeasurePresenter` to utilise the MeasureLegalActPresenter for legal acts
- [x] Added a `UkRegulationParser` class to parse the concatenated `information_text` for uk legal acts applying after 01 Jan 2021
- [x] Changed the `MeasureLegalActPresenter` to use `UkRegulationParser` to extract uk descriptions, urls and codes for the UK dataset, when the `officialjournal_number` == '1' and the `officialjournal_page` == 1
- [x] Added fallback handling for the above when the `information_text` field doesn't comply with the new concatenated format
- [x] Change the `MeasureLegalActPresenter` to return blank fields for regulation_url, regulation_code and description when either the regulation id is `IYY99990` or the measure type id is either `305` or `306`

### Why?

I am doing this because:

- Uk regulations after 01 Jan 2021 will be holding their regulation code, regulation url and description in a concatenated format inside of the `information_text` field instead of using the existing fields

This changes the format of the legal act API nodes in 3 ways
- for uk regulations after 01 Jan 2021 changes from the existing fields, to extracting regulation code/url/description from the information text
- it hides the code/url/description for all regulations with an id of `IYY99990` or measure type id of 305 or 306
- it adds a description attribute for all other regulations with the contents of the information text database field

### DIFF
--- a/app/presenters/api/v2/measures/measure_legal_act_presenter.rb
+++ b/app/presenters/api/v2/measures/measure_legal_act_presenter.rb
@@ -3,6 +3,7 @@ module Api
     module Measures
       class MeasureLegalActPresenter < SimpleDelegator
         attr_reader :regulation, :measure
+
         delegate :regulation_url, :regulation_code, :description,
                  to: :uk_regulation_data, prefix: :uk
 
@@ -48,7 +49,7 @@ module Api
 
         def uk?
           TradeTariffBackend.uk? &&
-            regulation.officialjournal_number == "1" &&
+            regulation.officialjournal_number == '1' &&
             regulation.officialjournal_page == 1
         end
 
--- a/app/presenters/api/v2/measures/measure_legal_act_presenter.rb
+++ b/app/presenters/api/v2/measures/measure_legal_act_presenter.rb
@@ -3,10 +3,17 @@ module Api
     module Measures
       class MeasureLegalActPresenter < SimpleDelegator
         alias_method :regulation, :__getobj__
+        attr_reader :measure
         delegate :regulation_url, :regulation_code, :description,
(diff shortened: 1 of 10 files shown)


app/presenters/api/v2/measures/measure_legal_act_presenter.rb
```diff
@@ -0,0 +1,80 @@
+module Api
+  module V2
+    module Measures
+      class MeasureLegalActPresenter < SimpleDelegator
+        alias_method :regulation, :__getobj__
```
willfish, line 5: This feels weird. Like we're exposing implementation detail of some python class :smile:. Not very ruby but :man_shrugging:

willfish:
Wondering whether you've run `bundle exec rubocop -a <files/i've/changed>`

willfish (approved):
Seems pretty straightforward if not a bit horrible how we have to handle weird delimiters and such.

app/presenters/api/v2/measures/measure_legal_act_presenter.rb
jebw: Yeh - I should this - its a hangover from an earlier commit when I wasn't creating a custom initialize method

PR #136, opened 2021-08-10
Updates schedule and triggers reindexing of es models after clearing redis

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-889

### What?

I have added/removed/altered:

- [x] After we've cleared the redis cache trigger TradeTariffBackend.reindex and TradeTariffBackend.recache
- [x] Updates UpdateSynchronizerWorker to reflect CDS changes of file availability

### Why?

I am doing this because:

- We only want to recache after a UpdateSynchronizerWorker has triggered but I felt that was too much of a change (affecting both sets of workers)
- This reduces the number of schedules we have to manage


jebw (approved):
LGTM :+1:

PR #137, opened 2021-08-10
887 measure type exclusions

jebw:
### Jira link

[HOTT-887](https://transformuk.atlassian.net/browse/HOTT-887)

### What?

I have added/removed/altered:

- [x] Added CSV exclusion lists for specific measure types for specific geographical areas (`MeasureTypeExclusion` model)
- [x] Change `Api::V2::MeasurePresenter` to utilise this for its serialized list of excluded Geographical Areas

### Why?

I am doing this because:

- The data we receive in the feeds does not contain the exclusion information for these measure types

- It extends the list of excluded countries in the API - it is entirely code based change, and as such revertable.

### Review notes

- If you go to https://dev.trade-tariff.service.gov.uk/xi/api/v2/commodities/9701100000 - and search for excluded countries - you'll see `GB` in the list of excluded countries for `measure_id`=`3088491`


spec/models/measure_type_exclusion_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  let(:test_exclusions) { {} }
+  let(:test_csv_file) { file_fixture('measure_type_exclusions.csv') }
+
+  describe '.load_from_file' do
+    subject { described_class.load_from_file(test_csv_file).exclusions }
+
+    it { is_expected.to eql(%w[735 1008] => %w[GB JE GG]) }
+  end
+
+  describe '.load_from_string' do
```
willfish, line 17: How do you feel about killing this method and specs since we're not using it?

jebw: Yeh, will do - put it in originally for testing - but then ended up adding a fixture to test the file loading anyway

PR #145, opened 2021-08-11
Removes mimemagic

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes the mimemagic gem

### Why?

I am doing this because:

- This was only required to magically handle/understand gzipped annual files. We never download these and can manually gunzip them if necessary by passing the correct flags
- mimemagic was requiring extra files to be provided as part of dockerising this application


mafonso (approved):
LGTM

PR #149, opened 2021-08-13
Bumps ruby to 3.0.2

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Bumps ruby to 3.0.2


config/application.rb
```diff
@@ -14,11 +14,8 @@ class Application < Rails::Application
     config.load_defaults 6.0
     require 'trade_tariff_backend'
 
-    # lib directory to be autoloadable.
     config.eager_load_paths << Rails.root.join('lib')
 
-    config.autoloader = :classic
```
jebw, line 20: This is orthogonal to the Ruby 3.0 upgrade? Are there definitely no auto load issues with old classes and Zeitwerk?

jebw (commented):
Looks good - only comment is around the switch to Zeitwek?

jebw (dismissed):
Looking at CI - looks like this is failing to boot on the servers?

config/application.rb
willfish: So what I noticed was some constants weren't loading properly. So I changed to zeitwerk for a couple of reasons:

1. It's the recommend way
2. Our constant loading is a bit insane and mixes different methods
3. It fixed the specific issues I was seeing

I'll have a play with this locally and see if I can't reproduce the failure. If I can and can fix it, I'll do a full loading of every constant we define in the app in a scripted way

mafonso (commented):
LGTM

mafonso (approved):
LGTM

PR #150, opened 2021-08-16
Migrate from classic autoloader to zeitwerk

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-900

### What?

I have added/removed/altered:

- [x] Moves to zeitwerk autoloader

### Why?

I am doing this because:

- See https://github.com/trade-tariff/trade-tariff-backend/pull/149

### DIFF
--- a/spec/unit/taric_importer_spec.rb
+++ b/spec/unit/taric_importer_spec.rb
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-describe TaricImporter do
+xdescribe TaricImporter do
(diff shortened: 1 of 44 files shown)


spec/unit/taric_importer_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-describe TaricImporter do
+xdescribe TaricImporter do
```
willfish, line 3: Fix me

PR #157, opened 2021-08-16
Use zeitwerk

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-900

### What?

I have added/removed/altered:

- [x] Move to zeitwerk from the antequated classic autoloader


jebw (approved):
LGTM - guessing you're running the e2e specs against dev as well - assuming those pass all looks great!

PR #159, opened 2021-08-17
Lint all touched files

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Automatically lints touched files

### Why?

I am doing this because:

- This is in a separate PR to isolate large linting changes


.rubocop.yml
```diff
@@ -42,6 +42,8 @@ Rails/FilePath:
   Enabled: false
 Style/Documentation:
   Enabled: false
+Style/FormatString:
```
willfish, line 45: We have format as an ovverride for formatted descriptions/expressions. Fixing these violations would break custom formatting

spec/models/measurement_unit_spec.rb
```diff
@@ -45,10 +45,11 @@
     end
 
     context 'with missing measurement unit present in database' do
+      subject! { MeasurementUnit.measurement_unit(unit_code) }
+
       before { allow(Raven).to receive(:capture_message).and_return(true) }
-      before { allow(MeasurementUnit).to receive(:measurement_units).and_return({}) }
 
-      subject! { MeasurementUnit.measurement_unit(unit_code) }
```
jebw, line 51: This looks wrong - previously the sequence was

1. First `before` - watch `Raven`
2. Second `before`
3. `subject!` - which will perform an action which could potentially trigger `Raven`

Now the sequence is

1. `subject!` - perform action which may trigger `Raven`
2. first `before` - set up watching of `Raven` - this is now too late
3. second `before`

jebw (changes requested):
Looks great - I think the PR should be against `master` and the other branch rebased onto it

Only minor issue is the change to the spec ordering in `measurement_unit_spec.rb` also looks like it'll fail the `Raven` testing

If RSpec ordering is enforced the spec will need tweaking to work differently I think

willfish:
My plan is to sort out the validation removal and once we'd gotten it onto production merge this with the fixes to the tests, etc

PR #160, opened 2021-08-17
Adds migration for measure with missing validity end date

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-901

### What?

I have added/removed/altered:

- [x] Adds migration for missing validity_end_date on xi measure

### DIFF
--- a/db/data_migrations/20210817124456_update_validity_end_date.rb
+++ b/db/data_migrations/20210817124456_update_validity_end_date.rb
@@ -6,8 +6,8 @@ TradeTariffBackend::DataMigrator.migration do
 
     apply do
       if TradeTariffBackend.xi?
-        end_date = Time.zone.parse( '2020-02-26')
-        measure = Measure.where(measure_sid:  3753097).last
+        end_date = Time.zone.parse('2020-02-26')
+        measure = Measure.where(measure_sid: 3_753_097).last
 
         return if measure.validity_end_date == end_date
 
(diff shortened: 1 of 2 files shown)


db/data_migrations/20210817124456_update_validity_end_date.rb
```diff
@@ -0,0 +1,21 @@
...
+    applicable { false }
+
+    apply do
+      if TradeTariffBackend.xi?
+        end_date = Time.zone.parse('2020-02-26')
+        measure = Measure.where(measure_sid: 3_753_097).last
+
+        return if measure.validity_end_date == end_date
+
+        measure.update(validity_end_date: end_date)
```
jebw, line 14: This assigns the new date if its anything but the expected one? Should it be checking that its blank first?

jebw (approved):
This doesn't quite do the same as Matt's SQL - matts version updates the existing Oplog entry

willfish:
I know @jebw . It achieves the same result, though. In fact this way is better, it makes a copy of the latest operation and sets the correct validity start date on the new value. It's easier/safer to rollback, etc

db/data_migrations/20210817124456_update_validity_end_date.rb
willfish: I don't think so. TBH, I'd kind of want to throw this migration away once we're done. I guess the visibility is good.

PR #162, opened 2021-08-18
Hott755 dockerise backend

mafonso:
### Jira link

HOTT-755

### What?

I have added/removed/altered:

- [ ] Build docker image for the backend app (DEV only)
- [ ] Deploy docker backend service in blue-green (DEV only)
- [ ] Deploy docker based worker (DEV only)

### Why?

I am doing this because:
- Moving to docker based deployment 

Note before merging:  This PR is using temporary target apps (-docker suffixed) to avoid clash with the current ones. 
This before merging, the apps need to be renamed in CF and this PR edited accordingly.


Dockerfile
```diff
@@ -0,0 +1,53 @@
+# Build compilation image
+FROM ruby:2.7.4-alpine3.13 as builder
```
willfish, line 2:
```suggestion
FROM ruby:3.0.2-alpine3.13 as builder
```

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN rm -rf node_modules log tmp && \
+      rm -rf /usr/local/bundle/cache && \
+      rm -rf .env && \
+      find /usr/local/bundle/gems -name "*.c" -delete && \
+      find /usr/local/bundle/gems -name "*.h" -delete && \
+      find /usr/local/bundle/gems -name "*.o" -delete && \
+      find /usr/local/bundle/gems -name "*.html" -delete
+
+# Build runtime image
+FROM ruby:2.7.4-alpine3.13 as production
```
willfish, line 37:
```suggestion
FROM ruby:3.0.2-alpine3.13 as production
```

willfish: Does the latest alpine.x.xx image work these days? Might be worth bumping?

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+# Copy all files to /app (except what is defined in .dockerignore)
+COPY . /app/
+
+
+# Cleanup to save space in the production image
+RUN rm -rf node_modules log tmp && \
+      rm -rf /usr/local/bundle/cache && \
+      rm -rf .env && \
+      find /usr/local/bundle/gems -name "*.c" -delete && \
```
willfish, line 31: Can you achieve this in one with a regex?

mafonso: Probably, but I think it clearer this way, as each line is targeting a specif "left over". 
Might not be worth the time getting it right, and there will be some risk.

mafonso:
We bumped the docker version in Dockerfile, but the Gemfile is still behind. 
Maybe it's best to deal with the Ruby version upgrade in a separate PR?

willfish (approved):
LGTM

PR #163, opened 2021-08-18
Removes spec requires

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- Handle rspec standard requires
- Remove spec requires handled by zeitwerk

### Why?

I am doing this because:

- We've moved to zeitwerk and all these constants should be automatically loaded


.rspec
```diff
@@ -1,2 +1,2 @@
 --color
---require spec_helper
```
willfish, line 2: spec helper is basically never used - I'd be tempted to remove it at some point

willfish:
I'm ignoring the noisy linting for the most part atm

PR #173, opened 2021-08-19
Relate measure units to measure components and measure condition components

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-907

### What?

I have added/removed/altered:

- [x] Surface the formatted composite primary key for both measure
  components and measure condition components in the measure unit service
- [x] Adds a standardised api for both types of components using a shared
  example

### Why?

I am doing this because:

- These are needed to understand how answers for unit relate to the specific component we're doing a duty calculation on


app/models/measure_condition.rb
```diff
@@ -122,22 +122,10 @@ def entry_price_system?
   end
 
   def expresses_unit?
-    condition_measurement_unit_code || measure_condition_components.any?(&:expresses_unit?)
+    measure_condition_components.any?(&:expresses_unit?)
   end
 
   def units
-    if condition_measurement_unit_code
```
willfish, line 129: The condition_measurement_unit_code is for trigger price indications as part of the trigger price system and isn't used to calculate duties

app/models/measure_component.rb
```diff
@@ -58,6 +58,18 @@ def ad_valorem?
       monetary_unit_code.nil?
   end
 
+  def expresses_unit?
+    measurement_unit_code
```
jebw, line 62: is truthy right here? could it end up as a blank string?

or should it be `.present?`

jebw (approved):
LGTM

app/models/measure_component.rb
willfish: Not required from my experience of measure components previously, no

PR #174, opened 2021-08-20
HOTT-846: Rules of origin api

jebw:
### Jira link

[HOTT-846](https://transformuk.atlassian.net/browse/HOTT-846)

### What?

I have added/removed/altered:

- [x] Added an api for querying rules and schemas relevant to commodities - loads mocked data unless feature flag enabled

### Why?

I am doing this because:

- As part of HOTT-846 we will need to query the api rules to show on the front end


app/models/rules_of_origin/data_set.rb
```diff
@@ -22,6 +24,29 @@ def default_heading_mappings
...
         end
       end
+
+      # FIXME: Mocked data - to be removed once data loading is turned on
+
+      def load_mocked
+        new mocked_scheme_set, mocked_rule_set, mocked_mappings
+      end
+
+      def mocked_scheme_set
```
alexdesi, line 34: it seems mocked_scheme_set, mocked_rule_set and mocked_mappings should be private methods

app/models/rules_of_origin/query.rb
```diff
@@ -11,7 +11,9 @@ def initialize(data_set, heading_code, country_code)
     end
 
     def rules
-      id_rules.map(&rule_set.method(:rule))
+      schemes_and_their_id_rules.transform_values do |id_rules|
+        id_rules.map(&rule_set.method(:rule))
+      end
```
alexdesi, line 16: so, this is returning a hash with the same keys, but the values are
the id_rules mapped to rule_set.method(:rule),
it would be nice to simplify this a bit, or maybe adding a comment 😄

db/rules_of_origin/mocked_schemes.json
```diff
@@ -0,0 +1,26 @@
...
+      "countries": [
+        "FR",
+        "ES",
+        "IT"
+      ],
+      "footnote": "This is a mocked footnote"
+    }
+  ]
+}
+
```
alexdesi, line 26: this is not `jsonapi` isn't it?
just asking, maybe don't want to use it here

spec/presenters/api/v2/rules_of_origin/scheme_presenter_spec.rb
```diff
@@ -0,0 +1,26 @@
+RSpec.describe Api::V2::RulesOfOrigin::SchemePresenter do
+  subject(:presenter) { described_class.new scheme, rules }
+
+  let(:scheme) { build :rules_of_origin_scheme }
+
+  let(:rules) do
+    build_list :rules_of_origin_rule, 3, scheme_code: scheme.scheme_code
+  end
+
+  it { is_expected.to be_instance_of described_class }
```
alexdesi, line 10: this sounds redundant (the described_class could be any class and this would be always true), we could remove it, up to you anyway 😄

spec/requests/api/v2/rules_of_origin_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      it { is_expected.to have_http_status :success }
+    end
+
+    context 'without matching country' do
+      let(:country_code) { 'ES' }
+
+      it { is_expected.to have_http_status :success }
+    end
+  end
+end
```
alexdesi, line 38: should also be testing for non-happy paths as well?

alexdesi (commented):
Nice one! I've added few comments but the logic seems correct (based on my short experience on the project).
Also some comments are just to understand how things work 😄

db/rules_of_origin/mocked_schemes.json
jebw: This is a mocked version of a json file in the system, schema is of Matt's designing

app/controllers/api/v2/rules_of_origin_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+        @presented_schemes = Api::V2::RulesOfOrigin::SchemePresenter.for_many(
+          @query.schemes,
+          @query.rules,
+        )
+
+        @serializer = Api::V2::RulesOfOrigin::SchemeSerializer.new(
+          @presented_schemes,
+          include: %i[rules],
+        )
```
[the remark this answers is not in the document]
jebw: I think if we're using it elsewhere we could look to convert to a service at that point - I'll drop the use of instance variables

app/models/rules_of_origin/data_set.rb
```diff
@@ -22,6 +24,29 @@ def default_heading_mappings
           importer.import(skip_invalid_rows: true)
         end
       end
+
+      # FIXME: Mocked data - to be removed once data loading is turned on
+
+      def load_mocked
+        new mocked_scheme_set, mocked_rule_set, mocked_mappings
```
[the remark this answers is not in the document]
jebw: :+1:

app/models/rules_of_origin/data_set.rb
jebw: These are private? just in the class' namespace?

app/models/rules_of_origin/query.rb
jebw: Yeh, good should - definitely hard to grok

I've moved the logic for fetching of multiple rules to RuleSet, and added comment in the `transform_values` to explain why we're calling RuleSet from here

spec/presenters/api/v2/rules_of_origin/scheme_presenter_spec.rb
jebw: Yeh, I know it looks unnecessary - its a stupid trivial test, but also very fast - I don't really care about the response but I want a very simple check to make sure the initializer is working correctly - all the other tests could fail because of different issues but this one can only fail if the initializer is not working right.

spec/requests/api/v2/rules_of_origin_controller_spec.rb
jebw: The second two context's are 'non-happy' paths - I'm trying to avoid getting into testing the serializer output in the request spec. I just want to establish that the controller is working as intended, ie not erroring on non-happy paths. 

I've extended to the two 'non-happy' paths to also check the response is json, as well as that its not erroring.

app/models/rules_of_origin/query.rb
alexdesi: thanks, I think it's much clear now!

alexdesi (approved):
thanks for the improvements!

PR #176, opened 2021-08-23
Handle extra qualifier in unit code key.

willfish:
### Jira link

HOTT-<TODO>

### What?

In the wild we retrieve the unit from the measure units file using the
code and the qualifier concatenated. This enables us to enrich the
descriptions a bit as they relate to the specific measure
component/measure condition component.

Our specific example `Factor` exists in the database but does not match
because it's code is FC1 and we were passing in the full key FC1X

```
"500 - Internal Server Error: Requested invalid measurement unit: FC1X"
```

There are only **18** measure condition components active in the database that have the unit key `FC1X`

I have added/removed/altered:

- [x] Changes the key that we use to look up the measurement unit
code in fallback scenarios to exclude the qualifier
- [x] Lint according to rubocop rules

### Why?

I am doing this because:

- This is needed for us to be able to present commodities with unsupported units

### DIFF
--- a/spec/models/measurement_unit_spec.rb
+++ b/spec/models/measurement_unit_spec.rb
@@ -1,5 +1,5 @@
 describe MeasurementUnit do
-  let(:measurement_unit) { create :measurement_unit, :with_description }
+  subject(:measurement_unit) { create :measurement_unit, :with_description }
 
   describe '#to_s' do
     it 'is an alias for description' do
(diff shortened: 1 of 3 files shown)


spec/models/measurement_unit_spec.rb
```diff
@@ -19,56 +17,60 @@
...
-      let(:unit) { MeasurementUnit.measurement_unit('UNKNOWN') }
+      let(:unit) { described_class.measurement_unit('UNKNOWN', 'UNKNOWN') }
 
-      it "will raise an InvalidMeasurementUnit exception" do
+      it 'will raise an InvalidMeasurementUnit exception' do
         expect { unit }.to raise_exception(MeasurementUnit::InvalidMeasurementUnit)
       end
     end
-
-    context 'without specifying a unit' do
```
willfish, line 70: Removed as this seemed defensive since we're the clients and the writers of the code that drives this and we'd know this error was propagated in the service if we weren't calling it with an argument

jebw (approved):
LGTM

PR #180, opened 2021-08-24
Handle generation of missing taric updates on the uk service

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Stop running the sequence worker on the uk service
- [x] Lint

### Why?

I am doing this because:

- That this is happening was noticed because we failed to do a deploy when a queued worker was getting failed responses from the Taric api
- It seemed simple enough to just stop this from running


jebw (approved):
LGTM - but the tests will need updating for the new prototype on `TaricSequenceChecker#new`

PR #181, opened 2021-08-24
HOTT-848: Scheme data 

jebw:
### Jira link

[HOTT-848](https://transformuk.atlassian.net/browse/HOTT-848)

### What?

I have added/removed/altered:

- [x] Added `RulesOfOrigin::Scheme`, `RulesOfOrigin::Link` and `RulesOfOrigin::Explainer`
- [x] Added `RulesOfOrigin::SchemeSet` to read the schemes file in
- [x] Assign the read data on boot to `RulesOfOrigin::SchemeSet.current`

### Why?

I am doing this because:

- We need to extract the schemes data from the supplied JSON files and present it to the user


app/models/rules_of_origin/scheme_set.rb
```diff
@@ -0,0 +1,65 @@
...
+      unless @source_file.extname == '.json' && @source_file.file? && @source_file.exist?
+        raise InvalidSchemesFile, 'Requires a path to a JSON file'
+      end
+
+      @base_path = @source_file.dirname
+      @source_file.read
+    end
+
+    def rebuild_countries_to_schemes_index
+      @_schemes.each.with_object({}) do |(scheme_code, scheme), countries|
```
willfish, line 57:
```suggestion
      @_schemes.each_with_object({}) do |(scheme_code, scheme), countries|
```

app/models/rules_of_origin/scheme_set.rb
```diff
@@ -0,0 +1,65 @@
+module RulesOfOrigin
+  class SchemeSet
+    class_attribute :current
+    attr_reader :base_path, :links
+
+    def initialize(source_file)
+      data = JSON.parse read_file(source_file)
```
willfish, line 7: How do you feel about not having side effects in your constructor? I guess you could either a) implement a builder for your class `def self.build` or b) pass the reading, parsing and error handled thing to the initialize method.

```ruby
def self.build(source_file)
  data = read_file(source_file)
  data = JSON.parse(data)
  links = data['links'].map(&Link.method(:new_with_check)).compact.freeze
  schemes = data['schemes'].map(&Scheme.method(:new))
                                 .index_by(&:scheme_code)
                                 .freeze
  countries = rebuild_countries_to_schemes_index.freeze
  new(links, schemes, countries)
end
```

I don't think you need the error handling around the ScopeDoesNotMatch since you're in control of the files that are being loaded?

willfish: Also, do we need to worry about runtime mutation of the objects? Probably not worth marking the containing data structures as immutable? It doesn't recurse through to each item in the containing structure AFAIK, too

db/rules_of_origin/roo_schemes_uk.json
```diff
@@ -0,0 +1,927 @@
+{
```
willfish, line 1: How do you feel about supporting a version of these files?

config/initializers/rules_of_origin.rb
```diff
@@ -0,0 +1,8 @@
+unless Rails.env.test?
+  Rails.application.reloader.to_prepare do
+    RulesOfOrigin::SchemeSet.current =
+      RulesOfOrigin::SchemeSet.new \
+        Rails.root.join \
+          "db/rules_of_origin/roo_schemes_#{TradeTariffBackend.service}.json"
```
willfish, line 6: If you moved this to being a constant on the SchemeSet model you could avoid haven't to pass special snowflake args to the builder and just have a consistent api between all of the models.

willfish:
```ruby
RulesOfOrigin::SchemeSet.build
```

app/models/rules_of_origin/scheme_set.rb
```diff
@@ -0,0 +1,65 @@
+module RulesOfOrigin
+  class SchemeSet
+    class_attribute :current
```
willfish, line 3: Do we want to avoid inheritance quirks with this way of setting class setters/getters? I guess it's the normal rails way to do this now :thinking:

app/models/rules_of_origin/scheme_set.rb
jebw: initialize stuff - yeh, agree - I've actually already got a similar change in a later PR I'm working on - I'll move it onto this. 

The immutable stuff helps with memory usage. I think I get the same benefits using frozen string literals pragma so I'll just add that to the top of all the files.

ScopesDoesNotMatch bit can be dropped - I figured the info was in the file so I'd check it against the version of the site it was getting loaded into - more as a protection against developer confusion.

app/models/rules_of_origin/scheme_set.rb
jebw: This is specifically intended to be for the full lifecycle of the application - as opposed to anything which changes

Could do something like assigning a constantant instead?

willfish: I actually don't mind. The effect is clear. It's just a bit new to me

PR #193, opened 2021-08-31
HOTT-868 Load rules of origin data

jebw:
### Jira link

[HOTT-868](https://transformuk.atlassian.net/browse/HOTT-868)

### What?

I have added/removed/altered:

- [x] Added models for `Rule` and `RuleSet`
- [x] Added an importer to load Rules from the CSV at boot time
- [x] Added the current CSV of rules and the CSV of headings
- [x] Added a model for HeadingMappings, which can import the mappings CSV at boot time
- [x] Added rake based validators for both the rules and the mappings CSV files
- [x] Feature flagged the Rules of Origin data loading at start up

### Why?

I am doing this because:

- We have 2 CSV files we need to read data from - the first is a list of rules belonging to a schema. The second is a set of mappings between heading codes and the rules from the first CSV

### Have you? (optional)

- Currently the loading of data at boot is feature flagged off

### Notes

- The supplied data files contain invalid data, for now I've set the loading of data to skip the invalid rows - this can be revisited once we have newer files


willfish (approved):
LGTM

PR #194, opened 2021-09-03
HOTT-868 - Rules of origin queries

jebw:
### Jira link

[HOTT-868](https://transformuk.atlassian.net/browse/HOTT-868)

### What?

I have added/removed/altered:

- [x] Added a `RulesOfOrigin::DataSet` model to group the various file based data sources together
- [x] Added a `RulesOfOrigin::Query` model to query the data set model
- [x] Fixed several edge case bugs in the other RoO data models
- [x] Added various factories to make it easier to generate data for testing

### Why?

I am doing this because:

- We need a way to query our file based data so that we can return the Rules of Origin data in the API

### DIFF
--- a/app/models/rules_of_origin/rule.rb
+++ b/app/models/rules_of_origin/rule.rb
@@ -21,5 +21,11 @@ module RulesOfOrigin
     validates :scheme_code, presence: true, format: SCHEME_CODE_FORMAT
     validates :heading, presence: true
     validates :rule, presence: true
+
+    def ==(other)
+      id_rule.present? &&
+        other.respond_to?(:id_rule) &&
+        other.id_rule == id_rule
+    end
   end
 end
--- a/app/models/rules_of_origin/rule.rb
+++ b/app/models/rules_of_origin/rule.rb
@@ -3,12 +3,19 @@
 module RulesOfOrigin
   class Rule
     include ActiveModel::Model
+    include ActiveModel::Attributes
 
     ID_RULE_FORMAT = %r{\A\d+\z}.freeze
     SCHEME_CODE_FORMAT = %r{\A[a-zA-Z\-]+\z}.freeze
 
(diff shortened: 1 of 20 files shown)


app/models/rules_of_origin/rule.rb
```diff
@@ -3,16 +3,29 @@
...
   class Rule
     include ActiveModel::Model
+    include ActiveModel::Attributes
 
     ID_RULE_FORMAT = %r{\A\d+\z}.freeze
     SCHEME_CODE_FORMAT = %r{\A[a-zA-Z\-]+\z}.freeze
 
-    attr_accessor :id_rule, :scheme_code, :heading, :description,
-                  :quota_amount, :quota_unit, :rule, :alternate_rule
+    attribute :id_rule
```
willfish, line 11: :heart:

spec/factories/rules_of_origin/data_set_factory.rb
```diff
@@ -0,0 +1,20 @@
+FactoryBot.define do
+  factory :rules_of_origin_data_set, class: 'RulesOfOrigin::DataSet' do
+    initialize_with { new scheme_set, rule_set, heading_mappings }
```
willfish, line 3: Crazy eval wizardry :dancers:

spec/models/rules_of_origin/heading_mappings_spec.rb
```diff
@@ -61,13 +61,13 @@
     context 'with unknown scheme code' do
       let(:scheme_code) { 'unknown' }
 
-      it { is_expected.to be_empty }
+      it { is_expected.to eq({}) }
```
willfish, line 64: :+1:

jebw: Found I was returning an empty array - when results should be a hash - realised the weakness of checking for `empty?` but not type

PR #195, opened 2021-09-03
Surface declarable commodities and headings and expose search as GET

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-811

### What?

I have added/removed/altered:

- [x] Adds get route for v2 search
- [x] Surface declarable on the heading and commodity

### Why?

I am doing this because:

- Expose the search so that Single Trade Window team can use the correct v2 api to do a search


app/serializers/api/v2/headings/heading_serializer.rb
```diff
@@ -8,7 +8,14 @@ class HeadingSerializer
 
         set_id :goods_nomenclature_sid
 
-        attributes :goods_nomenclature_item_id, :description, :bti_url, :formatted_description
+        attributes :goods_nomenclature_item_id,
+                   :description,
+                   :bti_url,
+                   :formatted_description
+
+        attribute :declarable do
```
willfish, line 16: This serializer is used in v2 land when the heading is not declarable

app/serializers/api/v2/headings/declarable_heading_serializer.rb
```diff
@@ -8,8 +8,14 @@ class DeclarableHeadingSerializer
...
+        attributes :validity_start_date,
+                   :validity_end_date,
+                   :goods_nomenclature_item_id,
+                   :description,
+                   :bti_url,
+                   :formatted_description,
+                   :basic_duty_rate,
+                   :meursing_code
 
         attribute :declarable do
```
willfish, line 20: This is used only in the declarable case

willfish:
We don't have serializer specs for these currently and it's a palava to add them. Are you okay if I add them when we get back @jebw?

Well, the actual situation is:

1. Commodity serializer spec - yes
2. Heading serializer spec - no
3. Declarable heading serializer spec - no

jebw (approved):
LGTM

PR #199, opened 2021-09-07
Fix for failing spec with hard coded date

jebw:
### Jira link

HOTT-???

### What?

I have added/removed/altered:

- [x] Fixed a failing spec - it had a hard coded date so specs passed on the day it was merged but not afterwards

### Why?

I am doing this because:

- Deploys rely on passing specs


alexdesi (approved):
It looks good!

PR #201, opened 2021-09-09
HOTT-811 Adds the field declarable to the search serializers

alexdesi:
### Jira link

HOTT-811

### What?
The reason is that we need to keep declarable in case of fuzzy search too,
which came from elasticsearch.

I have added/removed/altered:
- [x] Added declarable in the search heading and commodities serializers
- [x] Add the missing tests for those

### Why?

I am doing this because:
to help the SWT to check if an entity is declarable.

### Have you? (optional)

not yet!

Risk should be low, since we are adding a new field (and not removing any)

### DIFF
--- a/app/elastic_search_indexes/search/chapter_index.rb
+++ b/app/elastic_search_indexes/search/chapter_index.rb
@@ -20,7 +20,8 @@ module Search
                 title: { type: 'text' },
                 numeral: { type: 'keyword' }
               }
-            }
+            },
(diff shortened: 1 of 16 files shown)


willfish (approved):

alexdesi:
Allowing the STW team to call the FE endpoint is not a good future-proof implementation,
so we decided on exposing a similar implementation on the BE (to be implemented).
That's the reason we are closing this PR, with no merge.

PR #202, opened 2021-09-14
HOTT-930: Spec to catch when the CDS import failed

jebw:
Failure was caused by missing load of entities by entity mapper. Check to ensure the list of MAPPERS is not empty.

### Jira link

[HOTT-930](https://transformuk.atlassian.net/browse/HOTT-930)

### What?

I have added/removed/altered:

- [x] Added a spec to ensure the CDS importer has some entities to map

### Why?

I am doing this because:

- When the zeitwerk autoloader was enabled it broke autoloading of the entity mappers. This was not picked up because the specs mocked the list of entity mappers to use.
- So I've added a spec to ensure when the list isn't mocked, it is still not empty.


jebw:
Merging this PR since the same change on a broken branch was approved - see #203

PR #207, opened 2021-09-15
Revert "Surface declarable on the heading and commodity"

alexdesi:
### Jira link

HOTT-811

### What?

This PR reverts the surfacing of the declarable on the heading and commodity

### Why?
We need to re-think the implementation. Allowing to call a FE endpoint is a bad and not forward-thinking solution.


willfish:
Closing after chat

PR #208, opened 2021-09-16
wait for migrations to complete

mafonso:
### Jira link

HOTT-n/a

### What?

I have added/removed/altered:

- [ ] Upgraded CI's CF cli to 7.3.0
- [ ] --wait on migration task

### Why?

I am doing this because:

- (migration) tasks were just being submitted and actual execution status was not checked.
-
-

### Have you? (optional)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,7 @@ commands:
       - run:
           name: "Run Migrations"
           command: |
-            cf run-task "tariff-<< parameters.service >>-backend-worker-<< parameters.environment_key >>" --command "rake db:migrate" --name "db-migrate" --wait
+            cf run-task "tariff-<< parameters.service >>-backend-worker-<< parameters.environment_key >>" --command "cd app && rake db:migrate" --name "db-migrate" --wait
 
   sentry-release:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ commands:
       - run:
           name: "Setup CF CLI"
           command: |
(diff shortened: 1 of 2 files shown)


willfish (approved):

.circleci/config.yml
```diff
@@ -134,7 +134,7 @@ commands:
       - run:
           name: "Run Migrations"
           command: |
-            cf run-task "tariff-<< parameters.service >>-backend-worker-<< parameters.environment_key >>" --command "rake db:migrate" --name "db-migrate"
+            cf run-task "tariff-<< parameters.service >>-backend-worker-<< parameters.environment_key >>" --command "cd app && rake db:migrate" --name "db-migrate" --wait
```
jebw, line 137: The `cd app` part isn't right solution - the underlying issue is the Dockerfile should be changed to put the app in the correct location (`/app` instead of `/app/app`)

I'd suggest remove the cd part, push to the PR - we can then verify it fails to deploy (because migrations step _should_ fail at this point). This is the important bit because it means we know the migrations step will error in CI if developers add invalid migrations by mistake in the future.

Once we know the PR's CI run is failing in the manner we expect (its failing on the migrations step) - then add the missing `/` in the Dockerfile to 'fix' the PR?

jebw (changes requested):
I think we should be demonstrating the PR fails on the migrations step as a first part.

Then once thats demonstrated we can fix the running of migrations by moving the app to `/app` inside the container

.circleci/config.yml
mafonso: Yes, this commit was just a test, after suspecting a further commit targeting the Dockerfile  caused migrations to be out of sync. 
I probably should move this PR to draft again until this is sorted.

jebw (approved):
LGTM

PR #211, opened 2021-09-17
HOTT-810 Include search for singular/plural forms of the term.

alexdesi:
### Jira link
Jira: https://transformuk.atlassian.net/browse/HOTT-810

### What?

Include search for singular/plural forms of the term.

### Why?

This gives the user more flexibility, allowing us to search using either the singular or the plural form of the term.

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -1,7 +1,7 @@
 # SearchService is responsible for issuing search queries
 # it tries to do ExactSearch first in case search term consists
 # of numerical values. We can guess if that was a Section, Chapter,
-# Heading ir Commodity code and redirect the user straight there.
+# Heading or Commodity code and redirect the user straight there.
 #
 # Otherwise it issues FuzzySearch that searches for results inside
 # green pages (synonyms) index as well as the main goods nomenclature
@@ -15,7 +15,7 @@ class SearchService
   autoload :FuzzySearch, 'search_service/fuzzy_search'
   autoload :NullSearch,  'search_service/null_search'
 
-  INDEX_SIZE_MAX = 10000 # ElasticSearch does default pagination for 10 entries
+  INDEX_SIZE_MAX = 10_000 # ElasticSearch does default pagination for 10 entries
   # per page. We do not do pagination when displaying
   # results so have a constant much bigger than possible
   # index size for size value.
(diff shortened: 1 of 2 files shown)


app/services/search_service.rb
```diff
@@ -27,15 +27,14 @@ class SearchService
   class EmptyQuery < StandardError
   end
 
-  attr_accessor :q
-  attr_reader :result, :as_of, :data_serializer
+  attr_reader :q, :result, :as_of, :data_serializer
```
alexdesi, line 30: attr_accessor :q
create also the accessor writer `q=`
which is redundant since the method `q=` already exist.

app/services/search_service.rb
```diff
@@ -57,7 +56,7 @@ def q=(term)
     # if search term has no letters extract the digits
     # and perform search with just the digits (i.e., `no_alpha_regex`)
     # otherwise, ignore [ and ] characters to avoid range searches
-    @q = if m = cas_number_regex.match(term)
+    @q = if (m = cas_number_regex.match(term))
```
alexdesi, line 59: rubocop requirement

app/services/search_service.rb
```diff
@@ -88,8 +87,8 @@ def perform
     @result = if q.present?
                 ExactSearch.new(q, as_of).search!.presence ||
                   FuzzySearch.new(q, as_of).search!.presence
-              end
+              end || NullSearch.new(q, as_of)
 
-    @result ||= NullSearch.new(q, as_of)
+    @result
   end
```
alexdesi, line 93: rubocop requirement, it suggested to use 
"@perform= ..." which I do not like, 
I've done this change to avoid that, otherwise a bigger refactor would be required (which is a good idea, but not now)

app/services/search_service/exact_search.rb
```diff
@@ -80,16 +80,27 @@ def find_by_chemical(query)
...
         return gns.first if gns.length == 1
       end
     end
+
+    # Example:
+    # 'cookie' => ['cookie', cookies]
+    # 'leaves' => ['leaf', 'leaves']
+    def singular_and_plural(query)
+      if query == query.pluralize
+        [query.singularize, query]
```
willfish, line 100: Is it simpler just to do:

```ruby
[
  query.singularize, # singular form
  query.pluralize,  # plural form
  query  # original form for edge cases
]
```

willfish:
This is great/simple and I love it. Have you checked it out on the frontend and seen that it solves the fridge/fridges dilemma?

app/services/search_service/exact_search.rb
alexdesi: I like it! and we can add `.uniq` to avoid duplication

willfish (approved):

PR #212, opened 2021-09-20
HOTT 943: Add news items

jebw:
### Jira link

[HOTT-943](https://transformuk.atlassian.net/browse/HOTT-943)

### What?

I have added/removed/altered:

- [x] Added a News Item database table - this covers both XI and UK services in one table, we'll use the UK version and ignore the XI services version
- [x] Added data models with validation for News Items
- [x] Added an admin api for CRUD actions against the News Items - these endpoints are only added to the UK service, they are ignored on XI
- [x] Fixed an issue with creating mock users from the admin area in dev mode

### Why?

I am doing this because:

- We want to add a News Items feature - the stories will potentially span both services


willfish (approved):
Should we replicate some of the rubocop changes in other projects?

jebw:
Potentially - the irrelevant ActiveRecord cops are getting hit because the Sequel API has some similarities but is not the same (hence the hacks to the `User` model to make it compatible.

I suspect we are unlikely to hit them on other sites unless they are using ActiveModel derived models, but then I'd guess the ActiveModel api's will be compatible with the ActiveRecord ones anyway I expect.

jebw:
Note to self - looks like all the smoke tests are failing for some reason, investigate!

PR #213, opened 2021-09-20
HOTT-949: Updated RoO schemes files

jebw:
### Jira link

[HOTT-949](https://transformuk.atlassian.net/browse/HOTT-949)

### What?

I have added/removed/altered:

- [x] Updated the XI and UK schemes files
- [x] Added support for the additional attributes to `RulesOfOrigin::Scheme` model

### Why?

I am doing this because:

- We now have updated scheme files and there are additional attributes present in them


alexdesi (approved):
It looks good to me

PR #219, opened 2021-09-22
Default to v2 api locally

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Default to v2 in development and test modes
- [x] Default to v1 in production mode (same as before)
- [x] Remove default setting from the constraint instantiation in routes

### Why?

I am doing this because:

- This makes local development easier (we never use v1 and want to completely replace it eventually)


lib/api_constraints.rb
```diff
@@ -1,14 +1,27 @@
...
+  end
+
+  def development_default?
+    @version == 2
+  end
+
+  def production_default?
+    @version == 1
   end
 end
```
alexdesi, line 27: Are we sure we want that different based on behavior in prod/dev ?
it could bring to nasty surprises

It sounds like we could set an ENV var, for example `default_api_ver `, which we set only in local,  without leverage the dev/prod

alexdesi: maybe we would need only a change like this:
```
  def initialize(options)
    @version = options[:version]
    @default = options[:default]

    @default.merge(default: ENV('api_ver_default')) if ENV('api_ver_default')
  end
```

.env.development
```diff
@@ -1 +1,2 @@
 EXCHANGE_RATE_ACCESS_KEY=flibble
+DEFAULT_VERSION=2
```
willfish, line 2:
```suggestion
DEFAULT_API_VERSION=2
```

lib/api_constraints.rb
```diff
@@ -1,14 +1,19 @@
...
+  end
+
+  private
+
+  def default?
+    @version == default
+  end
+
+  def default
+    ENV.fetch('DEFAULT_VERSION', 1)
```
willfish, line 17:
```suggestion
    ENV.fetch('DEFAULT_API_VERSION', 1)
```

PR #220, opened 2021-09-22
Adjust autoscaling policy for all environments

mafonso:
### Jira link

HOTT-957

### What?

I have added/removed/altered:

- [ ] Adjusted the auto-scaler settings for all environments

In general the changes were:
-  scale up aggressively, and scale down in a softer way after the spike is gone.
- Drop the `throughput` metric, as it was useless (no data for it)

### Why?

I am doing this because:

- This morning's [blip](https://transformuk.atlassian.net/browse/HOTT-957) was likely caused but lack of resources on a flash spike in traffic. 
- This prompted an investigation, amongst other things, on the scaling policy. 

### Have you? 

We also want to adjust the memory footprint of each instance as we have it oversized. We want to be able to have more discrete instances with less memory each, and re-balance toward CPU usage instead. 

**Note: The autoscaler was actually already enabled for the backend in all environments, so this PR just adjusts it's settings.**

### DIFF
--- a/config/autoscaling/development-policy.json
+++ b/config/autoscaling/development-policy.json
@@ -1,22 +1,14 @@
 {
-  "instance_min_count": 1,
-  "instance_max_count": 3,
+  "instance_min_count": 2,
+  "instance_max_count": 6,
   "scaling_rules": [
     {
-      "metric_type": "throughput",
-      "breach_duration_secs": 120,
-      "threshold": 5,
-      "operator": "<",
-      "cool_down_secs": 120,
-      "adjustment": "-1"
-    },
-    {
-      "metric_type": "throughput",
-      "breach_duration_secs": 120,
-      "threshold": 5,
+      "metric_type": "memoryutil",
+      "breach_duration_secs": 60,
+      "threshold": 50,
       "operator": ">=",
-      "cool_down_secs": 120,
-      "adjustment": "+1"
+      "cool_down_secs": 60,
+      "adjustment": "+2"
     },
     {
       "metric_type": "memoryutil",
@@ -27,28 +19,20 @@
       "adjustment": "-1"
     },
     {
-      "metric_type": "memoryutil",
-      "breach_duration_secs": 120,
+      "metric_type": "cpu",
+      "breach_duration_secs": 60,
       "threshold": 50,
       "operator": ">=",
       "cool_down_secs": 120,
-      "adjustment": "+1"
+      "adjustment": "+2"
     },
     {
       "metric_type": "cpu",
-      "breach_duration_secs": 120,
-      "threshold": 60,
+      "breach_duration_secs": 60,
+      "threshold": 5,
       "operator": "<",
-      "cool_down_secs": 120,
+      "cool_down_secs": 60,
       "adjustment": "-1"
-    },
-    {
-      "metric_type": "cpu",
-      "breach_duration_secs": 120,
-      "threshold": 60,
-      "operator": ">=",
-      "cool_down_secs": 120,
-      "adjustment": "+1"
     }
   ]
 }
--- a/config/autoscaling/production-policy.json
+++ b/config/autoscaling/production-policy.json
@@ -1,54 +1,38 @@
 {
-  "instance_min_count": 5,
+  "instance_min_count": 8,
   "instance_max_count": 12,
   "scaling_rules": [
     {
-      "metric_type": "throughput",
+      "metric_type": "memoryutil",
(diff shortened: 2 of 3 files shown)


mafonso:
Here is a stress test run against dev, with the newer more reactive policies. 

```
dynamic          	succeeded     	4->3                 	2021-09-22T13:36:31+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	5->4                 	2021-09-22T13:34:31+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	3->5                 	2021-09-22T13:32:32+01:00     	+2 instance(s) because memoryutil >= 50% for 60 seconds
dynamic          	succeeded     	4->3                 	2021-09-22T13:30:31+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	2->4                 	2021-09-22T13:28:32+01:00     	+2 instance(s) because memoryutil >= 50% for 60 seconds
dynamic          	succeeded     	3->2                 	2021-09-22T13:26:31+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	1->3                 	2021-09-22T13:24:32+01:00     	+2 instance(s) because memoryutil >= 50% for 60 seconds
dynamic          	succeeded     	2->1                 	2021-09-22T13:22:32+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	3->2                 	2021-09-22T13:20:31+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	1->3                 	2021-09-22T13:18:31+01:00     	+2 instance(s) because memoryutil >= 50% for 60 seconds
dynamic          	succeeded     	2->1                 	2021-09-22T13:16:32+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	3->2                 	2021-09-22T13:15:31+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	1->3                 	2021-09-22T13:13:31+01:00     	+2 instance(s) because memoryutil >= 50% for 60 seconds
dynamic          	succeeded     	2->1                 	2021-09-22T13:11:31+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	3->2                 	2021-09-22T13:09:31+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	1->3                 	2021-09-22T13:07:31+01:00     	+2 instance(s) because memoryutil >= 50% for 60 seconds
dynamic          	succeeded     	2->1                 	2021-09-22T13:05:31+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
dynamic          	succeeded     	3->2                 	2021-09-22T13:03:31+01:00     	-1 instance(s) because memoryutil < 50% for 60 seconds
```

config/autoscaling/production-policy.json
```diff
@@ -1,54 +1,38 @@
 {
-  "instance_min_count": 5,
+  "instance_min_count": 8,
```
mafonso, line 2: Since we already have the policy applied, this just changes the minimum quantity to 8, so we can afford to scale down the memory per instance later. 

Fluctuation will be more aggressive going up, which should be safe, and softer on the scale down.
We can then look at the stats after a few day cycles and optimize for cost.

config/autoscaling/development-policy.json
```diff
@@ -27,28 +19,20 @@
...
     },
     {
       "metric_type": "cpu",
-      "breach_duration_secs": 120,
-      "threshold": 60,
+      "breach_duration_secs": 60,
+      "threshold": 5,
       "operator": "<",
-      "cool_down_secs": 120,
+      "cool_down_secs": 60,
```
mafonso, line 34: Dev is using very similar metrics to staging and prob with the following differences:

- Less instances overall (min and max) 
- More reactive to scale down to facilitate stress testing.

config/autoscaling/staging-policy.json
```diff
@@ -1,22 +1,14 @@
 {
-  "instance_min_count": 1,
-  "instance_max_count": 3,
+  "instance_min_count": 2,
+  "instance_max_count": 12,
```
mafonso, line 3: Staging has identical thresholds to production, but with a lower minimum instance limit.

PR #224, opened 2021-09-23
HOTT-954: Expose links data

jebw:
### Jira link

[HOTT-954](https://transformuk.atlassian.net/browse/HOTT-954)

### What?

I have added/removed/altered:

- [x] Added the Scheme links to the /rules_of_origin_schemes endpoint
- [x] Added the SchemeSet links to the /rules_of_origin_schemes endpoint

### Why?

I am doing this because:

- We want to show these links on the frontend


app/models/rules_of_origin/link.rb
```diff
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
+require 'digest'
+
 module RulesOfOrigin
   class Link
     include ActiveModel::Model
 
     attr_accessor :text, :url
+    attr_writer :id
```
willfish, line 10: Do you need this?

willfish (approved):

willfish:
Are deploys broken because of the partial ECR move?

app/models/rules_of_origin/link.rb
```diff
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
+require 'digest'
```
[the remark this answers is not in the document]
jebw: Core part of ruby so the autoloader wont handle it

If something else is loading it already, it'll be ignored - if its not then its required. Same situation as when you use the CSV features.

app/models/rules_of_origin/link.rb
jebw: It does get used as part of the specs I think, and seems generally useful

PR #225, opened 2021-09-23
HOTT-962: Update rules data

jebw:
### Jira link

[HOTT-962](https://transformuk.atlassian.net/browse/HOTT-962)

### What?

I have added/removed/altered:

- [x] Updated the rules of origin data
- [x] Added check for duplicate rules
- [x] Expose the Alternate rule data in the API
- [x] Allow blank rules in the validator

### Why?

I am doing this because:

- We require updated rules of origin data


app/models/rules_of_origin/rule_set.rb
```diff
@@ -5,7 +5,7 @@
 module RulesOfOrigin
   class RuleSet
     DEFAULT_SOURCE_PATH = Rails.root.join('db/rules_of_origin').freeze
-    DEFAULT_FILE = 'rules_of_origin_210728.csv'
+    DEFAULT_FILE = 'rules_of_origin_210923.csv'
```
willfish, line 8: Maybe this would be easier in an environment variable?

willfish (approved):
LGTM. Data in the repo creates so much noise

jebw:
Agreed, but we'd probably end up with the files in the repo anyway - they are big but not that big and shouldn't change that often and they gzip well because they're text. Pulling them in from external brings its own complications anyway.

app/models/rules_of_origin/rule_set.rb
jebw: Bit split on this - since the files are in the code base, controlling which file to use should probably be controlled within the codebase as well

PR #227, opened 2021-09-24
HOTT-940 Fix and extend the Search for quotas

alexdesi:
### What?

This PR Fixes and extends the Search for quotas

The search is smarter now, it supports:
- Chapter
- Heading
- Commodity
using the ancestries,
otherwise - in case the term does not look one of those - it uses a ilike search.

### Why?
I am doing this because:

To allow users to find easily quotas.

### Jira link
https://transformuk.atlassian.net/browse/HOTT-940

### DIFF
--- a/app/models/goods_nomenclature.rb
+++ b/app/models/goods_nomenclature.rb
@@ -10,7 +10,7 @@ class GoodsNomenclature < Sequel::Model
   plugin :nullable
   plugin :active_model
 
-  plugin :sti, class_determinator: ->(record) {
+  plugin :sti, class_determinator: lambda { |record|
     gono_id = record[:goods_nomenclature_item_id].to_s
 
     if gono_id.ends_with?('00000000')
@@ -30,10 +30,6 @@ class GoodsNomenclature < Sequel::Model
       .order(Sequel.desc(:goods_nomenclature_indents__validity_start_date))
   end
 
-  def goods_nomenclature_indent
-    goods_nomenclature_indents.first
-  end
-
   many_to_many :goods_nomenclature_descriptions, join_table: :goods_nomenclature_description_periods,
                                                  left_primary_key: :goods_nomenclature_sid,
                                                  left_key: :goods_nomenclature_sid,
@@ -43,20 +39,12 @@ class GoodsNomenclature < Sequel::Model
       .order(Sequel.desc(:goods_nomenclature_description_periods__validity_start_date))
   end
 
-  def goods_nomenclature_description
-    goods_nomenclature_descriptions.first || NullGoodsNomenclature.new
-  end
-
   many_to_many :footnotes, join_table: :footnote_association_goods_nomenclatures,
                            left_primary_key: :goods_nomenclature_sid,
                            left_key: :goods_nomenclature_sid,
                            right_key: %i[footnote_type footnote_id],
                            right_primary_key: %i[footnote_type_id footnote_id] do |ds|
     ds.with_actual(FootnoteAssociationGoodsNomenclature)
-  end
-
-  def footnote
-    footnotes.first
   end
 
   one_to_one :national_measurement_unit_set, key: :cmdty_code,
@@ -105,6 +93,18 @@ class GoodsNomenclature < Sequel::Model
     end
   end
 
+  def goods_nomenclature_indent
+    goods_nomenclature_indents.first
+  end
+
+  def goods_nomenclature_description
+    goods_nomenclature_descriptions.first || NullGoodsNomenclature.new
+  end
+
+  def footnote
+    footnotes.first
+  end
+
   def id
     goods_nomenclature_sid
   end
@@ -127,5 +127,16 @@ class GoodsNomenclature < Sequel::Model
 
   def bti_url
     'https://www.gov.uk/guidance/check-what-youll-need-to-get-a-legally-binding-decision-on-a-commodity-code'
+  end
+
+  # This method is safer than `constantize`
+  # which is considered dangeros by `brakeman` (a static analyzer)
+  def self.constantize_class_name(klass_name)
+    [
+      Commodity,
+      Heading,
+      Chapter,
+      GoodsNomenclature,
+    ].find { |klass| klass.to_s == klass_name }
   end
 end
--- a/app/services/find_ancestors_service.rb
+++ b/app/services/find_ancestors_service.rb
@@ -1,0 +1,39 @@
+class FindAncestorsService
+  def initialize(goods_nomenclature_item_id)
+    @goods_nomenclature_item_id = goods_nomenclature_item_id
+  end
+
+  def call
+    return [] if goods_nomenclature_item_id.blank?
+
+    goods_nomenclature.uptree.map(&:goods_nomenclature_item_id)
+  end
+
+  private
+
+  def goods_nomenclature
+    goods_nomenclature_class.by_code(goods_nomenclature_item_id).take
+  end
+
+  def goods_nomenclature_class
+    klass_name = GoodsNomenclature.class_determinator
+                                  .call(goods_nomenclature_item_id: goods_nomenclature_item_id)
+
+    GoodsNomenclature.constantize_class_name(klass_name)
+  end
+
+  def goods_nomenclature_item_id
+    length = @goods_nomenclature_item_id.length
+
+    case length
+    when 2 # chapter
+      "#{@goods_nomenclature_item_id}00000000"
+    when 4 # heading
+      "#{@goods_nomenclature_item_id}000000"
+    when 10 # fully-qualified
+      @goods_nomenclature_item_id.to_s
+    else # ilike
+      ''
+    end
+  end
+end
--- a/spec/services/find_ancestors_service_spec.rb
+++ b/spec/services/find_ancestors_service_spec.rb
@@ -1,0 +1,53 @@
+describe FindAncestorsService do
+  subject(:service) { described_class.new(goods_nomenclature_item_id).call }
+
+  context 'when the search term is empty' do
+    let(:goods_nomenclature_item_id) { '' }
+
+    it 'returns no ancestors' do
+      expect(service).to eq([])
+    end
+  end
+
+  context 'when the search term is fully-qualified (10 chars)' do
+    let(:commodity) { create(:commodity) }
+    let(:goods_nomenclature_item_id) { commodity.goods_nomenclature_item_id }
+
+    it 'returns the correct ancestors' do
+      expected_result = commodity.uptree.pluck(:goods_nomenclature_item_id)
+
+      expect(service).to eq(expected_result)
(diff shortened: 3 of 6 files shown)


lib/api_constraints.rb
```diff
@@ -1,19 +1,14 @@
 class ApiConstraints
-  def initialize(version:)
-    @version = version
+  def initialize(options)
```
jebw, line 2: This changes the method signature from always requiring a version, to the version being optional - I'm not sure if thats a good idea - it could instead be 

```
def initialize(version:, default: nil)
  @version = version
  @default = default
end
```

In general I think changing the API behaviour is a separate thing from this ticket should be in its own PR and reviewed independently

spec/services/quota_search_service_spec.rb
```diff
@@ -1,3 +1,4 @@
+# rubocop:disable RSpec/MultipleMemoizedHelpers
```
jebw, line 1: I appreciate this isn't your code but lots of memoized helpers is a bit of a code smell - I think either we should be deciding to ignore this for now (have have lint fail - allowing the next developer to revisit it) or fixing the issue.

Is it worth adding either extending the factories to allow easily building the desired scenario, or just adding a `shared_context` block to build up the desired data?

jebw (changes requested):
Looks great but I think the Api constraints changes should be spun out into their own PR. 

Does the calling of `Commodity#ancestors` end up with an N+1 ?

jebw:
FYI: If you rebase onto master, that'll fix the CI errors on deploy

spec/services/quota_search_service_spec.rb
alexdesi: yeah, all those lets make the code a bit smelly, I'll see how we can reduce those.

app/services/find_ancestors_service.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def goods_nomenclature
+    goods_nomenclature_class.by_code(goods_nomenclature_item_id).take
+  end
+
+  def goods_nomenclature_class
+    klass_name = GoodsNomenclature.class_determinator
+                                  .call(goods_nomenclature_item_id: goods_nomenclature_item_id)
+
+    StringToClass.convert(klass_name)
```
willfish, line 22: Does constantize not work?

lib/string_to_class.rb
```diff
@@ -0,0 +1,11 @@
+module StringToClass
+  ALLOWED_CLASSES = [Chapter, Heading, Commodity].freeze
```
willfish, line 2: This is wayyyyy to defensive. We control the factory definition and if a person has access to commit changes to that we've got significant and various other problems

lib/api_constraints.rb
willfish: I've already made this change @alexdesi (see master)

willfish: We default from the value in the environment https://github.com/trade-tariff/trade-tariff-backend/pull/219

spec/services/find_ancestors_service_spec.rb
```diff
@@ -0,0 +1,63 @@
+describe FindAncestorsService do
+  subject(:find_ancestors) { described_class.new(goods_nomenclature_item_id).call }
```
willfish, line 2:
```suggestion
  subject(:service) { described_class.new(goods_nomenclature_item_id).call }
```
See
```
spec/services/applicable_vat_options_service_spec.rb|2 col 3| subject(:service) { described_class.new(measures) }
spec/services/cached_geographical_area_service_spec.rb|2 col 3| subject(:service) { described_class.new(actual_date, countries) }
spec/services/cached_commodity_service_spec.rb|2 col 3| subject(:service) { described_class.new(commodity.reload, actual_date, filter_params) }
spec/services/quota_search_service_spec.rb|2 col 3| subject(:service) { described_class.new(filter, current_page, per_page) }
spec/services/applicable_additional_code_service_spec.rb|2 col 3| subject(:service) { described_class.new(measures) }
spec/services/exchange_rate_service_spec.rb|2 col 3| subject(:service) { described_class.new }
spec/services/measure_unit_service_spec.rb|2 col 3| subject(:service) { described_class.new(measures) }
```

spec/services/find_ancestors_service_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  context 'when the search term is empty' do
+    let(:goods_nomenclature_item_id) { '' }
+
+    it 'returns no ancestors' do
+      find_ancestors
```
willfish, line 8: I don't think you need this line?

spec/services/find_ancestors_service_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      expect(find_ancestors).to eq([])
+    end
+  end
+
+  context 'when the search term is Fully qualified (10 chars)' do
```
willfish, line 14:
```suggestion
  context 'when the search term is fully qualified (10 chars)' do
```

spec/services/find_ancestors_service_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    let(:commodity) { create :commodity }
+    let(:goods_nomenclature_item_id) { commodity.goods_nomenclature_item_id }
+
+    it 'returns the correct ancestors' do
+      find_ancestors
```
willfish, line 19: Do you need the service to execute before you run specify the expected result?

willfish: Okay, I can see that you do

spec/services/find_ancestors_service_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    it 'returns no ancestors' do
+      find_ancestors
+
+      expect(find_ancestors).to eq([])
+    end
+  end
+
+  context 'when the search term is Fully qualified (10 chars)' do
+    let(:commodity) { create :commodity }
```
willfish, line 15: How do you feel about not mixing dsl/non-dsl styles for method invocations? Happy if you both prefer this, though

```suggestion
    let(:commodity) { create(:commodity) }
```

spec/services/quota_search_service_spec.rb
```diff
@@ -52,14 +53,22 @@
   end
 
   describe '#call' do
-    context 'when filtering by the goods_nomenclature_item_id' do
+    context 'when filtering by a fully qualified goods_nomenclature_item_id' do
```
willfish, line 56:
```suggestion
    context 'when filtering by a fully-qualified goods_nomenclature_item_id' do
```

spec/services/quota_search_service_spec.rb
```diff
@@ -52,14 +53,22 @@
...
       let(:filter) { { 'goods_nomenclature_item_id' => measure1.goods_nomenclature_item_id } }
 
       it 'returns the correct quota definition' do
         expect(service.call).to eq([quota_definition1])
       end
     end
 
+    context 'when filtering by a NOT fully qualified goods_nomenclature_item_id' do
```
willfish, line 64:
```suggestion
    context 'when filtering by a NOT fully-qualified goods_nomenclature_item_id' do
```

spec/services/quota_search_service_spec.rb
```diff
@@ -87,7 +96,7 @@
     context 'when filtering by status exhausted' do
       let(:filter) { { 'status' => 'exhausted' } }
 
-      let!(:quota_exhaustion_event) do
+      before do
```
willfish, line 99: Love it!

willfish:
Re N + 1s @jebw  I believe the ancestors method is already immune to that (it pulls out indent and description relationships with an eager load.

spec/services/quota_search_service_spec.rb
willfish: Maybe do that in a separate PR @alexdesi and we can just accept that the linting doesn't pass for this file for now

alexdesi: Since it is not a fast thing, yes it's better to keep it disabled for now, so Madhu can start testing this on Monday

app/services/find_ancestors_service.rb
alexdesi: no, it is considered unsafe by brakman

lib/string_to_class.rb
alexdesi: that's because brakeman do not like it: see
https://stackoverflow.com/questions/23741259/brakeman-unsafe-reflection-method-constantize-called-with-model-attribute

anyway, 
I removed the module, and just using just a method now

The alternative would be to disable that check, but I guess it's better to stay safe and follow brakeman.

spec/services/find_ancestors_service_spec.rb
alexdesi: ok I'll change it,
but what do you think it's more readable
`find_ancestor` 
or
`service`?
😺

spec/services/find_ancestors_service_spec.rb
alexdesi: good catch 👍

app/services/find_ancestors_service.rb
willfish: That's a bit nuts, really. It is safe

lib/string_to_class.rb
willfish: I think this should move to the class_determinator

spec/services/find_ancestors_service_spec.rb
willfish: When it comes to the subject under test I normally just look at the subject description (class name) at the top of the file. It's a bit like using `described_class`  the fully qualified name versus `FindAncestors`. I guess the service pattern is so common it makes sense to just follow a standard that codes to the interface and not to this specific implementation.

app/models/goods_nomenclature.rb
```diff
@@ -128,4 +128,15 @@ def code
   def bti_url
     'https://www.gov.uk/guidance/check-what-youll-need-to-get-a-legally-binding-decision-on-a-commodity-code'
   end
+
+  # This method is safer than `constantize`
+  # which is considered dangeros by `brakeman` (a static analyzer)
+  def self.constantize_class_name(klass_name)
```
willfish, line 134: We can just call `GoodsNomenclature.sti_load(klass_name)` and remove this

willfish, line 134: This feels really wrong :thinking:

willfish:
```ruby
GoodsNomenclature.sti_load(goods_nomenclature_item_id: '1234000000').class => Heading
```

willfish: I think we can get rid of this and just use the methods in the Sti plugin

willfish (approved):

PR #228, opened 2021-09-24
Lazy load all measure v2 relationships

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-801

### What?

I have added/removed/altered:

- [x] Added lazy_load_data: true to all v2 measure serializer relationships

### Why?

I am doing this because:

- This change means that other users of the serializer do not have to load the entire world of un-eager loaded resources just to return a subset of the targeted relationships.
- This will make the meursing index endpoints which serialize meursing measures as measures signficantly faster and should not affect existing implementations


jebw (approved):
LGTM

PR #234, opened 2021-09-27
Replacing constantize with sti_load

alexdesi:
Make use of sti_load to constantize the good nomenclature class name.

Why?
The method String#constantize is considered unsafe by Brakeman.

### DIFF
--- a/app/services/find_ancestors_service.rb
+++ b/app/services/find_ancestors_service.rb
@@ -16,10 +16,7 @@ class FindAncestorsService
   end
 
   def goods_nomenclature_class
-    klass_name = GoodsNomenclature.class_determinator
-                                  .call(goods_nomenclature_item_id: goods_nomenclature_item_id)
-
-    GoodsNomenclature.constantize_class_name(klass_name)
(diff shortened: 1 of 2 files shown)


config/brakeman.yml
```diff
@@ -0,0 +1,3 @@
+---
```
willfish, line 1: Why not extend config/brakeman.ignore?

app/services/find_ancestors_service.rb
```diff
@@ -19,7 +19,7 @@ def goods_nomenclature_class
     klass_name = GoodsNomenclature.class_determinator
```
willfish, line 19:
```suggestion
    klass_name = GoodsNomenclature.sti_class(goods_nomenclature_item_id: goods_nomenclature_item_id).class
```

config/brakeman.yml
willfish: This file is not needed

alexdesi: yeah, that's too loose security-wise.

willfish (approved):

PR #237, opened 2021-09-27
Surface resolved duty expressions when passed a meursing code

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-801

### What?

I have added/removed/altered:

- [x] Surface meursing measure resolutions when the user specifies a meursing additional code id
- [x] Adds service for finding meursing measures
- [x] Adds service for resolving meursing measures
- [x] Adds presenter to decorate the formatted duty expression of a resolved meursing component with working duty expression id, formatting and abbreviations
- [x] Adds specs for all of the above

### Why?

I've recently added a tool to the online trade tariff which lets us find our meursing code for our input answers to starch, sucrose, milk fat and milk protein. We want to display the resolved duty expressions once we know our meursing code in the import tabs of both headings and commodities. In order to display the resolved components we need some mechanism for resolving them in the frontend or the backend. Since everything is already loaded in the backend I've opted to put the resolution of this in the backend and have the resolved duty expression surfaced on the measure.

### AC

- [x] Non-meursing calls to the api are not affected by the queries this resolution process generates
- [x] The api is performant for multiple measures


config/application.rb
```diff
@@ -16,7 +16,7 @@ class Application < Rails::Application
 
     config.eager_load_paths << Rails.root.join('lib')
 
-    config.autoloader = :classic
+    # config.autoloader = :classic
```
willfish, line 19: Revert me

app/models/meursing_measure.rb
```diff
@@ -0,0 +1,13 @@
+class MeursingMeasure < Measure
+  initial_dataset = where(goods_nomenclature_item_id: nil).where(measure_type_id: MeasureType::MEURSING_MEASURES).where(additional_code_type_id: AdditionalCode::MEURSING_TYPE_IDS)
```
jebw, line 2: Wrapping some of the `.wheres` onto new lines would ease readability of this?

app/controllers/api/v2/meursing_measures_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+          .actual
+          .eager(
+            :additional_code,
+            :geographical_area,
+            :measure_components,
+            :measure_type,
+            measure_components: [:duty_expression],
+          )
+          .all
+          .select(&:current?)
```
jebw, line 53: Does this need `current?` - shouldn't the `.actual` filter by timestamp? If not, ideally it would be better to add the filtering to the db query rather than filtering the returned dataset?

spec/models/meursing_measure_spec.rb
```diff
@@ -0,0 +1,27 @@
+RSpec.describe MeursingMeasure do
+  it { expect(described_class.primary_key).to eq(:measure_sid) }
+
```
jebw, line 3: `MeursingMeasure` is a filter onto an existing data set - it could do with some specs for the `where` clauses to check whats include/excluded?

jebw (changes requested):
Looks good - I think it could do with the specs adding for the dataset `where clauses` to check its behaving as expected.

app/controllers/api/v2/meursing_measures_controller.rb
willfish: Spoke about this offline. So I've inherited a broken TimeMachine mechanism from the Measure model. This wasn't noticed previously but was noticed for MeursingMeasure rows since most rows inherit their validity_end_date from the some generating regulation. The business logic for this is complex, needs fixing globally and wouldn't add an unnoticeable benefit performance wise for the complexity of writing the explicit query to do the logic and joins in the database.

I'm going to leave this until we approach the fix for the measures model

jebw (approved):
LGTM

spec/formatters/duty_expression_formatter_spec.rb
```diff
@@ -6,25 +6,14 @@
...
-    let!(:measurement_unit_abbreviation) do
+    let(:measurement_unit_abbreviation) do
       create(:measurement_unit_abbreviation, :with_measurement_unit, :include_qualifier)
     end
-    let!(:measurement_unit_qualifier) do
+    let(:measurement_unit_qualifier) do
       create(:measurement_unit_qualifier, measurement_unit_qualifier_code: measurement_unit_abbreviation.measurement_unit_qualifier)
     end
 
-    context 'for excise measure' do
```
willfish, line 16: We don't automatically convert formatted duty expressions anymore

app/formatters/duty_expression_formatter.rb
```diff
@@ -19,10 +19,14 @@ def format(opts = {})
       measurement_unit_qualifier = opts[:measurement_unit_qualifier]
       measurement_unit_abbreviation = measurement_unit.try :abbreviation,
                                                            measurement_unit_qualifier: measurement_unit_qualifier
+      resolved_meursing_component = opts[:resolved_meursing]
```
jebw, line 22: passing thought - would it make more sense to use named params instead of an opts hash? More control over which (if any) are required then

app/formatters/duty_expression_formatter.rb
```diff
@@ -19,10 +19,14 @@ def format(opts = {})
...
+      formatted = opts[:formatted]
+
+
```
jebw, line 25: super minor _but_ - additional whitespace

app/formatters/duty_expression_formatter.rb
```diff
@@ -77,14 +81,21 @@ def format(opts = {})
           output << monetary_unit
         end
         if measurement_unit_abbreviation.present?
-          output << if opts[:formatted]
+          output << if formatted
                       "/ <abbr title='#{measurement_unit.description}'>#{measurement_unit_abbreviation}</abbr>"
```
jebw, line 85: could use `helper.tag.abbr(measurement_unit_abbreviation, title: measurement_unit.description)` - handles HTML escaping itself then rather than the implicit trust of date with `.html_safe`

app/formatters/duty_expression_formatter.rb
```diff
@@ -77,14 +81,21 @@ def format(opts = {})
...
                       "/ #{measurement_unit_abbreviation}"
                     end
         end
       end
-      output.join(' ').html_safe
+
+      result = output.join(' ').html_safe
+
+      if resolved_meursing_component
+        "<strong>#{result}</strong>"
```
jebw, line 95: `helper.tag.strong(result)` ?

app/services/meursing_measure_component_resolver_service.rb
```diff
@@ -0,0 +1,34 @@
+class MeursingMeasureComponentResolverService
```
jebw, line 1: Is it worth namespacing these? `MeursingMeasure::ComponentServiceResolver`

jebw: Just to group related services together?

app/services/meursing_measure_finder_service.rb
```diff
@@ -0,0 +1,22 @@
+class MeursingMeasureFinderService
```
jebw, line 1: `MeursingMeasure::FinderService` ?

app/serializers/api/v2/measures/measure_serializer.rb
```diff
@@ -8,8 +8,23 @@ class MeasureSerializer
...
+                   :excise,
+                   :vat,
+                   :reduction_indicator,
+                   :meursing
+
+        attribute :resolved_duty_expression do |root_measure, params|
+          additional_code_id = params[:meursing_additional_code_id]
+
+          if additional_code_id.present?
+            root_measure.resolved_duty_expression_for(additional_code_id)
```
jebw, line 25: If I'm reading this correctly - this introduces an N+1

The MeasureSerializer will be run _once per measure_ 
- if there is an `additional_code_id.present?` 
- call `resolved_duty_expression_for`
- which calls `meursing_measures_for`
- if its a meursing measure, it then calls `MeursingMeasureFinderService`
- which in then performs a DB query?

app/formatters/duty_expression_formatter.rb
```diff
@@ -77,14 +81,21 @@ def format(opts = {})
...
+      else
+        result
+      end
     end
```
alexdesi, line 99: and here is one of my annoying generic comments 😄  this method is getting longer and longer, maybe it is doing too many things? can we split his responsibility?

jebw (approved):
Couple of syntactic suggestions - only point I'd raise is I _think_ it introduces an N+1 - and triggering a DB query from the serializer feels somewhat opaque?

app/serializers/api/v2/measures/measure_serializer.rb
willfish: So it does but I'm not so worried about it for a couple of reasons:

1. It performs fine in real life testing for even really heavy measures
2. We only do this for a seriously small subset of commodity requests
3. The number of measures isn't set to grow to beyond the number of measures I've tested against in the worst case for the subset of measures I've dealt with
4. We have indexes for most of the filters we're using from my understanding

If we do see a performance hit because some of these changes we can eager load a sefl-referential join as part of the cached commodity service process and extend the list of indexes. I suspect we'll never need to do this, though

app/formatters/duty_expression_formatter.rb
willfish: I'm probably going to completely rewrite the duty expression formatter at some point and I'll include that view helper/clarify a bunch of the business logic

app/services/meursing_measure_component_resolver_service.rb
willfish: Don't get the value of this? To be able to easily remove the meursing measure services?

app/formatters/duty_expression_formatter.rb
willfish: :+1:

app/formatters/duty_expression_formatter.rb
willfish: I agree. I actually think we pull most of these things from the component and could do with passing just the component plus some options (like formatted and resolved_meursing). We'll want to ask the component what it's duty expression is and under which category of duty expression it sits in future for example.

DutyExpression#compound?

app/formatters/duty_expression_formatter.rb
willfish: Yarp, one to be completely refactored

willfish:
Anyway, just to indicate some of my thinking around the duty expression formatter:

1. It codifies a lot of business logic that could be moved to the measure component and measure component's duty expression as dedicated business logic methods. This would clarify the meaning of arbitrary ids
2. It has a lot of responsibilities  for printing lots of different kinds of duties that might actually be better separated into typed duty expression formatters (e.g. CompoundDutyExpressionFormatter) which are individually picked by a duty expression factory.  This way we can clear most of the questions that the formatter has to ask and make it really clear what type of thing is getting printed.
3. The format method gets driven directly in a lot of cases but can also be mixed in (we should handle this better and just use one method)
4. The specs aren't necessarily exhaustive and miss quite a few scenarios and over test others

PR #240, opened 2021-09-29
Rename format -> custom_format

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Rename format -> custom_format in Formatter

### Why?

I am doing this because:

- We don't want to override a ruby or rails core methods that aren't attached clearly
to a constant (e.g. DescriptionFormatter.format)

### DIFF
--- a/app/models/concerns/formatter.rb
+++ b/app/models/concerns/formatter.rb
@@ -2,19 +2,21 @@ module Formatter
   extend ActiveSupport::Concern
 
   module ClassMethods
-    def format(attribute, options)
+    def custom_format(attribute, options)
       options.assert_valid_keys :with, :using, :defaults
 
-      formatter, using, defaults = options.values_at(:with,
-                                                     :using,
-                                                     :defaults)
+      formatter = options[:with]
+      using = options[:using]
+
       mod = Module.new do
         define_method(attribute) do
           opts = {}
 
-          [using].flatten.each do |field|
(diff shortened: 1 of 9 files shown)


app/models/concerns/formatter.rb
```diff
@@ -2,19 +2,21 @@ module Formatter
...
+      using = options[:using]
+
       mod = Module.new do
         define_method(attribute) do
           opts = {}
 
-          [using].flatten.each do |field|
-            opts[field] = result_of_attribute_or_method_call(field)
-          end if using.present?
+          if using.present?
```
alexdesi, line 15: yes, that's more readable 👍

app/models/concerns/formatter.rb
```diff
@@ -2,19 +2,21 @@ module Formatter
...
-    def format(attribute, options)
+    def custom_format(attribute, options)
       options.assert_valid_keys :with, :using, :defaults
 
-      formatter, using, defaults = options.values_at(:with,
-                                                     :using,
-                                                     :defaults)
+      formatter = options[:with]
+      using = options[:using]
+
```
alexdesi, line 10: I guess we did not need defaults

alexdesi (approved):
nice one!

app/models/concerns/formatter.rb
willfish: Weirdly not

PR #242, opened 2021-09-30
Extend list of clearable models

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-992

### What?

I have added/removed/altered:

- [x] Extended the list of clearable models to include every model that uses the time machine

### Why?

We have issues with the association cache (which caches the result of a
relational query on the class singleton for the model).

We need to make sure that this cache is fresh for every thread (requests
execute inside of a single puma thread) so that we avoid collisions in
which the wrong start and end dates are used for related objects as read out from the association cache.

We have a need to review the time machine plugin since it doesn't marry
well with sequel currently.

This change means that we stop the caching issue and have time for this review.


jebw (approved):
LGTM

PR #253, opened 2021-10-11
HOTT-1008 - Surface resolved measure components

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1008

### What?

I have added/removed/altered:

- [x] Surface resolved measure components when the meursing additional code is passed to the commodities api
- [x] Enable block-fetched resolved measure components to have their relationship rendered with an id method override

### Why?

I am doing this because:

- This is needed to perform duty calculations on measures with a meursing additional code answer


spec/serializers/api/v2/measures/measure_serializer_spec.rb
```diff
@@ -57,5 +58,119 @@
...
+        create(
+          :measure_component,
+          duty_amount: 0.0,
+          monetary_unit_code: 'EUR',
+          measurement_unit_code: 'DTN',
+          measure_sid: meursing_agricultural_measure.measure_sid,
+        )
+      end
+
+      let(:expected_pattern) do
```
alexdesi, line 105: I think we can use symbols notation instead of string here

alexdesi (approved):
It looks good to me

PR #257, opened 2021-10-14
HOTT-1031: Updated schemes file with corrected url

jebw:
### Jira link

[HOTT-1031](https://transformuk.atlassian.net/browse/HOTT-1031)

### What?

I have added/removed/altered:

- [x] Updated the Rules of Origin schemes file with one with an updated proofs URL

### Why?

I am doing this because:

- The link was previously pointing to google


alexdesi (approved):
👍 approved, but  I've just noticed that the link:
https://www.gov.uk/guidance/get-proof-of-origin-for-your-goods#generalised-scheme-of-preferences-form-a
does not scroll to the internal anchor title "Generalised Scheme of Preferences form A"

PR #259, opened 2021-10-15
HOTT-1036: Cache whether a commodity has any children

jebw:
### Jira link

[HOTT-1036](https://transformuk.atlassian.net/browse/HOTT-1036)

### What?

I have added/removed/altered:

- [x] Cache whether a commodity has children for a particular date

### Why?

I am doing this because:

- We currently check for whether a commodity has children outside of the cached commodity service, on commodity code 2204219319 this is resulting in hundreds of SQL queries

### Notes

I am including the `actual_date`, ie the TimeMachine date in the cache key since its conceivable (but improbable) that a commodity may change whether it has children as a result of data updates

### DIFF
--- a/app/controllers/api/v2/commodities_controller.rb
+++ b/app/controllers/api/v2/commodities_controller.rb
@@ -26,7 +26,7 @@ module Api
                               .eager(:goods_nomenclature_indents, :goods_nomenclature_descriptions, :footnotes)
                               .take
 
-        raise Sequel::RecordNotFound if @commodity.children.any?
+        raise Sequel::RecordNotFound if commodity_has_children?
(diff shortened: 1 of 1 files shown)


app/controllers/api/v2/commodities_controller.rb
```diff
@@ -37,6 +37,14 @@ def cached_commodity
       def filter_params
         params.require(:filter).permit(:geographical_area_id, :meursing_additional_code_id) if params[:filter].present?
       end
+
+      def commodity_has_children?
+        cache_key = "commodity-#{@commodity.goods_nomenclature_sid}-#{actual_date}-has-children?"
+
+        Rails.cache.fetch(cache_key, expires_in: CachedCommodityService::TTL) do
+          @commodity.children.any?
+        end
```
alexdesi, line 46: should we expire this when a new child is added to the commodity?

PR #263, opened 2021-10-18
Remove Northern Ireland (XI) from XI service countries list.

alexdesi:
### Jira link
HOTT-983

### What?
Remove Northern Ireland (XI) from XI service countries list.

### DIFF
--- a/app/controllers/api/v2/geographical_areas_controller.rb
+++ b/app/controllers/api/v2/geographical_areas_controller.rb
@@ -6,7 +6,7 @@ module Api
       end
 
       def countries
-        render json: CachedGeographicalAreaService.new(actual_date, true).call
+        render json: CachedGeographicalAreaService.new(actual_date, countries: true).call
       end
     end
   end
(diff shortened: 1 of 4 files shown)


app/controllers/api/v2/geographical_areas_controller.rb
```diff
@@ -6,7 +6,7 @@ def index
       end
 
       def countries
-        render json: CachedGeographicalAreaService.new(actual_date, true).call
+        render json: CachedGeographicalAreaService.new(actual_date, countries: true).call
```
willfish, line 9: Good shout. More revealing

spec/services/cached_geographical_area_service_spec.rb
```diff
@@ -1,9 +1,9 @@
 RSpec.describe CachedGeographicalAreaService do
```
alexdesi, line 1: this time, I decided to fix the excess of lets, rubocop was right here 😄

PR #269, opened 2021-10-22
HOTT-1042: Add alerter for excess query counts

jebw:
### Jira link

[HOTT-1042](https://transformuk.atlassian.net/browse/HOTT-1042)

### What?

I have added/removed/altered:

- [x] Added a controller action to monitor for high query counts - 
- [x] This will either alert to Sentry in production mode
- [x] It will raise an exception in Test and Development modes
- [x] It doesn't run unless configured - the thinking is to enable this on staging, and developers can enable it locally if they want

### Why?

I am doing this because:

- We had a high number of queries occurring on our commodities page but nothing was alerting us to this fact
- The actual threshold we can tweak down over time but its about catching outlier pages with really high counts
-


app/controllers/concerns/check_query_count.rb
```diff
@@ -0,0 +1,48 @@
+module CheckQueryCount
+  class ExcessQueryCountException < StandardError; end
+
+  class CountChecker
+    MAX_QUERY_THRESHOLD = Rails.env.test? ? 30 : 300
```
alexdesi, line 5: in this case, it might be better to use the 
/config/environments/development.rb (production.rb and test.rb) to set that value:

```
config.max_query_threshold = 30 # in test.rb

config.max_query_threshold = 300  # in development.rb, and production.rb
```

spec/requests/concerns/check_query_count_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+           :with_description,
+           :declarable
+  end
+
+  let(:alerter) { described_class::CountChecker.new(true) }
+
+  context 'with request count under threshold' do
+    before { get_page }
+
+    it 'does not raise an exception' do
```
alexdesi, line 27: since we are not checking the exception in the test it sounds better to state:

`it 'returns success' do`

spec/requests/concerns/check_query_count_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      expect(response).to have_http_status :ok
+    end
+
+    it 'does not notify sentry' do
```
alexdesi, line 31: 👍

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,13 @@
 require 'clearable'
 
 class ApplicationController < ActionController::Base
+  include CheckQueryCount
+
   respond_to :json, :html
 
   before_action :clear_association_queries
   around_action :configure_time_machine
+  after_action  :check_query_count if TradeTariffBackend.check_query_count?
```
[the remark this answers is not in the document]
jebw: This was an optimisation, ie only add it to the controller if we're enabling it - rather then doing the check on every action _but_ it makes it hard to test so I think its probably better with a deferred check as you suggest

app/controllers/concerns/check_query_count.rb
jebw: Yeh - good shout - may make it configurable via an env var instead though

spec/requests/concerns/check_query_count_spec.rb
jebw: in this scenario we're testing for the alerter behaviour so I'd say no, what we care about is whether the alerter triggered an exception

willfish (approved):

PR #278, opened 2021-10-28
Refactor measure presenter

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Refactor measure presenter

### DIFF
--- a/app/presenters/api/v2/measures/measure_presenter.rb
+++ b/app/presenters/api/v2/measures/measure_presenter.rb
@@ -27,7 +27,7 @@ module Api
         end
 
         def additional_code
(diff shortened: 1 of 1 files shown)


jebw (changes requested):
Is it worth using `alias_method` for this - would shorten the delegation methods?

More signficantly it looks like there's a recursive call in the change so the specs are failing?

willfish:
Fixed the overflow

willfish:
Seems like alias_method works against self rather than __getobj__ or whatever the underlying method is so it KOs when you try

jebw (approved):

PR #283, opened 2021-11-02
HOTT-1082: Expose resolved measure component units

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1082

### What?

I have added/removed/altered:

- [x] Added resolved measure components to the list of components that express units
- [x] Made the logic of whether a component expresses units simpler
- [x] Added a request-global Thread variable for the meursing additional code that is accessible to serializers and presenters when trying to fetch measure units

### Why?

I am doing this because:

- This is required to make sure the user is asked the proper questions about resolved measure component units in the duty calculator

### DIFF
--- a/app/controllers/api/v2/commodities_controller.rb
+++ b/app/controllers/api/v2/commodities_controller.rb
@@ -2,6 +2,7 @@ module Api
   module V2
     class CommoditiesController < ApiController
       before_action :find_commodity, only: %i[show changes]
+      before_action :set_meursing_additional_code, only: :show
 
       def show
         render json: cached_commodity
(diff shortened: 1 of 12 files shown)


app/controllers/api/v2/commodities_controller.rb
```diff
@@ -35,7 +36,9 @@ def cached_commodity
       end
 
       def filter_params
-        params.require(:filter).permit(:geographical_area_id, :meursing_additional_code_id) if params[:filter].present?
+        return {} if params[:filter].blank?
+
+        params.require(:filter).permit(:geographical_area_id, :meursing_additional_code_id)
```
willfish, line 41:
```suggestion
        params.fetch(:filter, {}).permit(:geographical_area_id, :meursing_additional_code_id)
```

alexdesi: have you tried this change?

PR #286, opened 2021-11-05
HOTT-1096: Removes unused endpoints

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1096

### What?

I have added/removed/altered:

- [x] Removed authenticated endpoints that were replaced with the admin-scoped ones
- [x] Added TODO about strange authentication skip
- [x] Removed unused views

### Why?

I am doing this because:

- These are unused and make the backend feel more cluttered than it is


alexdesi (commented):
I think we need to update the routes.rb too

config/routes.rb
```diff
@@ -132,8 +132,7 @@
           get :tree
         end
         scope module: 'sections', constraints: { section_id: /\d{1,2}/, id: /\d+/ } do
-          resource :section_note, only: %i[show create update destroy]
-          resources :search_references, only: %i[show index destroy create update]
+          resource :section_note, only: %i[show]
```
alexdesi, line 135: 👍

alexdesi (approved):
It looks G🟢🟢D

PR #287, opened 2021-11-08
HOTT-880: Remove unused endpoints

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-880

### What?

I have added/removed/altered:

- [x] Removed unused endpoints

### Why?

I am doing this because:

- These were discovered as part of an investigation into the removal of unused authenticated endpoints in V1


.circleci/config.yml
```diff
@@ -297,14 +297,14 @@ jobs:
 
   smoketest_dev:
     docker:
-      - image: 'cypress/base:latest'
+      - image: 'cypress/base:16.5.0'
```
willfish, line 300: The latest tag for cypress/base hasn't been updated in a year (and therefore clearly isn't the latest)

willfish: I've updated this since it was out of step with the cypress test dependencies

jebw (approved):
Approving on the basis the NewRelic logs demonstrate these are unused?

alexdesi:
amazing cleaning!
just double check we do not forget factories, 
- tests (actually I don't think this, they'd fail)
- jsons used for tests
 - (rake) tasks only relevant to V1
 - routes (I think these are all right)

willfish:
@jebw New relic is actually how I enumerated these. I went through every v1 controller and searched for requests for each one in over a month

willfish:
We wouldn't use fixtures, here @alexdesi since we're in a position to be able to rely on factories that integrate with the db. Those factories are used in other places and are controller independent. I've looked for rake tasks and can't find any that run against these endpoints.

PR #288, opened 2021-11-08
Handle invalid/missing meursing additional code id

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1011

### What?

I have added/removed/altered:

- [x] Added handling for missing meursing additional code - reflected in no results being returned by the MeursingMeasureFinderService

### Why?

I am doing this because:

- This makes for a cleaner user experience when there is no meursing measures for the meursing additional code specified by the user

### DIFF
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -463,13 +463,13 @@ class Measure < Sequel::Model
   end
 
   def meursing_measures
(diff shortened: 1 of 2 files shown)


jebw (commented):
Looks good but the specs are failing?

jebw (approved):
LGTM

PR #292, opened 2021-11-09
Fix CVE-2015-9284

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-880

### What?

I have added/removed/altered:

- [x] Fixes [CVE-2015-9284](https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284)

### Why?

I am doing this because:

- Bundler audit was notifying us about this


willfish:
I've deployed this and was able to use the admin interface without causing a kerfuffle/breakages

jebw (approved):
LGTM

PR #294, opened 2021-11-11
HOTT-1052-surface-measurement-units

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1052

### What?

I have added/removed/altered:

- [x] Surface measurement units on the measure component serializer in the v2 commodities api

### Why?

I am doing this because:

- This is necessary to enable the work for the declarable context ticket


willfish:
:shipit:

PR #295, opened 2021-11-11
HOTT-1084: Migrate to using callbacks for specific behaviours on specific mappers

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1084

### What?

We are in a situation where, because of CDS non-compliance with
taric data structure and processing standards we need to special
snowflake a bunch of behaviour when specific kinds of entity are
imported.

This change makes this whole process a bit neater by supplying a before
insert callback mechanism which (as the name suggests) enables us to
implement root- and branch- specific behaviour before a given entity  has an operation row created.

Following this specific change will be more changes to the MeasureMapper
around FootnoteAssociationMeasures

I have added/removed/altered:

- [x] Added a callback mechanism for executing arbitrary callbacks before we insert a given mapped node

### Why?

I am doing this because:

- This is significantly neater than putting the callback code inline with the orchestrating layer and requires less logic

### DIFF
--- a/lib/cds_importer/entity_mapper.rb
+++ b/lib/cds_importer/entity_mapper.rb
@@ -21,11 +21,10 @@ class CdsImporter
                            .sort_by { |m| m.mapping_path ? m.mapping_path.length : 0 }
 
       mappers.each.with_object({}) do |mapper, oplog_inserts_performed|
-        remove_excluded_geographical_areas! if mapper == CdsImporter::EntityMapper::MeasureMapper
-
-        transform! if mapper == CdsImporter::EntityMapper::GeographicalAreaMembershipMapper
-
         instances = mapper.new(xml_node).parse
+
+        mapper.before_oplog_inserts_callbacks.each { |callback| callback.call(xml_node) }
+
         instances.each do |i|
           oplog_inserts_performed[i.operation_klass.to_s] ||= 0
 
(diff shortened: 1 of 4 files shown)


lib/cds_importer/entity_mapper.rb
```diff
@@ -21,11 +21,10 @@ def import
                            .sort_by { |m| m.mapping_path ? m.mapping_path.length : 0 }
 
       mappers.each.with_object({}) do |mapper, oplog_inserts_performed|
-        remove_excluded_geographical_areas! if mapper == CdsImporter::EntityMapper::MeasureMapper
+        instances = mapper.new(xml_node).parse
 
-        transform! if mapper == CdsImporter::EntityMapper::GeographicalAreaMembershipMapper
+        mapper.before_oplog_inserts.each { |callback| callback.call(xml_node) }
```
jebw, line 26: Can these have 'callback' in the name to clarify thats what they are, thanks

lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -11,6 +12,12 @@ class << self
         attr_accessor :entity_class, :entity_mapping, :mapping_path, :mapping_root, :exclude_mapping,
                       :entity_mapping_key_as_array, :entity_mapping_keys_to_parse
 
+        def before_oplog_inserts(&block)
```
jebw, line 15: Reading and writing via the same method feels confusing because they are used in different contexts - the write use case is intended for within the class whilst the read use case is external where there is less contextual information

Could this instead be

`def before_oplog_inserts` for creating the callback, this could be `private`?

`def before_oplog_insert_callbacks` for retrieving them - this would then be clearer at the callsite (ie EntityMapper)

jebw (commented):
This looks great, and its a much nicer approach. Only quibble is the double use case on the `before_oplog_inserts` method feels quite confusing - splitting the use cases up would allow for clearer naming as well?

lib/cds_importer/entity_mapper/base_mapper.rb
willfish: We could separate these, yeah

jebw (approved):
Looks good - nice refactor

PR #296, opened 2021-11-11
HOTT-1084: Remove footnote associations for every measure

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1084

### What?

I have added/removed/altered:

- [x] Added mapper callback for the Measure node which removes footnote associations which will get recreated by the related associations that exist on the node

### Why?

I am doing this because:

- This is required to handle the fact that CDS are not compliant with Taric processing standards and have decided that every Measure node is now a source of truth for all relationships.

### DIFF
--- a/spec/integration/cds_importer/entity_mapper_spec.rb
+++ b/spec/integration/cds_importer/entity_mapper_spec.rb
@@ -26,18 +26,49 @@ RSpec.describe CdsImporter::EntityMapper do
 
   describe '#import' do
     before do
-      stub_const(
-        'CdsImporter::EntityMapper::ALL_MAPPERS',
-        [
-          CdsImporter::EntityMapper::MeasureMapper,
-          CdsImporter::EntityMapper::MeasureExcludedGeographicalAreaMapper,
-          CdsImporter::EntityMapper::AdditionalCodeMapper,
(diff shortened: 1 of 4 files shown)


lib/cds_importer/entity_mapper/measure_mapper.rb
```diff
@@ -30,13 +30,11 @@ class MeasureMapper < BaseMapper
       self.entity_mapping_keys_to_parse = mapping_keys_to_parse.freeze
 
       before_oplog_inserts do |xml_node|
-        if xml_node['sid'].blank?
```
willfish, line 33: We've been warning about this and it's never happened. I thought this was defensive at the time and I do now. If the Measure sid is missing we won't soft-delete the associations or even be able to update the Measure record and it's other associations.

spec/integration/cds_importer/entity_mapper_spec.rb
```diff
@@ -26,20 +26,51 @@
 
   describe '#import' do
     before do
-      stub_const(
```
willfish, line 29: This isn't needed to get the tests to pass and misdirected me into thinking I need to change something I didn't

alexdesi (approved):
🟢 
just one minor comment

PR #297, opened 2021-11-12
HOTT-1113: Move to the zeitwerk autoloader

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1113

### What?

I have added/removed/altered:

- [x] Move to using the zeitwerk autoloader
- [x] Removes eager loading lib directory (this initializes db connections when running rake tasks) and enables zeitwerk to work with lib files by `git mv lib/ app/lib`
- [x] Now we have sensible zeitwerk loading move to rails to run migrations

### Why?

I am doing this because:

- We've done most of the migration of our AST load order to move away from classic and have tests of the ETL process now.
- Zeitwerk is better/requires fewer changes and makes us compatible with future versions of rails (upcoming rails 7)


Rakefile
```diff
@@ -5,12 +5,3 @@
 require_relative 'config/application'
 
 Rails.application.load_tasks
-
-if Rails.env.development?
```
willfish, line 9: This was causing some loading issues. Since we don't use this/it's currently broken I've gone and removed it

config/application.rb
```diff
@@ -1,4 +1,5 @@
 require_relative 'boot'
+require_relative '../app/lib/core_ext/object'
```
jebw, line 2: I think this does belong in `/lib`, and `/lib` in is the load path but not the auto load path - so can be loaded from `/lib` with `require 'core_ext/object'`

config/initializers/inflections.rb
```diff
@@ -1,16 +0,0 @@
-# Be sure to restart your server when you modify this file.
```
jebw, line 1: This file should stay - even though it is empty, it is a placeholder for if inflectors are required - an experienced ruby developer will just recreate the file but someone newer to the language might not realise they should. Same logic as using `.gitkeep` files

config/initializers/mime_types.rb
```diff
@@ -1,4 +0,0 @@
-# Be sure to restart your server when you modify this file.
```
jebw, line 1: Again this should stay as a a placeholder

config/initializers/new_framework_defaults.rb
```diff
@@ -1,20 +0,0 @@
-# Be sure to restart your server when you modify this file.
```
jebw, line 1: This file should be retained

This actually changes behaviours - the first two have no effect in an API app but the third may do. They are old defaults and should probably be dropped but in a separate PR since they do not relate to the switch to Zeitwerk

config/initializers/raven.rb
```diff
@@ -1,3 +1,5 @@
+require 'raven'
+
 if defined?(Raven) && ENV['VCAP_APPLICATION'].present?
```
jebw, line 3: I think rather then loading Raven - we always load it anyway, it's just the `defined?` is failing because of autoloading, both the `require 'raven'` and the `defined?(Raven)` should be removed

I think this is a hangover from when Raven was only loaded in production environment but I changed it to always load Raven a while back (since with out the env var it does nothing and it then allows testing alerts in the test suite)

jebw (changes requested):
This looks really good - it'll be nice to actually have all the importer classes being properly autoloaded. There's a couple of minor technical points I'd raise

1. Some of the initializers should stay - they are no-ops but they act as place holders and will only add microseconds on boot time.
2. The defaults initializer is old defaults and _should_ be being dropped but not as part of this PR since its unrelated to the switch to zeitwerk
3. The loading of Raven to allow the `defined?` check to pass works, but the actual check for Raven can be dropped instead

In general I'd suggest moving the importer related classes to `/app/importers` which more accurately describes them rather then the generic `/app/lib`

The other point is more subjective - personally I'm not convinced we need `/app/lib` - its normally a solution to `/lib` not being in the autoload path - but that is by design (`/lib` actually used to be in the autoload path and the rails team consciously swapped from that). Its intended for things that are used by the app but not really part of the app - so `require 'core_ext/object'` with `CoreExt::Object` being in `/lib` makes sense given that context?

jebw (approved):
LGTM

PR #299, opened 2021-11-16
Fix rake tasks

willfish:
### Jira link

HOTT-<TODO>

### What?

We depend on a bunch of models being loaded and haven't got explicit
requires that do this to enable doing various things with these models.

This includes things like unrestricting the primary key of each model so
the import process can update them or enumerating the models that are backed by oplogs
so we can clean oplog tables of rolled back files.

I've also updated a rake task not to try and load a constant and assign
a class variable to it before we've had a chance to load the environment

I have added/removed/altered:

- [x] Fix loading data migration rake tasks before the environment has loaded and the autoloaders are ready
- [x] Make it so that the Rakefile knows about the rake tasks under app/lib
- [x] When running tasks in development, make sure we're eager loading constants if we need them

### Why?

I am doing this because:

- So we can follow the common rails/zeitwerk idioms and not get confusing behaviour around rollbacks (I was getting confusing behaviour)
- So we can list rake tasks that - a bug I introduced when we moved to making lib work with zeitwerk under app/lib
- So we actually rollback successfully when debugging/testing ETL changes locally
- The main changes in this PR bring us up-to-date with how production is run

### DIFF
--- a/app/lib/tariff_synchronizer.rb
+++ b/app/lib/tariff_synchronizer.rb
@@ -34,7 +34,6 @@ module TariffSynchronizer
   mattr_accessor :taric_initial_update_date
   self.taric_initial_update_date = Date.new(2012, 6, 6)
 
-  # TODO:
   # set initial update date
   # Initial dump date + 1 day
   mattr_accessor :cds_initial_update_date
@@ -103,7 +102,7 @@ module TariffSynchronizer
     # running the apply task it is wrapped with a redis lock
     TradeTariffBackend.with_redis_lock do
       # Updates could be modifying primary keys so unrestricted it for all models.
-      Sequel::Model.subclasses.each(&:unrestrict_primary_key)
+      sequel_models.each(&:unrestrict_primary_key)
 
       date_range = date_range_since_last_pending_update
       date_range.each do |day|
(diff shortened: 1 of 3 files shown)


app/lib/tariff_synchronizer.rb
```diff
@@ -248,7 +249,7 @@ def rollback_cds(rollback_date, keep = false)
         end
       end
 
-      instrument('rollback.tariff_synchronizer', date: date, keep: keep)
+      # instrument('rollback.tariff_synchronizer', date: date, keep: keep)
```
willfish, line 252: Revert me

app/lib/tasks/data_migrations.rake
```diff
@@ -1,40 +1,43 @@
-TradeTariffBackend::DataMigrator.reporter = TradeTariffBackend::DataMigrator::ConsoleReporter
```
willfish, line 1: This is only visible when the environment is loaded

app/lib/tariff_synchronizer.rb
```diff
@@ -288,11 +289,19 @@ def sync_variables_set?
...
-    Sequel::Model.subclasses.select do |model|
+    sequel_models.select do |model|
       model.plugins.include?(Sequel::Plugins::Oplog)
     end
   end
 
+  def sequel_models
+    # Sequel::Model subclasses need to load into the ruby AST before they are visible
+    # This only affects running this code in development mode which does not eager load in the normal course of events
+    Rails.autoloaders.main.eager_load if Rails.env.development?
```
jebw, line 300: This is using `Rails.env.development?` to imply that eager loading is disabled, its better to be explicit and use that configuration value directly, eg `unless Rails.configuration.eager_load`

jebw (changes requested):
Looks great and makes good sense - only nitpick is using the environment to imply the eager loading configuration - whilst they are the same in the current config rails' defaults may change in the future so its more robust to explicitly test the config value.

jebw (approved):
Looks great

PR #301, opened 2021-11-17
Filter out excise measures on xi and refactor measure presenter

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1120

### What?

Also see: https://github.com/trade-tariff/trade-tariff-backend/pull/300

I have added/removed/altered:

- [x] Rename MeasurePresenter -> MeasureCollection and give it a collection/scope interface
- [x]  Update consumers of the MeasureCollection implementation
- [x] Adds filtering to the MeasureCollection class to exclude excise measures when running the xi service
- [x] Adds MeasureCollection spec

### Why?

I am doing this because:

- We're being bitten by the fact that we have excise measures from CHIEF that are no longer maintained and could do with being end dated at some stage. these measures should be filtered out in XI


app/controllers/api/v1/commodities_controller.rb
```diff
@@ -32,7 +32,7 @@ def show
             :full_temporary_stop_regulations,
             :measure_partial_temporary_stops
           ).all, @commodity
-        ).validate!
+        ).deduplicate.filter
```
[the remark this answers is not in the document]
alexdesi: what about renaming 
`deduplicate` to  `uniq`
shorter and simpler :)

spec/models/measure_collection_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  subject(:collection) { described_class.new(measures, declarable) }
+
+  let(:declarable) { build(:commodity) }
+
+  describe '#filter' do
+    before do
+      allow(TradeTariffBackend).to receive(:service).and_return(service)
+    end
+
+    let(:service) { 'uk' }
```
alexdesi, line 11: do we need this?

app/models/measure_collection.rb
```diff
@@ -0,0 +1,66 @@
...
+        measure.measure_type_id == THIRD_COUNTRY_DUTY_ID && measure.additional_code.blank? && measure.goods_nomenclature_sid != @declarable.goods_nomenclature_sid
+      }.compact!
+    end
+  end
+
+  def delete_duplicate_vat_measures(type:)
+    return unless select { |measure| measure.measure_type_id == type }.any?
+
+    delete_if { |measure| measure.measure_type_id == type && measure.goods_nomenclature_sid != @declarable.goods_nomenclature_sid }
+    latest = select { |measure| measure.measure_type_id == type }.sort_by(&:effective_start_date).pop
```
alexdesi, line 49: pop implies we are removing the last element, but is't not clear from where,
does not select create a copy of the collection?
should you use last? 🤷

app/models/measure_collection.rb
```diff
@@ -0,0 +1,66 @@
...
+  def third_country_duty_dedup
+    # TODO: Fix measures with additional codes that are duplicate
+    if select { |measure| measure.measure_type_id == THIRD_COUNTRY_DUTY_ID }.size > 1
+      delete_if { |measure|
+        measure.measure_type_id == THIRD_COUNTRY_DUTY_ID && measure.additional_code.blank? && measure.goods_nomenclature_sid != @declarable.goods_nomenclature_sid
+      }.compact!
+    end
+  end
+
+  def delete_duplicate_vat_measures(type:)
```
alexdesi, line 45: delete_duplicate is a synonym for `unique`, so we could rename it to 
`vat_measure_unique!`
up to you

app/models/measure_collection.rb
```diff
@@ -0,0 +1,66 @@
...
+    return unless select { |measure| measure.measure_type_id == type }.any?
+
+    delete_if { |measure| measure.measure_type_id == type && measure.goods_nomenclature_sid != @declarable.goods_nomenclature_sid }
+    latest = select { |measure| measure.measure_type_id == type }.sort_by(&:effective_start_date).pop
+    delete_if { |measure| measure.measure_type_id == type && matching_item_id(measure) }
+    prepend(latest) unless include?(latest)
+    compact!
+  end
```
alexdesi, line 53: this is a bit complex, it  is filtering by measure_type_id 4 times, I am wondering if we can simplify it

`{ |measure| measure.measure_type_id == type }`

spec/models/measure_collection_spec.rb
willfish: See line 8 and the one spec that isn't interested in the service

app/models/measure_collection.rb
willfish: This is copied and pasted for the most part. I'm planning on having another pop at this file later

app/models/measure_collection.rb
willfish: This is copied and pasted for the most part. The main bits I've changed is the interface and the addition of the filter method.

alexdesi (approved):

alexdesi:
have a look to the linters issue, they seems easy to fix

PR #306, opened 2021-11-25
Tighten validation on Rules of Origin query

jebw:
### Jira link

[HOTT-1001](https://transformuk.atlassian.net/browse/HOTT-1001)

### What?

I have added/removed/altered:

- [x] Added parameter checks to the RulesOfOrigin::Query

### Why?

I am doing this because:

- Because we are going to publicly expose the Rules of Origin api

### DIFF
--- a/app/models/rules_of_origin/query.rb
+++ b/app/models/rules_of_origin/query.rb
@@ -36,7 +36,7 @@ module RulesOfOrigin
       scheme_set.links + schemes.map(&:links).flatten
     end
 
-    class InvalidParams < StandardError; end
+    class InvalidParams < ArgumentError; end
 
     private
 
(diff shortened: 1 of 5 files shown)


app/models/rules_of_origin/query.rb
```diff
@@ -1,13 +1,19 @@
+# frozen_string_literal: true
+
 module RulesOfOrigin
   class Query
+    HEADING_CHECKER = /\A\d{6}\z/
+    COUNTRY_CHECKER = /\A[A-Z]{2}\z/
```
alexdesi, line 6: what if we pass strings like "XY" ?
I guess it's ok not to check the country at that detail level, just the format here 🤔

jebw: Yeh - they'll just get no matches - intention is to just not act like everything is in the API request when its not - previously we just returned empty data which will potentially confuse the person accessing the API

PR #312, opened 2021-12-02
HOTT-1143-sentry-rails: Migrate to sentry-rails and sentry-sidekiq

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1143

### What?

I have added/removed/altered:

- [x] Added sentry-rails and sentry-sidekiq gem
- [x] Added monkey patch for hard dependency on ActiveRecord
- [x] Fixed references in specs and code from `Raven` to `Sentry`

### Why?

I am doing this because:

- sentry-raven is out of support

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,8 @@ gem 'nokogiri'
 gem 'omniauth-rails_csrf_protection'
 gem 'ox'
 gem 'plek'
-gem 'sentry-raven'
+gem 'sentry-rails'
+gem "sentry-sidekiq"
 
 # API related
 gem 'ansi'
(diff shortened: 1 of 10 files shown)


jebw (approved):
Looks good - only query would be around the monkey patch

This overrides the monkey-patch included in `sentry-rails` _but_ that does check for `defined(ActiveRecord)` before applying its monkey patch - would we be better not having ActiveRecord defined, presumably this will potentially cause problems in other gems.

https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/railtie.rb#L63

Or is this necessary for the date type casting stuff?

PR #313, opened 2021-12-03
Update new RoR - 24 nov 2021

alexdesi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1152

### What?

- [x] Add the new RoR dated on 24 nov 2021
- [x] Check if they are correct (Res: validation was successful)
- [x] Add a doc file about Rules of Origin (how to update them)

### Why?
RoO needs to be updated.


docs/rules_of_origin.md
```diff
@@ -0,0 +1,52 @@
...
+  Containing the mapping of rules and commodities
+
+- roo_schemes_uk.json
+  Is the json scheme of the UK RoO
+
+- roo_schemes_xi.json
+  Is the JSON scheme of the Northern Ireland RoO
+
+
+There are many docs in *md* (markdown) format that will be rendered in the UI in the he folders:
```
jebw, line 22: `s/he/following/`

docs/rules_of_origin.md
```diff
@@ -0,0 +1,52 @@
...
+
+`/roo_schemes_uk/fta_intro/`
+`/roo_schemes_uk/fta_intro/introductory_notes/`
```
jebw, line 25: I think this should be `/roo_schemes_uk/introductory_notes`

docs/rules_of_origin.md
```diff
@@ -0,0 +1,52 @@
...
+
+for the UK, and
+
+`/roo_schemes_uk/fta_intro/`
+`/roo_schemes_uk/fta_intro/introductory_notes/`
```
jebw, line 30: I think this should be `/roo_schemes_xi/introductory_notes`

jebw (approved):
LGTM - just a couple of minor tweaks to the documentation file - thanks for adding this :+1:

app/models/rules_of_origin/heading_mappings.rb
```diff
@@ -7,7 +7,7 @@ class HeadingMappings
     SERVICES = %w[uk xi both].freeze
     SUB_HEADING_FORMAT = %r{\A\d{6}\z}
     DEFAULT_SOURCE_PATH = Rails.root.join('db/rules_of_origin').freeze
-    DEFAULT_FILE = 'rules_to_commodities_210929.csv'
+    DEFAULT_FILE = 'rules_to_commodities_211124.csv'
```
willfish, line 10: Given we change these files so frequently is there any appetite for decoupling the uploading of the files from the codebase?

We still get the persistence buckets for dev, staging and production so we can verify them first

alexdesi: That's a good point, this is done, but happy to discuss it next time we must update those rules.

willfish: Sure :)

PR #315, opened 2021-12-06
HOTT-1182: Surface compound measurement units

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1182

### What?

I have added/removed/altered:

- [x] Altered .measurement_unit so that it surfaces compound measurement units for the duty calculator
- [x] Updated tests to reflect this change

### Why?

I am doing this because:

- ASVX measurement units are currently calculated incorrectly and we do not ask for all the units that apply in this case.

### DIFF
--- a/app/models/measurement_unit.rb
+++ b/app/models/measurement_unit.rb
@@ -15,11 +15,23 @@ class MeasurementUnit < Sequel::Model
   delegate :description, to: :measurement_unit_description
 
   class << self
-    def measurement_unit(unit_code, unit_key)
-      measurement_units[unit_key] ||= build_missing_measurement_unit(unit_code, unit_key)
+    def units(unit_code, unit_key)
+      unit = measurement_units[unit_key] || build_missing_measurement_unit(unit_code, unit_key)
+
+      if unit['compound_units'].present?
+        compound_units_for(unit)
+      else
+        [unit]
+      end
     end
 
     private
+
+    def compound_units_for(unit)
+      unit['compound_units'].flat_map do |unit_key|
+        units(unit['measurement_unit_code'], unit_key)
+      end
+    end
 
     def measurement_units
       @measurement_units ||=
--- a/app/services/measure_unit_service.rb
+++ b/app/services/measure_unit_service.rb
@@ -9,9 +9,11 @@ class MeasureUnitService
       unit_qualifier_code = unit[:measurement_unit_qualifier_code]
       unit_key = "#{unit_code}#{unit_qualifier_code}"
 
-      if acc[unit_key].blank?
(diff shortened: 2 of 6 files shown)


app/models/measurement_unit.rb
```diff
@@ -15,12 +15,24 @@ class MeasurementUnit < Sequel::Model
   delegate :description, to: :measurement_unit_description
 
   class << self
-    def measurement_unit(unit_code, unit_key)
-      measurement_units[unit_key] ||= build_missing_measurement_unit(unit_code, unit_key)
+    def units(unit_code, unit_key)
+      unit = measurement_units[unit_key] ||= build_missing_measurement_unit(unit_code, unit_key)
```
alexdesi, line 19: it would be much clear if we do not memoize something and returns something else.

One possibility is to leave the `measurement_unit` method as it is and create another method  that returns 
``` 
if unit['compound_units'].present?
        compound_units_for(unit)
      else
        [unit]
      end
```
or maybe you can just remove the memoization and split this line
`unit = measurement_units[unit_key] ||= build_missing_measurement_unit(unit_code, unit_key)` 
which is a bit confusing (you have to think about the priority of the operator `=` and `||=` )

app/services/measure_unit_service.rb
```diff
@@ -9,9 +9,15 @@ def call
       unit_qualifier_code = unit[:measurement_unit_qualifier_code]
       unit_key = "#{unit_code}#{unit_qualifier_code}"
 
-      if acc[unit_key].blank?
-        acc[unit_key] = {}
-        acc[unit_key] = MeasurementUnit.measurement_unit(unit_code, unit_key)
+      next if acc[unit_key].present?
```
alexdesi, line 12: does this serve to filter out the acc having unit_key?
can we do it before the loop?
`units.reject{|k,v| v[:measurement_unit_code] || v[:measurement_unit_qualifier_code] }`
so we can simplify it?

alexdesi (commented):
I am not sure about the assignement + memoization in the same line,
+
it would be nice to filter out instead of using next (good old functional approach 😏 )

app/models/measurement_unit.rb
willfish: I suspect the memoization was done to reduce the database call/construction of the fallback when there is a missing measure unit in the file or the database

willfish: I've removed this now

app/services/measure_unit_service.rb
willfish: No it does not filter. It avoids having to accumulate already accumulated measure units on the same measure.

PR #321, opened 2021-12-14
Push updates back a bit and chain the reindex when the updates actually happen

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1203

### What?

I have added/removed/altered:

- [x] Updated sidekiq schedule by 30 minutes for the update synchronizer worker
- [x] Removed the sidekiq schedule for the cache clearing
- [x] Added an optional parameter to the worker which kicks off the relevant worker after the update has finished and only when there are no pending or failed updates
- [x] Linted the specs to reflect better_specs principles https://www.betterspecs.org/
- [x] Replaced the now-defunct use_cds? business method with the uk? service method and removed additional specs
- [x] Added specs for the tariff synchronizer apply_cds and apply methods which now kick off the sidekiq client with the cache clear worker

### Why?

I am doing this because:

- The extra 30 minutes enables us to better capture slow to drop api files on the cds files
- Only running when all updates are applied  and there are no failed updates will catch extra work which doesn't add value
- Chaining the Update and Cache workers will reduce the arbitrary latency/schedules we have setup for these two related jobs and should speed up the process overall

### DIFF
--- a/app/workers/updates_synchronizer_worker.rb
+++ b/app/workers/updates_synchronizer_worker.rb
@@ -11,11 +11,11 @@ class UpdatesSynchronizerWorker
     if TradeTariffBackend.use_cds?
       TariffSynchronizer.download_cds
       logger.info 'Applying...'
-      TariffSynchronizer.apply_cds
+      TariffSynchronizer.apply_cds(reindex_all_indexes: true)
     elsif TradeTariffBackend.xi?
       TariffSynchronizer.download
       logger.info 'Applying...'
-      TariffSynchronizer.apply
+      TariffSynchronizer.apply(reindex_all_indexes: true)
     end
   end
 end
--- a/app/workers/updates_synchronizer_worker.rb
+++ b/app/workers/updates_synchronizer_worker.rb
@@ -7,8 +7,7 @@ class UpdatesSynchronizerWorker
     logger.info 'Running UpdatesSynchronizerWorker'
     logger.info 'Downloading...'
 
-    # TODO: this check can be removed when we switch to CDS data.
-    if TradeTariffBackend.use_cds?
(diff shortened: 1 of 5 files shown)


jebw (commented):
Looks great - just needs a description and a spec to check the job gets stored

I wonder if it should be half five? Gives more time for the file to turn up and a lot of smaller updates only take 5 - 10 mins to apply so we gain 20-25 mins by running the cache clear and reindex immediately after.

app/workers/updates_synchronizer_worker.rb
```diff
@@ -7,15 +7,14 @@ def perform
     logger.info 'Running UpdatesSynchronizerWorker'
     logger.info 'Downloading...'
 
-    # TODO: this check can be removed when we switch to CDS data.
-    if TradeTariffBackend.use_cds?
+    if TradeTariffBackend.uk?
```
willfish, line 10: uk? and use_cds? are the same. We could do with getting rid of the use_cds flag

config/sidekiq.yml
```diff
@@ -19,14 +19,11 @@
   # Online converter: https://crontab.guru/#0_22_*_*_*
   #
   UpdatesSynchronizerWorker:
-    cron: "30 4 * * *"
+    cron: "0 5 * * *"
```
willfish, line 22:
```suggestion
    cron: "30 5 * * *"
```

app/workers/updates_synchronizer_worker.rb
jebw: Yeh - lets go with that - we lose the ability to disable CDS sync but why we'd want to I don't know :man_shrugging:  and its another potential for misconfiguration

We can't disable Tariq in XI anyway

PR #323, opened 2021-12-15
First attempt to remove recursion from GoodsNomenclatureMapper.

alexdesi:
### What?
First attempt to remove recursion from GoodsNomenclatureMapper.

### Why?
To simplify the GoodsNomenclatureMapper - Still too complex anyway, that's only the first step.

### DIFF
--- a/app/lib/goods_nomenclature_mapper.rb
+++ b/app/lib/goods_nomenclature_mapper.rb
@@ -14,7 +14,7 @@ class GoodsNomenclatureMapper
     def leaf?
       children.empty?
     end
-    alias :leaf :leaf?
+    alias_method :leaf, :leaf?
 
     def ancestors
       @ancestors ||= []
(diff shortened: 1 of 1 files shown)


app/lib/goods_nomenclature_mapper.rb
```diff
@@ -52,20 +52,19 @@ def for_goods_nomenclature(ref_goods_nomenclature)
...
+  def traverse(goods_nomenclatures, current_goods_nomenclature)
+    current_good = current_goods_nomenclature
+
+    while (index_gn = goods_nomenclatures.index(current_good))
+      next_good = goods_nomenclatures[index_gn + 1]
+      return unless next_good
+
+      map_goods_nomenclatures(current_good, next_good)
+      current_good = next_good
     end
```
alexdesi, line 68: this is the refactor,
the rest of the changes are made by Rubocop.

app/lib/goods_nomenclature_mapper.rb
```diff
@@ -52,20 +52,19 @@ def for_goods_nomenclature(ref_goods_nomenclature)
...
-    unless goods_nomenclatures.index(secondary).blank?
-      next_goods_nomenclature = goods_nomenclatures[goods_nomenclatures.index(secondary) + 1]
-      if next_goods_nomenclature.present? # we are not at the end of the goods_nomenclature array
-        map_goods_nomenclatures(secondary, next_goods_nomenclature)
-        traverse(goods_nomenclatures, secondary, next_goods_nomenclature)
-      end
+  def traverse(goods_nomenclatures, current_goods_nomenclature)
+    current_good = current_goods_nomenclature
+
+    while (index_gn = goods_nomenclatures.index(current_good))
```
willfish, line 62: This is basically just an `each_cons(2)` no? I've basically never seen a need for while in ruby unless the condition for exiting the loop is outside of the data-structure.

willfish (changes requested):
I'm loath to make changes to this before we've added specs that cover the necessary and sufficient aspects of this class and it's api

app/lib/goods_nomenclature_mapper.rb
alexdesi: yeah, I thought about it, in this first step I strictly followed the logic in the recursion, but it seems to be just a each loop, I'll change it.

PR #324, opened 2021-12-16
Bugfix comparison of Integer with nil failed

alexdesi:
### Jira link
[HOTT-1150](https://transformuk.atlassian.net/browse/HOTT-1150)

### What?
This Patch fixes the bug "comparison of Integer with nil failed" on the goods_nomenclature_mapper.
Bug on Sentry: https://sentry.io/organizations/engine-le/issues/2732997178/events/ea7919efbaa443a299ee0db53622cb32/?project=5557005&statsPeriod=24h

### Why?
We do not like bugs

### DIFF
--- a/app/lib/goods_nomenclature_mapper.rb
+++ b/app/lib/goods_nomenclature_mapper.rb
@@ -14,7 +14,7 @@ class GoodsNomenclatureMapper
     def leaf?
       children.empty?
     end
-    alias :leaf :leaf?
+    alias_method :leaf, :leaf?
 
     def ancestors
       @ancestors ||= []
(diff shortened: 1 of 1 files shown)


app/lib/goods_nomenclature_mapper.rb
```diff
@@ -128,4 +129,11 @@ def parent_of(goods_nomenclature)
   def heading_map?(primary, secondary)
     primary.is_a?(Heading) && secondary.is_a?(Heading)
   end
+
+  def reload_if_nil(goods_nomenclature)
+    # This is a patch to fix the error: "comparison of Integer with nil failed"
+    logger.warn("goods_nomenclature.number_indents is nil! - 'goods_nomenclature': #{goods_nomenclature}")
```
alexdesi, line 135: I thought it could be useful to log this for debugging purposes.

app/lib/goods_nomenclature_mapper.rb
```diff
@@ -128,4 +129,11 @@ def parent_of(goods_nomenclature)
...
+
+    goods_nomenclature.reload if goods_nomenclature.number_indents.nil?
```
alexdesi, line 137: since the Sentry alerts happen only a few times per day, the reload should only be triggered a few times per day too.

willfish (approved):
:lipstick: :pig:

app/lib/goods_nomenclature_mapper.rb
```diff
@@ -128,4 +129,11 @@ def parent_of(goods_nomenclature)
   def heading_map?(primary, secondary)
     primary.is_a?(Heading) && secondary.is_a?(Heading)
   end
+
+  def reload_if_stale(goods_nomenclature)
+    # This is a patch to fix the error: "comparison of Integer with nil failed"
```
willfish, line 134: We need to return unless the indents are nil at the start, here

willfish (changes requested):
We need to not log lots

app/lib/goods_nomenclature_mapper.rb
alexdesi: done.

willfish (approved):

PR #328, opened 2021-12-23
HOTT-1224 fix v1 api with v2 default

jebw:
### Jira link

[HOTT-1224](https://transformuk.atlassian.net/browse/HOTT-1224)

### What?

I have added/removed/altered:

- [x] Added a request spec for the commodities controller which checks the API responses for both v1 and v2
- [x] Added a spec for the ApiConstraints class which verifies the behaviour with v1 and v2 defaults
- [x] Made changes to the ApiConstraints to not rely on ordering of usage in the routes file

### Why?

I am doing this because:

- The ApiConstraints class is always returns true if the default api version is the same as the instance of ApiConstraints is configured for. This works at present because the V1 routes come after the v2 routes, so provided the default is v1, the router has already determined the route is not a v2 route at the point it selects a v1 api controller. If the default is v2, this means the v2 routes (which are processed before the v1 routes) get matched to the v1 request and return an incorrect API response.

- This corrects a bug in the Api routing - it _shouldn't_ affect the routing.

### DIFF
--- a/app/lib/api_constraints.rb
+++ b/app/lib/api_constraints.rb
@@ -1,19 +1,35 @@
 class ApiConstraints
+  VERSION_HEADER_MATCHER = %r{application/vnd\.uktt\.v\d}
+
+  class << self
+    def default_version
+      ENV.fetch('DEFAULT_API_VERSION', '1')
+    end
+  end
+
   def initialize(version:)
     @version = version.to_s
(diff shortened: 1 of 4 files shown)


willfish:
Is this still relevant @jebw? Quite an oldy 
image

PR #331, opened 2021-12-24
Allow developers to assign log_levels locally

jebw:
### Jira link

HOTT-????

### What?

I have added/removed/altered:

- [x] Added an env var to control log levels in the development environment

### Why?

I am doing this because:

- So that I can have query logging on by default without requiring other developers to do so


alexdesi (approved):
This is handy 👍

PR #338, opened 2022-01-05
HOTT-1225: Make news item target and service optional parameters

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1225

### What?

I have added/removed/altered:

- [x] Removed forcing the user to supply a service and target for the v2 new items api
- [x] Removed propagation of not found for targets and services under our own control
- [x] Added shared examples for unfiltered variants of news item target and service

### Why?

I am doing this because:

- We need to be able to return all of the target and service options
- We don't need to be defensive for this internal api

### DIFF
--- a/spec/models/news_item_spec.rb
+++ b/spec/models/news_item_spec.rb
@@ -45,14 +45,17 @@ RSpec.describe NewsItem do
       let(:both_page) { create :news_item, show_on_uk: true, show_on_xi: true }
       let(:neither_page) { create :news_item, show_on_uk: false, show_on_xi: false }
 
-      context 'without service name' do
-        let(:service_name) { nil }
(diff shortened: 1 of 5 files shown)


spec/models/news_item_spec.rb
```diff
@@ -71,32 +74,27 @@
...
+        it { is_expected.to include updates_page }
+        it { is_expected.to include both_page }
+        it { is_expected.to include neither_page }
       end
 
-      let :neither_page do
-        create :news_item, show_on_home_page: false, show_on_updates_page: false
-      end
+      it_behaves_like 'a non-filtering target filter invocation', ''
+      it_behaves_like 'a non-filtering target filter invocation', nil
```
alexdesi, line 97: should you add an example with 'unknown'?

alexdesi (approved):
✅

PR #340, opened 2022-01-07
HOTT-1240: Adds productline suffix to search references

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1240

### What?

I have added/removed/altered:

- [x] Added migration to add productline_suffix to search references

### Why?

I am doing this because:

- This is required to differentiate commodity subheadings


alexdesi (approved):
it looks good, maybe we should add a test, but I guess we can do it later when we start using that field.

alexdesi:
It came to mind, should we backfill the existing entries?

willfish:
That's what the default does @alexdesi https://www.postgresql.org/docs/9.5/ddl-alter.html

PR #350, opened 2022-01-13
HOTT-1261: Fix sequencing of transforms during CDS ETL

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1261

### What?

I have added/removed/altered:

- [x] Added spec to avoid regressions in ETL callback sequencing.
- [x] Added new callback on CDS ETL which enables transforms to be passed to the model.
- [x] Altered the sequencing of the TaricSynchronizer::EntityMapper code such that transformed attributes take.

### Why?

I am doing this because:

- When we moved to callbacks I missed the sequencing mattered and broke imports of memberships from CDS
- We want to make sure there's coverage to avoid this happening in future


jebw (approved):
LGTM

PR #353, opened 2022-01-14
HOTT-1276: Tidy up search references

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1276

### What?

I have added/removed/altered:

- [x] Removed unused scope filtering methods
- [x] Removed search references being preloaded and associated with serialized commodities in the CachedHeadingService
- [x] Removed eager loading custom behaviour

### Why?

I am doing this because:

- The cached headings do not even serialize the commodity search references and this is just overhead: https://github.com/trade-tariff/trade-tariff-backend/blob/main/app/serializers/cache/heading_serializer.rb#L73-L123. I believe these were put in place as part of the V2 api previously being used for admin ui purposes before the separation: https://github.com/trade-tariff/trade-tariff-backend/commit/0c9f7a120a75c4cf31be6f7649372733ae93f602.
- The scope filtering methods just are not used
- The eager loading is also not used - I suspect it would have been part of enabling eager loading of search references as part of the CachedHeadingService

### DIFF
--- a/app/elastic_search_indexes/search/search_reference_index.rb
+++ b/app/elastic_search_indexes/search/search_reference_index.rb
@@ -19,8 +19,8 @@ module Search
                   properties: {
                     position: { type: 'long' },
                     title: { type: 'text' },
-                    numeral: { type: 'keyword' }
-                  }
+                    numeral: { type: 'keyword' },
+                  },
                 },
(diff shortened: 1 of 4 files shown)


jebw (approved):
LGTM

PR #354, opened 2022-01-17
HOTT-1260 Retry data loads

jebw:
### Jira link

[HOTT-1260](https://transformuk.atlassian.net/browse/HOTT-1260)

### What?

I have added/removed/altered:

- [x] Added a reschedule feature to UpdatesSynchronizer

### Why?

I am doing this because:

- Sometimes the CDS update file arrives later then when our synchronizer is scheduled to run. If that is the case - re-schedule every 20 minutes until the cut off time is reached. Once the cut off time is reached, if there still isn't a file for today, the job will continue as per normal.

- Changes the behaviour of the overnight sync job

### DIFF
--- a/spec/unit/tariff_synchronizer/cds_update_spec.rb
+++ b/spec/unit/tariff_synchronizer/cds_update_spec.rb
@@ -20,6 +20,28 @@ RSpec.describe TariffSynchronizer::CdsUpdate do
     end
   end
 
+  describe '.downloaded_todays_file?' do
+    subject { described_class.downloaded_todays_file? }
+
+    context 'when todays file is present in table' do
+      # Note published day after file name, so todays file has yesterdays date
+      before { create :cds_update, example_date: Time.zone.yesterday }
+
+      it { is_expected.to be true }
+    end
+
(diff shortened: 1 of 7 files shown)


spec/unit/tariff_synchronizer/cds_update_spec.rb
```diff
@@ -20,6 +20,28 @@
     end
   end
 
+  describe 'downloaded_todays_file?' do
+    subject { described_class.downloaded_todays_file? }
+
+    context 'when todays file present in table' do
```
willfish, line 26:
```suggestion
    context 'when todays file is present in table' do
```

spec/unit/tariff_synchronizer/cds_update_spec.rb
```diff
@@ -20,6 +20,28 @@
...
+      # Note published day after file name, so todays file has yesterdays date
+      before { create :cds_update, issue_date: Time.zone.yesterday }
```
willfish, line 28: Should this be pending?

willfish (approved):
Awesome :pray:

PR #356, opened 2022-01-18
HOTT-1250: Refactor by extraction - commodity annotations

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1250

### What?

I have added/removed/altered:

- [x] Extract the commodity tree annotation from the CachedHeadingService
- [x] Added exhaustive example heading to cover annotation outcomes for < indents > indents and == indent situations

### Why?

I am doing this because:

- I wanted to make sure I can trust the test suite for when I extend the functionality of the commodity tree annotations for the subheading work.
- I also wanted to make sure we can reuse this behaviour and easily test it

### DIFF
--- a/app/services/annotated_commodity_service.rb
+++ b/app/services/annotated_commodity_service.rb
@@ -4,7 +4,7 @@
 #
 # This is then consumed by the frontend tariff application to easily build a tree UI for perusal
 # of headings, subheadings and declarable commodities.
-class CommoditiesTreeAnnotationService
+class AnnotatedCommodityService
   def initialize(heading)
     @heading = heading
   end
--- a/app/services/heading_service/cached_heading_service.rb
+++ b/app/services/heading_service/cached_heading_service.rb
@@ -108,7 +108,7 @@ module HeadingService
     end
 
     def build_commodities_tree
-      CommoditiesTreeAnnotationService.new(result).call
+      AnnotatedCommodityService.new(result).call
     end
   end
 end
--- a/spec/services/annotated_commodity_service_spec.rb
+++ b/spec/services/annotated_commodity_service_spec.rb
@@ -1,4 +1,4 @@
-RSpec.describe CommoditiesTreeAnnotationService do
+RSpec.describe AnnotatedCommodityService do
   subject(:service) { described_class.new(heading) }
 
   describe '#call' do


app/services/commodities_tree_annotation_service.rb
```diff
@@ -0,0 +1,89 @@
...
+    if @heading.commodities.index(current_goods_nomenclature).present?
+      next_goods_nomenclature = @heading.commodities[@heading.commodities.index(current_goods_nomenclature) + 1]
+
+      if next_goods_nomenclature.present? # we are not at the end of the goods_nomenclature array
+        map_goods_nomenclatures(current_goods_nomenclature, next_goods_nomenclature)
+
+        traverse(next_goods_nomenclature)
+      end
+    end
+  end
```
alexdesi, line 32: I think we can just iterate, in my first attempt I wrote this:
```
def traverse(goods_nomenclatures, current_goods_nomenclature)
    current_good = current_goods_nomenclature

    while (index_gn = goods_nomenclatures.index(current_good))
      next_good = goods_nomenclatures[index_gn + 1]
      return unless next_good

      map_goods_nomenclatures(current_good, next_good)
      current_good = next_good
    end
  ```
  which is much simpler and can be improved further anyway.

Comment (willfish) [reply app/services/commodities_tree_annotation_service.rb:32]: So the brief of this refactor was to extract/preserve original behaviour and add coverage.

Comment (willfish) [reply app/services/commodities_tree_annotation_service.rb:32]: See this https://github.com/trade-tariff/trade-tariff-backend/pull/356/files#diff-fe5d0c8f58fc8ad73105d76099811869197f913b7728926b15a87016ebea9ebcR20

Comment (willfish) [reply app/services/commodities_tree_annotation_service.rb:32]: At some point I will remove the recursion and iterate over staggered pairs of commodities

File: spec/services/commodities_tree_annotation_service_spec.rb [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+        true,  # ---- Other
+        true,  # Asses
+        true,  # Other
+      ]
+      actual_leafs = service.call.commodities.map { |commodity| commodity['leaf'] }
+
+      expect(actual_leafs).to eq(expected_leafs)
+    end
+  end
+end
```
alexdesi, line 51: are these the only cases? I expected more, maybe I am wrong.

alexdesi (approved):
Nice one!

app/services/commodities_tree_annotation_service.rb
```diff
@@ -0,0 +1,89 @@
+# Walk the commodities of a heading/subheading (assumed to be ordered correctly)
+# and annotate them with whether they are a leaf (have no children) and
+# with the id of their parent goods_nomenclature
+#
+# This is then consumed by the frontend tariff application to easily build a tree UI for perusal
+# of headings, subheadings and declarable commodities.
+class CommoditiesTreeAnnotationService
```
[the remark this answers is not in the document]
willfish: Just following the naming of the other services. Have renamed it AnnotatedCommodityService to be more in keeping

willfish:
Validated headings still surfacing on dev

PR #359, opened 2022-01-20
HOTT-1286: Removes missing file generator.

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Removes the missing file update generation code, sidekiq schedule and related spec files

### Why?

I am doing this because:

- This generator negatively impacts the download process by impacting
the start range from which we check for applicable downloads in Taric
and is superfluous/adds no perceivable value from a debug/functional
perspective.


alexdesi (approved):
nice cleaning up

PR #360, opened 2022-01-20
HOTT-1252: Check the last 10 files for a correct sequence in Taric

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1252

### What?

I have added/removed/altered:

- [x] Alter the file sequence check business logic in TaricUpdate such that we check the last 10 files for correct sequence

### Why?

I am doing this because:

- We want this because we can have multiple files arrive out of order.
This can lead to us treating a correct sequence of the last two files as
a false positive indicator to proceed when we should be flapping in
slack.

### DIFF
--- a/app/lib/tariff_synchronizer.rb
+++ b/app/lib/tariff_synchronizer.rb
@@ -309,7 +309,7 @@ module TariffSynchronizer
   end
 
   def check_sequence
-    unless update_type.correct_recent_filename_sequence?
+    unless update_type.correct_filename_sequence?
       raise FailedUpdatesError, 'Wrong sequence between the pending and applied files. Check the admin updates UI.'
     end
   end
--- a/app/lib/tariff_synchronizer.rb
+++ b/app/lib/tariff_synchronizer.rb
@@ -95,7 +95,7 @@ module TariffSynchronizer
   # Taric
   def apply(reindex_all_indexes: false)
     check_tariff_updates_failures
-    check_sequence_for_pending_file
+    check_sequence
 
     applied_updates = []
     import_warnings = []
@@ -129,7 +129,7 @@ module TariffSynchronizer
 
   def apply_cds(reindex_all_indexes: false)
     check_tariff_updates_failures
-    check_sequence_for_pending_file
+    check_sequence
 
     applied_updates = []
     import_warnings = []
@@ -308,8 +308,8 @@ module TariffSynchronizer
     Sequel::Model.subclasses
   end
 
-  def check_sequence_for_pending_file
-    unless update_type.correct_filename_sequence?
+  def check_sequence
+    unless update_type.correct_recent_filename_sequence?
       raise FailedUpdatesError, 'Wrong sequence between the pending and applied files. Check the admin updates UI.'
     end
   end
--- a/app/lib/tariff_synchronizer/taric_update.rb
+++ b/app/lib/tariff_synchronizer/taric_update.rb
@@ -6,18 +6,16 @@ module TariffSynchronizer
     SEQUENCE_APPLICABLE_STATES = [APPLIED_STATE, PENDING_STATE].freeze
 
     class << self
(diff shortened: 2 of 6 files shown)


willfish:
I've validated this works with a current dump of the staging XI data

spec/unit/tariff_synchronizer/taric_update_spec.rb
```diff
@@ -43,43 +43,76 @@
...
   end
 
-  describe '.correct_filename_sequence?' do
-    subject(:taric_update) { described_class }
+  describe '.correct_recent_filename_sequence?' do
+    subject(:update) { described_class }
+
+    context 'when there are updates with an unbroken sequence' do
+      before do
+        create(:taric_update, :missing, example_date: Date.parse('2021-12-04'), sequence_number: 'foo')
```
alexdesi, line 51: do we need this in the example, maybe it's not necessary

spec/unit/tariff_synchronizer/taric_update_spec.rb
```diff
@@ -43,43 +43,76 @@
...
 
-    let(:applied_date) { Date.parse('2021-12-01') }
-    let(:applied_sequence_number) { '002' }
+    context 'when there are only applied updates' do
+      before do
+        create(:taric_update, :applied, example_date: Date.parse('2021-12-02'), sequence_number: '202')
+        create(:taric_update, :applied, example_date: Date.parse('2021-12-01'), sequence_number: '200')
+      end
+
+      it { is_expected.to be_correct_recent_filename_sequence }
```
alexdesi, line 76: this was a bit surprising to me since the sequence of the applied was not correct

app/lib/tariff_synchronizer.rb
```diff
@@ -309,7 +309,7 @@ def sequel_models
   end
 
   def check_sequence_for_pending_file
-    unless update_type.correct_filename_sequence?
+    unless update_type.correct_recent_filename_sequence?
```
alexdesi, line 312: this could be a bit confusing, it's not easy to say the difference between
`correct_filename_sequence?`
`correct_recent_filename_sequence?`

should be rename `correct_filename_sequence?` to something like `compare_pair_sequence?`
and 
`correct_recent_filename_sequence?` to just `correct_filename_sequence`?

spec/unit/tariff_synchronizer/taric_update_spec.rb
willfish: Yeah, we do for now, since in live environments we have these updates and they should be filtered out. This is an implicit test of that

spec/unit/tariff_synchronizer/taric_update_spec.rb
willfish: We could push the check for pending updates to the tariff synchronizer apply method.

willfish: Actually this is already the case

app/lib/tariff_synchronizer/taric_update.rb
```diff
@@ -1,24 +1,34 @@
 module TariffSynchronizer
   class TaricUpdate < BaseUpdate
-    REGEX_TARIC_SEQUENCE = /^\d{4}-\d{2}-\d{2}_TGB(?<year>\d{2})(?<sequence>\d+).xml$/
     NEW_YEAR_STARTING_SEQUENCE_NUMBER = 1
+    REGEX_TARIC_SEQUENCE = /^\d{4}-\d{2}-\d{2}_TGB(?<year>\d{2})(?<sequence>\d+).xml$/
+    SEQUENCE_APPLICABLE_UPDATE_LIMIT = 10
+    SEQUENCE_APPLICABLE_STATES = [APPLIED_STATE, PENDING_STATE].freeze
 
     class << self
+      # Validates the last n of updates are in the correct sequence in order to know whether we're safe to apply pending updates. Out of order updates happen when the Taric api publishes files later than the date they're meant to be downloaded and should halt the applying of the update process.
```
alexdesi, line 9: should we move this long comment on the tests?

alexdesi (approved):
nice one! 🍏

PR #361, opened 2022-01-20
HOTT-1286: Extend download date range logic to be 20 days ago

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Altered the download updates behaviour by defaulting to 20 days as the first date of the files that we try to download.
- [x] Altered the download updates behaviour so that we ignore missing updates and only pay attention to the pending, applied and failed ones

### Why?

I am doing this because:

- Now that we've removed the missing files functionality and we know that the files can come out of order we need to try and download previously unavailable files which now exist. Our tolerance is set to the last 20 days just as a starting point.
- I've applied this behaviour to the CdsUpdate type to keep things
consistent between the implementations.

### DIFF
--- a/app/lib/tariff_synchronizer/base_update.rb
+++ b/app/lib/tariff_synchronizer/base_update.rb
@@ -1,5 +1,7 @@
 module TariffSynchronizer
   class BaseUpdate < Sequel::Model(:tariff_updates)
+    DOWNLOAD_FROM = 20.days
+
     delegate :instrument, to: ActiveSupport::Notifications
 
     # Used for TARIC updates only.
@@ -62,6 +64,10 @@ module TariffSynchronizer
         where(state: FAILED_STATE)
       end
 
+      def pending_applied_and_failed
+        where(state: [PENDING_STATE, APPLIED_STATE, FAILED_STATE])
+      end
+
       def pending_or_failed
         where(state: [PENDING_STATE, FAILED_STATE])
       end
@@ -76,6 +82,10 @@ module TariffSynchronizer
 
       def last_applied
         applied.order(:issue_date).first
+      end
+
+      def last_failed
+        failed.order(:issue_date).first
       end
 
       def descending
@@ -140,7 +150,8 @@ module TariffSynchronizer
       delegate :instrument, to: ActiveSupport::Notifications
 
       def sync
-        (pending_from..Date.current).each { |date| download(date) }
+        applicable_download_date_range.each { |date| download(date) }
+
         notify_about_missing_updates if last_updates_are_missing?
       end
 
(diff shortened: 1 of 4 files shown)


app/lib/tariff_synchronizer/base_update.rb
```diff
@@ -140,28 +150,44 @@ class << self
...
+
+      def download_end_date
+        Time.zone.today
+      end
+
+      def download_start_date
+        if pending_applied_and_failed.count.zero?
           TariffSynchronizer.initial_update_date_for(update_type)
+        else
+          last_download = (last_pending || last_applied || last_failed)
```
alexdesi, line 176: Maybe this should just be the less recent:
`[last_pending, last_applied, last_failed].min`
ignoring the priority between them?

app/lib/tariff_synchronizer/base_update.rb
```diff
@@ -140,28 +150,44 @@ class << self
...
+
+          # Handle missing downloads when older downloads came late
+          if last_download.issue_date < DOWNLOAD_FROM.ago.to_date
+            last_download.issue_date
+          else
+            DOWNLOAD_FROM.ago.to_date
+          end
```
alexdesi, line 183: is not this the same as:
`[last_download.issue_date, DOWNLOAD_FROM.ago.to_date].min`
?

spec/unit/tariff_synchronizer/base_update_spec.rb
```diff
@@ -40,12 +40,76 @@
...
   end
 
+  describe '.applicable_download_date_range' do
+    shared_examples_for 'an applicable download date range' do |update_factory|
+      let(:today) { Time.zone.today }
+      let(:pending_issue_date) { today - 21.days }
+      let(:applied_issue_date) { today - 22.days }
+      let(:failed_issue_date) { today - 23.days }
+
+      context 'when choosing a pending update older than the default download from date' do
```
alexdesi, line 50: what about being more explicit here?
`when a pending update is older than 20 days ago?`

If we update the constant, the tests will break and we need to update them, and the description.

spec/unit/tariff_synchronizer/base_update_spec.rb
```diff
@@ -40,12 +40,76 @@
...
+
+      context 'when choosing a failed update older than the default download from date' do
+        before do
+          create(update_factory, :failed, issue_date: failed_issue_date)
+        end
+
+        it { is_expected.to eq(failed_issue_date..today) }
+      end
+
+      context 'when there is an update issued 20 days ago or less' do
```
alexdesi, line 77: this is superclear 👍

spec/unit/tariff_synchronizer/base_update_spec.rb
willfish: I'd rather not update the context description as well as the constant personally - more to maintain.

spec/unit/tariff_synchronizer/base_update_spec.rb
```diff
@@ -40,12 +40,76 @@
...
+    it_behaves_like 'an applicable download date range', :cds_update do
+      subject(:applicable_download_date_range) { TariffSynchronizer::CdsUpdate.applicable_download_date_range }
+
+      let(:initial_date) { Date.new(2020, 9, 1) }
+    end
+  end
+
   describe '.sync' do
-    it 'Calls the download method for each date since the last issue_date to the current date' do
+    it 'calls the download method for each date since for the last 20 days to the current date' do
```
alexdesi, line 106: is "..since for.." correct?

willfish: No it isn't. I'll correct it

spec/unit/tariff_synchronizer/base_update_spec.rb
willfish: We could stub the constant and then we can change it independent of the test. :thinking:

app/lib/tariff_synchronizer/base_update.rb
willfish: Yeah, it is the same. Not sure which is clearer. I guess yours is terser.

app/lib/tariff_synchronizer/base_update.rb
alexdesi: ok, I got it

alexdesi (approved):
💚 💚 💚

PR #362, opened 2022-01-20
HOTT-1286: Remove creation of missing updates

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Removed creating missing updates when downloading Taric files

### Why?

I am doing this because:

- This isn't helpful and in fact creates noise in the process of
reviewing files which haven't been downloaded correctly.


alexdesi (approved):
👍

PR #364, opened 2022-01-21
HOTT-1252: Fix new scopes for update model

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1252

### What?

I have added/removed/altered:

- [x] Added coverage for existing used base update model scopes
- [x] Tweaked scope naming to reflect chronology rather than couple to order
- [x] Tweaked scope behaviour to reflect pulling out earliest updates

### Why?

I am doing this because:

- These break the cds sequence checker and therefore are holding up applies

### DIFF
--- a/app/lib/tariff_synchronizer/base_update.rb
+++ b/app/lib/tariff_synchronizer/base_update.rb
@@ -77,15 +77,19 @@ module TariffSynchronizer
       end
 
       def last_pending
-        pending.order(:issue_date).first
+        pending.ascending.first
       end
 
-      def last_applied
-        applied.order(:issue_date).first
+      def most_recent_applied
+        applied.descending.first
       end
 
-      def last_failed
-        failed.order(:issue_date).first
+      def most_recent_failed
+        failed.descending.first
+      end
+
+      def ascending
+        order(Sequel.asc(:issue_date))
       end
 
       def descending
@@ -173,7 +177,7 @@ module TariffSynchronizer
         if pending_applied_or_failed.count.zero?
           TariffSynchronizer.initial_update_date_for(update_type)
         else
-          last_download = (last_pending || last_applied || last_failed)
+          last_download = (last_pending || most_recent_applied || most_recent_failed)
 
           [last_download.issue_date, DOWNLOAD_FROM.ago.to_date].min
         end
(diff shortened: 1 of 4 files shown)


app/lib/tariff_synchronizer/base_update.rb
```diff
@@ -77,15 +77,19 @@ def applied_or_failed
       end
 
       def last_pending
-        pending.order(:issue_date).first
+        pending.ascending.first
```
alexdesi, line 80: yeah, I guess that's fine, it uses the ID so it should be the same as before.

app/lib/tariff_synchronizer/base_update.rb
```diff
@@ -77,15 +77,19 @@ def applied_or_failed
...
       end
 
-      def last_applied
-        applied.order(:issue_date).first
+      def most_recent_applied
```
alexdesi, line 83: I agree, that's explicit and clearer 👍

app/lib/tariff_synchronizer/base_update.rb
```diff
@@ -77,15 +77,19 @@ def applied_or_failed
       end
 
       def last_pending
```
alexdesi, line 79: I think we should rename this to
`most_recent_pending`
too.

app/lib/tariff_synchronizer/base_update.rb
alexdesi: oh, I just noticed, you added `def ascending`, that's fine

spec/unit/tariff_synchronizer/base_update_spec.rb
```diff
@@ -168,4 +168,65 @@
       end
     end
   end
+
+  describe '.last_pending' do
+    subject(:last_pending) { described_class.last_pending }
+
+    context 'when there are updates' do
+      before do
+        create(:cds_update, :pending, issue_date:  Time.zone.yesterday) # Target
```
alexdesi, line 177: what about using let on the top to dry these a little bit?
```
let(:today){ Time.zone.today)
let(:yesterday){ Time.zone.yesterday
```

alexdesi (approved):
✅ ✅ ✅

PR #366, opened 2022-01-21
Hott-1284 Add Slack notification.

alexdesi:
### Jira link

[HOTT-<1284](https://transformuk.atlassian.net/browse/HOTT-1284)

### What?

Add Slack notification in case of FailedUpdatesError


app/services/slack_notifier_service.rb
```diff
@@ -0,0 +1,34 @@
...
+        return
+      end
+
+      notifier.ping(message)
+    end
+
+  private
+
+    def notifier
+      @msg_notifier || slack_notifier
```
alexdesi, line 21: I am injecting the  @msg_notifier to simplify tests, even though I could probably use allowAnyInstanceOf (which is deprecated anyway)

app/services/slack_notifier_service.rb
```diff
@@ -0,0 +1,34 @@
+class SlackNotifierService
```
jebw, line 1: This is very much personal opinion but I'd suggest rather then having this as a static class, ie

```
SlackNotifierService.call
```

instead differentiate between 'action' data and maintained state data, eg

```
SlackNotifierService.new(msg_notifier).notify('my message to slack')
```

app/services/slack_notifier_service.rb
```diff
@@ -0,0 +1,34 @@
...
+  CHANNEL = '#tariffs-etl'.freeze
+  USERNAME = 'Trade Tariff BE'.freeze
```
jebw, line 3: What is BE? If this is backend, can it be changed to `Trade Tariff Backend` - that would remove the risk of a future developer not knowing what BE means at the cost of 5 letters.

jebw (dismissed):
Looks good to me - only comments are a code style suggestion and the abbreviation needs to be replaced.

app/lib/tariff_synchronizer.rb
```diff
@@ -324,5 +328,13 @@ def check_tariff_updates_failures
...
       raise FailedUpdatesError
     end
+  rescue FailedUpdatesError => e
+    notify_slack_app(e)
+
+    raise
+  end
+
+  def notify_slack_app(exception)
+    SlackNotifierService.call("Error #{exception.class}: #{exception.message}")
```
willfish, line 338: I don't think this passes the signature requirements of the call method?

jebw (changes requested):
Looks good to me - only comments are a code style suggestion and the abbreviation needs to be replaced.

Marked as needs changing because the PRs still got failing specs?

app/lib/tariff_synchronizer.rb
```diff
@@ -324,5 +328,13 @@ def check_tariff_updates_failures
                  file_names: BaseUpdate.failed.map(&:filename))
       raise FailedUpdatesError
     end
+  rescue FailedUpdatesError => e
+    notify_slack_app(e)
+
+    raise
+  end
+
+  def notify_slack_app(exception)
```
willfish, line 337: Should we add a test for this integration with the service? I suspect this isn't actually passing?

app/services/slack_notifier_service.rb
willfish: I wonder if it would be better to configure slack notifier as part of an initialiser for this gem?

willfish: That also feels like a better place to access environment variables

app/lib/tariff_synchronizer.rb
willfish: Just a light touch one:

allow(slack_notifier_service).to receive(:call)
expect(slack_notifier_service).to have_receive(:call).with(expected_message)

app/services/slack_notifier_service.rb
alexdesi: it sounds good to me, my choice was influenced by the way we do services. I'll give it a try.

config/initializers/slack_notifier.rb
```diff
@@ -0,0 +1,11 @@
+require 'slack-notifier'
+
+if ENV['SLACK_WEB_HOOK_URL'] && ENV['SLACK_CHANNEL'] && ENV['SLACK_USERNAME']
```
willfish, line 3: Maybe just add handling for production only env and for the tests you can construct a specific double/mock/stub in the spec helper?

return unless Rails.env.production?

spec/services/slack_notifier_service_spec.rb
```diff
@@ -0,0 +1,11 @@
+RSpec.describe SlackNotifierService do
+  before do
+    allow(Rails.application.config.slack_notifier).to receive(:ping)
```
willfish, line 3: Perfect :)

config/initializers/slack_notifier.rb
alexdesi: not sure, in that case we should probably move that to the `/environments/production.rb`

willfish: This is the same approach you've taken elsewhere, though https://github.com/trade-tariff/trade-tariff-backend/commit/052c310553c107a7ebf5381e0532ac3c1f487df0

willfish: I think when you're configuring a specific gem regardless of how we want it behave the approach you've taken elsewhere is the correct one.

willfish (approved):

willfish:
A couple of the lint violations are to do  with the files we've added :smile:

willfish (approved):

app/services/slack_notifier_service.rb
alexdesi: I made some changes, now the notifier is stored in the Rails.application.config.slack_notifier.

PR #367, opened 2022-01-21
HOTT-1290: Adds patched downloader for broken Taric file api

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1290

### What?

I have added/removed/altered:

- [x] Added patched download behaviour to enable downloading based on anticipating the next increment
- [x] Added feature flag for controlling the enabling/disabling of this feature
- [x] Added test coverage of both approaches

### Why?

I am doing this because:

- The taric api doesn't work as advertised
- I've patched this in a separate module to enable easier rollback later


app/lib/tariff_synchronizer/taric_update_downloader.rb
```diff
@@ -1,5 +1,3 @@
-require 'tariff_synchronizer/taric_file_name_generator'
```
willfish, line 1: We use Zeitwerk so don't need this require

app/lib/tariff_synchronizer/taric_file_name_generator_patched.rb
```diff
@@ -0,0 +1,14 @@
+module TariffSynchronizer
+  # TODO: Remove once Taric has fixed their api
+  class TaricFileNameGeneratorPatched
```
alexdesi, line 3: this class does not do much, I wonder if we can remove this and add a method in `TariffSynchronizer`

app/lib/tariff_synchronizer/taric_update.rb
```diff
@@ -31,14 +45,33 @@ def correct_sequence_pair?(next_update, previous_update)
...
+      def applicable_updates
+        # Pull out the most recent update
+        current_update = most_recent_pending || most_recent_applied
+
+        # Bail if there isn't one (impossible from where the data is now)
+        return [] if current_update.blank?
+        # Bail if we need to correct with a manual rollback
+        return [] unless correct_filename_sequence?
+
+        accumulator = []
```
alexdesi, line 61: maybe rename this to `updates` or something less generic?

app/lib/tariff_synchronizer/taric_update.rb
```diff
@@ -31,14 +45,33 @@ def correct_sequence_pair?(next_update, previous_update)
...
+
+        SEQUENCE_APPLICABLE_UPDATE_LIMIT.times do
+          current_update = new(
+            filename: current_update.next_update_sequence_update_filename,
+            issue_date: current_update.next_update_issue_date,
+          )
+
+          accumulator << current_update.dup
```
alexdesi, line 69: it seems we do not need `.dup` in this case

app/lib/tariff_synchronizer/taric_update.rb
willfish: This is a common patter for accumulating in iterators

app/lib/tariff_synchronizer/taric_update.rb
willfish: We should do it, yeah

app/lib/tariff_synchronizer/taric_file_name_generator_patched.rb
willfish: I'm going to leave it since the TariffSynchronizer is multi-purpose and we'd end up with three public methods and a bunch of private methods that is probably beyond the scope of this task

willfish: It would be a halfway-house refactor

app/lib/tariff_synchronizer/taric_update.rb
willfish: Removed. I don't think Sequel::Model implements initailize_dup or initialize_copy callbacks anyway so this basically does nothing useful. Just an end of day brain fart

app/lib/tariff_synchronizer/taric_file_name_generator_patched.rb
willfish: I've opted to move these into the downloader/reduce the number of seemingly arbitary objects we instantiate. Ultimately I totally agree this is very greedy about what the TariffSynchronizer knows and should be there. For the moment I can kill one evil

app/lib/tariff_synchronizer/taric_update.rb
```diff
@@ -1,11 +1,25 @@
...
+        dates_or_updates.each { |date_or_update| download(date_or_update) }
+      end
+
+      def download(date_or_update)
+        if TradeTariffBackend.patch_broken_taric_downloads?
+          TaricUpdateDownloaderPatched.new(date_or_update).perform
+        else
+          TaricUpdateDownloader.new(date_or_update).perform
+        end
+      end
```
alexdesi, line 21: What about drying this up a bit? :

```
def download(date_or_update)
  taric_update_downloader_klass.new(date_or_update).perform
end

private

def taric_update_downloader_klass
 TradeTariffBackend.patch_broken_taric_downloads? ?  TaricUpdateDownloaderPatched :  TaricUpdateDownloader     
end
```

spec/unit/tariff_synchronizer/taric_update_spec.rb
```diff
@@ -112,8 +137,115 @@
...
+          { 'filename' => '2021-12-11_TGB21211.xml', 'issue_date' => '2021-12-11' },
+          { 'filename' => '2021-12-12_TGB21212.xml', 'issue_date' => '2021-12-12' },
+          { 'filename' => '2021-12-13_TGB21213.xml', 'issue_date' => '2021-12-13' },
+        ]
+      end
+
+      it { is_expected.to eq(expected_updates) }
+    end
+
+    context 'when there are unbroken sequence updates but only applied updates' do
```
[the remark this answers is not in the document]
alexdesi: or maybe, even better,  "of" 😄 :

'when there are unbroken sequence of applied updates'

app/lib/tariff_synchronizer/taric_update_downloader.rb
```diff
@@ -62,5 +60,17 @@ def update_or_create(state, file_name)
     def log_request_to_taric_update
       instrument('get_taric_update_name.tariff_synchronizer', date: date, url: url)
     end
+
+    def date_api_url
+      sprintf(taric_query_url_template, host: host, date: date.strftime('%Y%m%d'))
+    end
+
+    def file_api_urls
```
alexdesi, line 68: I cannot find a test for this, I think we should write one.

willfish: Multiple times, actually

alexdesi (approved):
✅

PR #375, opened 2022-01-26
Upgraded circleci provisioned remote docker

jebw:
### Jira link

HOTT-???

### What?

I have added/removed/altered:

- [x] Upgraded the version of the remote docker provisioned by Circle

### Why?

I am doing this because:

- its taking a long time to provision and I suspect this is because we're using a version falling into disuse
- later versions may also potentially help us avoid any already fixed bugs in docker

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -234,7 +234,7 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker:
-          version: 20.10.2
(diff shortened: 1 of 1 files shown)


willfish:
@jebw I think this is fixed here: https://github.com/trade-tariff/trade-tariff-backend/pull/374

willfish:
If you're happy we can just undo the revert

PR #376, opened 2022-01-26
HOTT-1250: Fixes broken heading api

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1305

### What?

When the heading hasn't been preloaded, we try to serialize using the
json api fast serializer a Mash which calls Array(our_mash) on the
intermediate serialized result. We have an intermediate result because
we use a different structure for the ES cache DDL to the JSON API
schema.

We have a special snowflake Mash implementation that was nested in the
SearchClient and generically named Response. This was implemented
because of an incompatibility with the JSON API gem we're using where
Array() implicitly calls to to_a on the Hashie::Mash and creates the
wrong result for includes.

I've moved this Mash implementation, added specs for it and made sure
that we can properly serialize all Mashes we instantiate.

I have added/removed/altered:

- [x] Renamed and added tests for our custom implementation of Hashie::Mash
- [x] Fixed a broken controller test which wasn't properly instantiating a chapter and therefore not implicitly verifying includes

### Why?

I am doing this because:

- So the api works for headings regardless of the state of ES cache

### DIFF
--- a/app/lib/hashie/tariff_mash.rb
+++ b/app/lib/hashie/tariff_mash.rb
@@ -1,0 +1,11 @@
+module Hashie
+  class TariffMash < Hashie::Mash
+    disable_warnings
+
+    # Need to wrap object in array because serializer gem does Array(obj) and it breaks Hashie::Mash object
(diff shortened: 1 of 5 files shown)


alexdesi (approved):
✅

PR #379, opened 2022-01-27
HOTT-1279 validity dates api for commodities and headings

jebw:
### Jira link

[HOTT-1279](https://transformuk.atlassian.net/browse/HOTT-1279)

### What?

I have added/removed/altered:

- [x] Added an API to query the validity dates for a given commodity or heading code

### Why?

I am doing this because:

- we wish to show the dates when commoditiies and headings are valid for on their relevant 404 pages on the frontend


app/controllers/api/v2/validity_dates_controller.rb
```diff
@@ -0,0 +1,31 @@
+module Api
+  module V2
+    class ValidityDatesController < ApiController
```
willfish, line 3: So this isn't really resourceful. How do the jsonapi docs say we should do this? Should we just surface the validity dates on the resources as they are currently?

willfish (approved):

willfish (approved):

PR #382, opened 2022-01-27
HOTT-1302 Adds a new endpoint to get all the certification codes.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1302

### What?
This adds a new API Endpoint that lists the document codes that are available at any given time, according to their validity start and end dates.


app/serializers/api/v2/certificates/certificates_serializer_list.rb
```diff
@@ -0,0 +1,23 @@
+module Api
+  module V2
+    module Certificates
+      class CertificatesSerializerList
```
alexdesi, line 4: I added a new serializer in order to avoid conditional attributes (it require passing in param attributes, which makes things hard to read: https://github.com/jsonapi-serializer/jsonapi-serializer#conditional-attributes)

app/models/certificate.rb
```diff
@@ -14,17 +14,28 @@ class Certificate < Sequel::Model
...
+                                              primary_key: %i[certificate_type_code]
+
   many_to_many :measures, join_table: :measure_conditions,
                           left_key: %i[certificate_code certificate_type_code],
                           right_key: :measure_sid
 
+  one_to_many :certificate_types, key: :certificate_type_code,
+                                  primary_key: :certificate_type_code do |ds|
+    ds.with_actual(CertificateType)
+  end
```
alexdesi, line 27: Note: I did not add this,  it's just the github file comparison

app/models/certificate.rb
```diff
@@ -14,17 +14,28 @@ class Certificate < Sequel::Model
   one_to_many :measure_conditions, key: %i[certificate_type_code certificate_code],
                                    primary_key: %i[certificate_type_code certificate_code]
 
+  one_to_many :certificate_type_descriptions, key: %i[certificate_type_code],
+                                              primary_key: %i[certificate_type_code]
+
```
alexdesi, line 19: This is used by new serializer.

app/serializers/api/v2/certificates/certificates_serializer_list.rb
willfish, line 4: Probably I'd rename the other CertificateSerializer to CertificateSearchSerizalizer and steal that. If this is more useful (and I think it probably is) we'll probably want to use this in the future

spec/controllers/api/v2/certificates_controller_index.spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      certificate_type
+    end
+
+    it 'returns the list of certificates' do
+      get :index, format: :json
+
+      expect(response.body).to match_json_expression expected_response
+    end
+
+    context 'when `as_of` param is specified' do
```
willfish, line 46: So even if the as_of value is not specified the TimeMachine is called, BTW

willfish: It gets called with Time.zone.today

spec/controllers/api/v2/certificates_controller_search.spec.rb
```diff
@@ -1,5 +1,5 @@
 RSpec.describe Api::V2::CertificatesController, type: :controller do
-  context 'certificates search' do
+  describe '#search' do
```
willfish, line 2:
```suggestion
  describe 'GET #search' do
```

spec/controllers/api/v2/certificates_controller_index.spec.rb
```diff
@@ -0,0 +1,56 @@
+RSpec.describe Api::V2::CertificatesController, type: :controller do
+  describe '#index' do
```
willfish, line 2:
```suggestion
  describe 'GET #index' do
```

app/controllers/api/v2/certificates_controller.rb
```diff
@@ -1,14 +1,23 @@
...
 
       def search
         options = {}
         options[:include] = [:measures, 'measures.goods_nomenclature']
         render json: Api::V2::Certificates::CertificatesSerializer.new(@certificates, options.merge(serialization_meta)).serializable_hash
       end
 
+      def index
+        certificates = Certificate.actual
+                                  .eager(:certificate_descriptions)
```
willfish, line 14:
```suggestion
                                  .eager(:certificate_descriptions, :certificate_type_descriptions)
```

willfish: Still had n+1s here

app/models/certificate.rb
```diff
@@ -14,17 +14,28 @@ class Certificate < Sequel::Model
...
 
-  one_to_many :certificate_types, key: :certificate_type_code,
-                                  primary_key: :certificate_type_code do |ds|
-    ds.with_actual(CertificateType)
+  def certificate_type_description
+    certificate_type_descriptions.first
+  end
+
+  def certificate_description_period
+    CertificateDescriptionPeriod[certificate_description.certificate_description_period_sid]
```
willfish, line 38: This feels and reads wrong :eyes:

willfish:
We should add a spec for the serializer.

app/models/certificate.rb
willfish: All descriptions are fetched by the applicable period that is relevant. We should create this relationship so that it follows the same pattern as certificate_descriptions

app/serializers/api/v2/certificates/certificates_serializer_list.rb
```diff
@@ -0,0 +1,23 @@
...
+        set_id :id
+
+        attributes :certificate_type_code, :certificate_code, :description, :formatted_description
+
+        attribute :certificate_type_description do |certificate|
+          certificate&.certificate_type_description&.description
+        end
+
+        attribute :validity_start_date do |certificate|
+          certificate&.certificate_description_period&.validity_start_date
```
willfish, line 18:
```suggestion
          certificate&.certificate_description&.certification_descriptions_period&&.validity_start_date
```

app/controllers/api/v2/certificates_controller.rb
alexdesi: I agree but that does not work and that's why I created that new ticket, it triggers an obscure error message not easy to fix.

app/serializers/api/v2/certificates/certificates_serializer_list.rb
alexdesi: I agree with the renaming (certificates_list_serializer and certificates_search_serializer),
about the `stealing` do you mean, reusing it?

In this case some duplication is useful to avoid ugly conditional attributes (in the serializers and where those are called)

app/controllers/api/v2/certificates_controller.rb
willfish: Ah that's because its singular

willfish:
```suggestion
                                  .eager(:certificate_descriptions, :certificate_type_description, :certificate_description_periods)
```

app/models/certificate.rb
```diff
@@ -14,17 +14,28 @@ class Certificate < Sequel::Model
   one_to_many :measure_conditions, key: %i[certificate_type_code certificate_code],
                                    primary_key: %i[certificate_type_code certificate_code]
 
+  one_to_many :certificate_type_descriptions, key: %i[certificate_type_code],
```
willfish, line 17: many_to_one

app/models/certificate.rb
```diff
@@ -14,17 +14,28 @@ class Certificate < Sequel::Model
...
+                                              primary_key: %i[certificate_type_code]
+
   many_to_many :measures, join_table: :measure_conditions,
                           left_key: %i[certificate_code certificate_type_code],
                           right_key: :measure_sid
 
+  one_to_many :certificate_types, key: :certificate_type_code,
```
willfish, line 24: This is actually many_to_one

willfish: Not that you added this :)

app/models/certificate.rb
```diff
@@ -14,17 +14,28 @@ class Certificate < Sequel::Model
...
+  end
+
   def certificate_description
     certificate_descriptions.first
   end
 
-  one_to_many :certificate_types, key: :certificate_type_code,
-                                  primary_key: :certificate_type_code do |ds|
-    ds.with_actual(CertificateType)
+  def certificate_type_description
```
willfish, line 33: Redundant when relationship is fixed

app/serializers/api/v2/certificates/certificates_serializer_list.rb
willfish: Yeah, so this becomes CertificateSerializer

app/models/certificate.rb
alexdesi: I think Matt said that is not the case.
So that should be fine.

app/models/certificate.rb
alexdesi: Removed, and added the new relationship to `certificate_description_period`

app/serializers/api/v2/certificates/certificates_serializer_list.rb
alexdesi: `certificate.certificate_description_period.validity_start_date`

app/controllers/api/v2/certificates_controller.rb
alexdesi: this is still a performance issue

willfish (approved):

PR #384, opened 2022-01-31
HOTT-1308: Surface subheading for redirect in search

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1308

### What?

I have added/removed/altered:

- [x] Added search reference endpoint exact match response for subheadings

### Why?

I am doing this because:

- This is so that we can redirect to subheadings with search references


alexdesi (approved):
We've discussed all the points ✅

PR #396, opened 2022-02-08
Hott 1297 Add measures types api endpoint

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1297

### What?
Add a new endpoint to list all the measure_types and relevant measure_type_series_description

### Why?
The members of the trade consider useful this endpoint.

### DIFF
--- a/app/controllers/api/admin/measure_types_controller.rb
+++ b/app/controllers/api/admin/measure_types_controller.rb
@@ -6,13 +6,13 @@ module Api
       def index
         @measure_types = MeasureType.eager(:measure_type_description).national.all
 
-        render json:  Api::Admin::MeasureTypeSerializer.new(@measure_types).serializable_hash
+        render json: Api::Admin::MeasureTypeSerializer.new(@measure_types).serializable_hash
       end
 
       def show
         @measure_type = MeasureType.national.with_pk!(measure_type_pk)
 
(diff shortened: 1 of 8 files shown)


app/controllers/api/admin/measure_types_controller.rb
```diff
@@ -12,7 +12,7 @@ def index
       def show
         @measure_type = MeasureType.national.with_pk!(measure_type_pk)
 
-        render json:  Api::Admin::MeasureTypeSerializer.new(@measure_type, { is_collection: false }).serializable_hash
```
willfish, line 15: Probably easier to just remove this

app/serializers/api/v2/measures/measure_type_serializer.rb
```diff
@@ -8,10 +8,18 @@ class MeasureTypeSerializer
...
+                   :order_number_capture_code,
+                   :trade_movement_code,
+                   :validity_end_date,
+                   :validity_start_date
 
-        attribute :id do |measure_type|
-          measure_type.measure_type_id
+        attribute :id, &:measure_type_id
+
+        attribute :measure_type_series_description do |measure_type|
```
willfish, line 21: The pattern we use elsewhere is to have this description delegator on the model.

PR #398, opened 2022-02-09
Optimize headings

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Added presenters for the heading endpoint that use eager to preload all of the relationships for each heading commodity
- [ ] Added worker to preload headings into redis as part of our morning ETL

### Why?

I am doing this because:

- We preload all of the heading as part of the morning ETL currently in elastic search. As far as I can see there is no need for this extra layer - we should be able to preload expensive headings in redis

### DIFF
--- a/app/models/concerns/declarable.rb
+++ b/app/models/concerns/declarable.rb
@@ -3,7 +3,7 @@ module Declarable
   include Formatter
 
   included do
-    one_to_many :measures, primary_key: {}, key: {}, dataset: -> {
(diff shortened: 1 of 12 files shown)


willfish:
Despite this being a smidge faster we're not supporting dynamic reloading of the underlying resources based on the time machine functionality. We need this in order to give users the most accurate information. I need to go back to the drawing board on this.

PR #400, opened 2022-02-09
HOTT-1244: Update downloads to handle long rollbacks in XI

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1244

### What?

I have added/removed/altered:

- [x] Altered patch download behaviour to continue downloading until we stop finding files
- [x] Altered existing tests for the new patched download behaviour

### Why?

I am doing this because:

- We are about to do a long rollback and whilst we can't distinguish between an actual 404 and a new year/or a genuinely non present file we can bail at the point the next increment isn't there. This will enable me to rollback a long way and bring us up-to-date with the patched downloader a year at a time.

### DIFF
--- a/app/lib/tariff_synchronizer/taric_update.rb
+++ b/app/lib/tariff_synchronizer/taric_update.rb
@@ -7,13 +7,15 @@ module TariffSynchronizer
 
     class << self
       def sync
-        dates_or_updates = TradeTariffBackend.patch_broken_taric_downloads? ? applicable_updates : applicable_download_date_range
-
-        dates_or_updates.each { |date_or_update| download(date_or_update) }
+        applicable_download_date_range.each { |issue_date| download(issue_date) }
       end
 
-      def download(date_or_update)
-        downloader_class.new(date_or_update).perform
+      def sync_patched
+        TaricUpdateDownloaderPatched.new(applicable_update).perform
+      end
+
+      def download(issue_date)
+        TaricUpdateDownloader.new(issue_date).perform
       end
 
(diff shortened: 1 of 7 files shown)


app/lib/tariff_synchronizer/taric_update.rb
```diff
@@ -45,33 +47,25 @@ def update_type
...
-          current_update = new(
-            filename: current_update.next_update_sequence_update_filename,
-            issue_date: current_update.next_update_issue_date,
-          )
-
-          updates_to_check << current_update
-        end
-
-        updates_to_check
+        new(
```
willfish, line 59: Just use next update

willfish:
```suggestion
        current_update.next_update
```

spec/unit/tariff_synchronizer/taric_update_spec.rb
```diff
@@ -183,31 +179,20 @@
     it { is_expected.to eq('TGB21203.xml') }
   end
 
-  describe '.applicable_updates' do
-    subject(:applicable_updates) { described_class.applicable_updates.as_json }
+  describe '.applicable_update' do
+    subject(:applicable_update) { described_class.applicable_update.as_json }
```
alexdesi, line 183: does not it break when  `.applicable_update` returns nil?

alexdesi (approved):
✅

spec/unit/tariff_synchronizer/taric_update_spec.rb
willfish: Niet

PR #401, opened 2022-02-10
Add support for ?as_of param + eagerloading fix

alexdesi:
### Jira link
HOTT-1297

### What?
Add support for ?as_of param, and fix eagerloading on measure_type_series_description.


jebw (approved):
LGTM - would be better if this query was tested some how though?

Maybe add a `MeasureType.current_types_with_descriptions` class method on MeasureType - you could then test the beahviour in the context of time machine?

PR #403, opened 2022-02-11
Fix subheading footnotes

willfish:
### Jira link

HOTT-<TODO>

### What?
 
When there are footnotes for a subheading we want to present them under the footnotes in the UI. In the current implementation we were returning footnotes but the ids were never matching up/were using the wrong id.

I have added/removed/altered:

- [x] Altered the footnote serializers and presenters to consistently use the composite primary key that uniquely identifies the current footnote

### Why?

I am doing this because:

- We were getting a garbled jsonapi result which was breaking the frontend integration on subheadings with their own footnotes

### DIFF
--- a/app/models/footnote.rb
+++ b/app/models/footnote.rb
@@ -60,24 +60,9 @@ class Footnote < Sequel::Model
     end
   end
 
-  # FO4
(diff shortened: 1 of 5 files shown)


jebw (approved):
LGTM

PR #404, opened 2022-02-11
HOTT-1270 Use a connection pool for RedLock

jebw:
### Jira link

[HOTT-1270](https://transformuk.atlassian.net/browse/HOTT-1270)

### What?

I have added/removed/altered:

- [x] Added a connection pool for Redlock

### Why?

I am doing this because:

Whilst working on the CDS rollback and reapply a few weeks ago we observed that scheduling a rollback didn't work, but did then subsequently ran much later on. I had observed similar behaviour whilst dealing with another CDS rollback some months prior but I'd only just started on the project and assumed I was doing something wrong.

The issue turns out to be redlock hanging waiting to acquire a lock - I can reliably reproduce this by scheduling a rollback, then a roll forward. This will result in hang between 1 and 3 iterations - killing sidekiq then causes the job to run on shutdown. Potentially much later then expected.

I believe this is caused by our Redlock configuration which is not thread safe but being used in a Sidekiq environment which is threaded. We use the Sidekiq redis client - but that is a single client, not a pool of clients. Redlock can work with a pool if provided with one (and if given a single redis client it actually monkey patches it to behave like its a connection pool).

- This impacts the critical nightly sync job so should be well tested prior to roll out to production - plan is to leave it on DEV over the weekend and see how it fairs.

### Notes

I believe this is also the underlying issue affecting the CDS sync retry code.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ gem 'aws-sdk-s3'
 gem 'rubyzip'
 
 # Background jobs
+gem 'connection_pool'
 gem 'redis-rails'
 gem 'redlock'
 gem 'sidekiq'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -458,6 +458,7 @@ DEPENDENCIES
   aws-sdk-s3
   bootsnap
   brakeman
+  connection_pool
   curb
   database_cleaner-sequel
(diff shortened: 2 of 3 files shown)


willfish (approved):
LGTM

PR #405, opened 2022-02-11
HOTT-1301: Surface measure actions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1301

### What?

I have added/removed/altered:

- [x] Adds measure action controller, routes, serializers and relevant specs

### Why?

I am doing this because:

- This is a feature request

### DIFF
--- a/app/controllers/api/v2/measure_actions_controller.rb
+++ b/app/controllers/api/v2/measure_actions_controller.rb
@@ -1,0 +1,18 @@
+module Api
+  module V2
+    class MeasureActionsController < ApiController
+      def index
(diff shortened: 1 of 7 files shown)


test.json
```diff
@@ -0,0 +1,9068 @@
+{
```
jebw: This file looks like it was added by mistake ?

jebw (changes requested):
Looks good, works great. Suspect it needs the extraneous test.json file dropping from the PR though :+1:

jebw (approved):
LGTM

PR #406, opened 2022-02-11
HOTT-1299: Surface measure condition codes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1299

### What?

I have added/removed/altered:

- [x] Added controller, routes, serializers and specs for surfacing measure condition codes

### Why?

I am doing this because:

- This is a feature request


jebw (approved):
LGTM

PR #414, opened 2022-02-15
HOTT-1313: Fix intermittent geographical area failures

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1313

### What?

I have added/removed/altered:

- [x] Added spec cache clear before every geographical area controller request

### Why?

I am doing this because:

- We seem to get the wrong response from the cache intermittently


alexdesi (approved):
👍

willfish:
So for some background, I was able to reproduce (though intermittently) this issue by running the specs for the geographical areas controller 50 times. When I added the cache clearing the spec failures went away.

PR #417, opened 2022-02-15
HOTT-1341: Avoid collisions in meursing additional code values

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1341

### What?

I have added/removed/altered:

- [x] Added around_action hook to safely manage meursing additional codes

### Why?

I am doing this because:

- This is required to avoid colliding meursing additional code behaviour

### DIFF
--- a/app/controllers/api/v2/commodities_controller.rb
+++ b/app/controllers/api/v2/commodities_controller.rb
@@ -2,7 +2,7 @@ module Api
   module V2
     class CommoditiesController < ApiController
       before_action :find_commodity, only: %i[show changes]
-      before_action :set_meursing_additional_code, only: :show
+      around_action :configure_meursing_additional_code, only: :show
 
       def show
         render json: cached_commodity
@@ -50,8 +50,12 @@ module Api
         end
       end
 
-      def set_meursing_additional_code
+      def configure_meursing_additional_code
         Thread.current[:meursing_additional_code_id] = filter_params[:meursing_additional_code_id]
+
+        yield
+      ensure
+        Thread.current[:meursing_additional_code_id] = nil
       end
     end
   end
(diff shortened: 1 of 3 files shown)


app/controllers/api/v2/commodities_controller.rb
```diff
@@ -2,7 +2,7 @@ module Api
   module V2
     class CommoditiesController < ApiController
       before_action :find_commodity, only: %i[show changes]
-      before_action :set_meursing_additional_code, only: :show
+      around_action :configure_meursing_additional_code, only: :show
```
alexdesi, line 5: I see that this is not just a simple set, but I think it's fine and clearer to keep the name as before: `set_meursing_additional_code`

spec/controllers/api/v2/commodities_controller_spec.rb
```diff
@@ -93,10 +93,32 @@
...
+          :with_meursing_measures,
+          :declarable,
+        )
+      end
+
+      before do
+        Thread.current[:meursing_additional_code_id] = 'foo'
+      end
+
+      it 'sets the value to nil when we are done making the request' do
```
alexdesi, line 112: small wording improvement:

sets the value to nil when the request completes
or
sets the value to nil when the request  is completed

spec/factories/goods_nomenclature_factory.rb
```diff
@@ -50,6 +50,46 @@
...
+        )
+        # Placeholder meursing measure component - agricultural component
+        create(
+          :measure_component,
+          :with_duty_expression,
+          measure_sid: root_measure.measure_sid,
+          duty_expression_id: '12',
+        )
+
+        root_measure.reload
```
alexdesi, line 79: it seems we are not using `root_measure` this later.

alexdesi (approved):
It looks good, 
I added minor comments.

app/controllers/api/v2/commodities_controller.rb
willfish: I'm actually following the same approach as the time machine, basically. I could extract this into a MuersingConfigurator or equivalent at some point

PR #419, opened 2022-02-16
HOTT-1342: Adjust balance to initial volume when no balance events

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1342

### What?

I have added/removed/altered:

- [x] Altered QuotaDefinition#balance such that it uses the initial_volume when there are no balance events

### Why?

I am doing this because:

- This is the correct behaviour


alexdesi (approved):
✅

PR #420, opened 2022-02-17
HOTT-1343: Filter regulation measures correctly

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1343

### What?

I have added/removed/altered:

- [x] Extended distinct regulation filters to actual filter based off of the role/roles that the generating regulation for each union should have
- [x] Adds specs to cover the case where we have multiple base regulations with different roles as well as filtering on modification regulations
- [x] Tidied the ordering of the unioned measures in the declarable implementation

### Why?

I am doing this because:

- We currently show measures that should be end dated because we're not filtering on regulation role and are picking the wrong regulation effective end date
- We need to make sure the base regulation filter includes all possible role ids (this caught us out recently)

## AC

- [x] Representative price no longer pulls back the base_regulation for the effective end date (effectively its gone from the UK data) https://www.trade-tariff.service.gov.uk/commodities/1703100000
- [x] Trade remedies display properly https://www.trade-tariff.service.gov.uk/commodities/1516209821

### DIFF
--- a/app/models/concerns/declarable.rb
+++ b/app/models/concerns/declarable.rb
@@ -2,6 +2,19 @@ module Declarable
   extend ActiveSupport::Concern
   include Formatter
 
+  COMMON_UNION_MEASURE_ORDER = [
+    Sequel.desc(:measures__measure_generating_regulation_id),
+    Sequel.desc(:measures__measure_generating_regulation_role),
+    Sequel.desc(:measures__measure_type_id),
+    Sequel.desc(:measures__goods_nomenclature_sid),
+    Sequel.desc(:measures__geographical_area_id),
+    Sequel.desc(:measures__geographical_area_sid),
+    Sequel.desc(:measures__additional_code_type_id),
+    Sequel.desc(:measures__additional_code_id),
+    Sequel.desc(:measures__ordernumber),
+    Sequel.desc(:effective_start_date),
+  ].freeze
+
   included do
     one_to_many :measures, primary_key: {}, key: {}, dataset: -> {
       Measure.join(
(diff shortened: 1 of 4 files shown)


app/models/concerns/declarable.rb
```diff
@@ -2,36 +2,49 @@ module Declarable
   extend ActiveSupport::Concern
   include Formatter
 
+  COMMON_UNION_MEASURE_ORDER = [
+    Sequel.desc(:measures__measure_generating_regulation_id),
+    Sequel.desc(:measures__measure_generating_regulation_role),
```
willfish, line 7: Now that we're doing a distinct on the role, postgres told me I needed to add the corresponding ordering for the role

app/models/concerns/declarable.rb
```diff
@@ -2,36 +2,49 @@ module Declarable
...
+  ].freeze
+
   included do
     one_to_many :measures, primary_key: {}, key: {}, dataset: -> {
       Measure.join(
         Measure.with_base_regulations
                .with_actual(BaseRegulation)
                .where({ measures__goods_nomenclature_sid: uptree.map(&:goods_nomenclature_sid) })
                .where { Sequel.~(measures__measure_type_id: MeasureType.excluded_measure_types) }
-               .order(Sequel.desc(:measures__measure_generating_regulation_id), Sequel.desc(:measures__measure_type_id), Sequel.desc(:measures__goods_nomenclature_sid), Sequel.desc(:measures__geographical_area_id), Sequel.desc(:measures__geographical_area_sid), Sequel.desc(:measures__additional_code_type_id), Sequel.desc(:measures__additional_code_id), Sequel.desc(:measures__ordernumber), Sequel.desc(:effective_start_date))
```
willfish, line 12: I'm not changing these, I'm consolidating so I can make the same change in one place

app/models/measure.rb
```diff
@@ -280,6 +276,26 @@ def expired_before(candidate_measure)
     def non_invalidated
       where(measures__invalidated_at: nil)
     end
+
+    def distinct_measure_regulation_query(role:)
+      query = if model.point_in_time.present?
+                distinct(
+                  :measure_generating_regulation_id,
+                  :measure_generating_regulation_role,
```
willfish, line 284: Added this

spec/models/measure_spec.rb
```diff
@@ -1224,4 +1224,54 @@
...
     it { expect(measure.meursing_measures).to eq(meursing_measures) }
   end
+
+  shared_context 'with regulation measures' do
+    around do |example|
+      TimeMachine.now { example.run }
+    end
+
+    before do
+      create(:base_regulation, base_regulation_id: 'R9726580', base_regulation_role: 1)
```
willfish, line 1234: There are three types of base regulation (as indicated by BASE_REGULATION_ROLES). I'm filtering on these to make sure we get the correct regulation end date - we were getting nil because we were indifferent to the role and the role for base regulation with no end date came first

app/models/concerns/declarable.rb
```diff
@@ -2,36 +2,49 @@ module Declarable
   extend ActiveSupport::Concern
   include Formatter
 
+  COMMON_UNION_MEASURE_ORDER = [
```
alexdesi, line 5: 👍

alexdesi (approved):
lgtm ✅ ,
two minor comments.

spec/models/measure_spec.rb
```diff
@@ -1224,4 +1224,54 @@
...
+        measure_type_id: 1,
+        measure_generating_regulation_id: 'R9726580',
+        additional_code_type_id: 'F',
+        additional_code_id: '001',
+      }
+
+      # Base regulation measures
+      create(:measure, { measure_sid: 1, measure_generating_regulation_role: 1 }.merge(non_distinct_measure_opts))
+      create(:measure, { measure_sid: 2, measure_generating_regulation_role: 2 }.merge(non_distinct_measure_opts))
+      create(:measure, { measure_sid: 3, measure_generating_regulation_role: 3 }.merge(non_distinct_measure_opts))
```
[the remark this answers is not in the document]
willfish: I thought about, but it felt like I'd only use it once

PR #423, opened 2022-02-18
HOTT-1148 exclude measures without goods nomenclatures

jebw:
### Jira link

[HOTT-1148](https://transformuk.atlassian.net/browse/HOTT-1148)

### What?

I have added/removed/altered:

- [x] Filter out Measures which don't have a goods_nomenclature_id in the Additional Code search

### Why?

I am doing this because:

- Whilst the missing id (ie orphaned) normally makes the Measure inaccessible - it is still accessible via the Additional Code search, and trying to render that measure results in an error


app/serializers/cache/additional_code_serializer.rb
```diff
@@ -10,7 +10,10 @@ def initialize(additional_code)
     end
 
     def as_json
-      measures = additional_code.measures.select do |measure|
+      measures = additional_code.measures_dataset
+                                .exclude(goods_nomenclature_item_id: nil)
```
willfish, line 14: This will be meursing measures I expect :)

PR #425, opened 2022-02-21
HOTT-1298: Surface quota order numbers.

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1298

### What?

I have added/removed/altered:

- [x] Added api endpoint for surfacing active quota order numbers
- [x] Added coverage for this endpoint and its serializers
- [x] Added dataset module method for returning only active quota order numbers that have a quota definition

### Why?

I am doing this because:

- This is a feature request that will be used by Defra https://www.gov.uk/government/organisations/department-for-environment-food-rural-affairs

### AC

- [x] We need to make sure that these are only returned when there is a
corresponding quota definition for the given time machine specified
date.
- [x] We do not care about quota balance events for the moment since these can
be fetched using a different api (to come later)

### DIFF
--- a/spec/factories/quota_factory.rb
+++ b/spec/factories/quota_factory.rb
@@ -12,6 +12,11 @@ FactoryBot.define do
   end
 
   factory :quota_order_number do
+    transient do
+      quota_definition_sid { generate(:sid) }
+      quota_definition_validity_end_date { nil }
+    end
+
     quota_order_number_sid { generate(:quota_order_number_sid) }
     quota_order_number_id  { generate(:quota_order_number_id) }
     validity_start_date { Date.current.ago(4.years) }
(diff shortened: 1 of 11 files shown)


spec/factories/quota_factory.rb
```diff
@@ -20,6 +25,35 @@
     trait :xml do
       validity_end_date { Date.current.ago(1.year) }
     end
+
+    trait :current do
+      validity_end_date { nil }
+    end
+
+    trait :non_current do
```
jebw, line 33: `:historical` or `:expire` might make more sense here?

spec/factories/quota_factory.rb
```diff
@@ -20,6 +25,35 @@
...
+      validity_end_date { Time.zone.yesterday }
+    end
+
+    trait :current_definition do
+      quota_definition_validity_end_date { nil }
+    end
+
+    trait :non_current_definition do
```
jebw, line 41: as above?

spec/factories/quota_factory.rb
```diff
@@ -67,6 +101,8 @@
     monetary_unit_code              { Forgery(:basic).text(exactly: 3) }
     measurement_unit_code           { Forgery(:basic).text(exactly: 3) }
     measurement_unit_qualifier_code { generate(:measurement_unit_qualifier_code) }
+    validity_start_date             { Date.current.ago(4.years) }
```
jebw, line 104: This is a timestamp field - `4.years.ago` might make more sense but no strong feelings

jebw (approved):
LGTM - only comment would be around trait naming in quota_factory?

spec/factories/quota_factory.rb
willfish: I prefer expire over historical. Happy to change it :)

PR #426, opened 2022-02-21
HOTT-1352: Move to consistent Date/Time api usage

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1352

### What?

I have added/removed/altered:

- [x] Move from using Date.current to Time.zone.today
- [x] Move from using Date.yesterday to Time.zone.yesterday
- [x] Move from using Date.current.ago(3.years) to 3.years.ago.beginning_of_day

### Why?

I am doing this because:

- For consistency
- For clarity on the result
- To follow community standards

### DIFF
--- a/app/lib/changes_table_populator/importer.rb
+++ b/app/lib/changes_table_populator/importer.rb
@@ -11,7 +11,7 @@ module ChangesTablePopulator
       ].freeze
       DB = Sequel::Model.db
 
-      def populate(day: Date.current)
+      def populate(day: Time.zone.today)
         elements = DB[source_table]
           .where(where_condition(day: day))
           .select &select_condition
@@ -21,7 +21,7 @@ module ChangesTablePopulator
           .import IMPORT_FIELDS, import_records(elements: elements, day: day)
       end
 
-      def populate_backlog(from: Date.current - 3.months, to: Date.current)
+      def populate_backlog(from: Time.zone.today - 3.months, to: Time.zone.today)
         from = from.to_date
         to = to.to_date
         (from..to).each do |day|
@@ -37,7 +37,7 @@ module ChangesTablePopulator
         raise NotImplementedError, 'Implement this method in the subclasses'
       end
 
-      def where_condition(day: Date.current) # rubocop:disable Lint/UnusedMethodArgument
+      def where_condition(day: Time.zone.today) # rubocop:disable Lint/UnusedMethodArgument
         raise NotImplementedError, 'Implement this method in the subclasses'
       end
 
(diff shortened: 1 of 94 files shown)


app/lib/changes_table_populator/importer.rb
```diff
@@ -21,7 +21,7 @@ def populate(day: Date.current)
           .import IMPORT_FIELDS, import_records(elements: elements, day: day)
       end
 
-      def populate_backlog(from: Date.current - 3.months, to: Date.current)
+      def populate_backlog(from: Time.zone.today - 3.months, to: Time.zone.today)
```
willfish, line 24: Possibly another candidate for 3.months.ago

spec/controllers/api/admin/rollbacks_controller_spec.rb
```diff
@@ -19,7 +19,7 @@
     it 'performs a rollback' do
       Sidekiq::Testing.inline! do
         expect {
-          create(:rollback, date: Date.current.ago(1.month).to_date)
+          create(:rollback, date: 1.month.ago.beginning_of_day)
```
willfish, line 22:
```suggestion
          create(:rollback, date: 1.month.ago.to_date)
```

app/models/change.rb
```diff
@@ -1,7 +1,7 @@
 class Change < Sequel::Model
   plugin :time_machine
 
-  def self.cleanup(older_than: Date.current.ago(3.months))
+  def self.cleanup(older_than: 3.months.ago.beginning_of_day)
```
jebw, line 4: for consistency this needs `.to_date` on the end to convert to a `Date` object? Not sure if it would make any difference but this does change from `Date` to `TimeWithZone`

spec/unit/changes_table_populator/measure_created_or_updated_spec.rb
```diff
@@ -16,7 +16,7 @@
       before do
         measure = create :measure
 
-        db.run("UPDATE measures_oplog SET operation = 'C', operation_date = '#{Date.current}' " \
+        db.run("UPDATE measures_oplog SET operation = 'C', operation_date = '#{Time.zone.today}' " \
```
jebw, line 19: These are broken in the original but this could do with being `Time.zone.today.to_formatted_s(:db)`

Currently this only works because the default date format in the app is the same as the `:db` format but if that gets changed all these specs will fail

jebw (approved):
LGTM - couple of suggestions around Dates that are getting changed to TimeWithZone - and spotted an existing issue around the update SQL in specs

spec/unit/changes_table_populator/measure_created_or_updated_spec.rb
willfish: Will fix in a follow on PR

PR #431, opened 2022-02-22
HOTT-1260 Retry downloading CDS

jebw:
### Jira link

[HOTT-1260](https://transformuk.atlassian.net/browse/HOTT-1260)

### What?

I have added/removed/altered:

- [x] Retry downloading CDS if todays file hasn't yet arrived, unless we're past a cut off time
- [x] Moves Taric sync back an hour to 5am

### Why?

I am doing this because:

- We want to try and import as early as possible but if the files are arriving later, then we should still be importing them

- Changes behaviour of overnight sync

### DIFF
--- a/app/workers/updates_synchronizer_worker.rb
+++ b/app/workers/updates_synchronizer_worker.rb
@@ -1,20 +1,50 @@
 class UpdatesSynchronizerWorker
   include Sidekiq::Worker
 
+  TRY_AGAIN_IN = 20.minutes
+  CUT_OFF_TIME = '07:30'.freeze
+
   sidekiq_options queue: :sync, retry: false
 
-  def perform
+  def perform(check_for_todays_file = true)
     logger.info 'Running UpdatesSynchronizerWorker'
     logger.info 'Downloading...'
 
     if TradeTariffBackend.uk?
       TariffSynchronizer.download_cds
-      logger.info 'Applying...'
-      TariffSynchronizer.apply_cds(reindex_all_indexes: true)
+
+      if check_for_todays_file &&
+          still_time_to_reschedule? &&
+          todays_file_has_not_yet_arrived?
+
+        self.class.perform_in(TRY_AGAIN_IN, true)
+        logger.info "Daily file missing, retrying at #{TRY_AGAIN_IN.from_now}"
+      else
+        logger.info 'Applying...'
+        TariffSynchronizer.apply_cds(reindex_all_indexes: true)
+      end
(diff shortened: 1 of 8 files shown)


app/workers/updates_synchronizer_worker.rb
```diff
@@ -1,20 +1,50 @@
...
     end
   end
+
+private
+
+  def cut_off_date_time
+    @cut_off_date_time ||= begin
+      hour, minute = CUT_OFF_TIME.split(':', 2).map(&:to_i)
+
+      Time.zone.now.at_beginning_of_day + hour.hours + minute.minutes
```
willfish, line 39:
```suggestion
      Time.zone.now.beginning_of_day + hour.hours + minute.minutes
```

app/workers/updates_synchronizer_worker.rb
```diff
@@ -1,20 +1,50 @@
 class UpdatesSynchronizerWorker
   include Sidekiq::Worker
 
+  TRY_AGAIN_IN = 20.minutes
+  CUT_OFF_TIME = '07:30'.freeze
```
willfish, line 5: Should we put this in the environment? Might make it a bit easier to configure? Maybe calling it CDS_UPDATES_RETRY_UNTIL and CDS_UPDATES_RETRY_EVERY?

app/workers/updates_synchronizer_worker.rb
```diff
@@ -1,20 +1,50 @@
...
+
   sidekiq_options queue: :sync, retry: false
 
-  def perform
+  def perform(check_for_todays_file = true)
```
willfish, line 9: Should we just not make this optional?

willfish (dismissed):
Small configuration suggestion but otherwise this looks good to me

willfish (approved):
Small config suggestion but LGTM

app/workers/updates_synchronizer_worker.rb
jebw: I'm a bit wary of adding too much configuration - if its something we end up needing we can add it in env vars easily enough later on but considering the other timings are held in the codebase (`sidekiq.yml`) - I think these should also be in the codebase for now.

app/workers/updates_synchronizer_worker.rb
jebw: Done - wonder when that swapped as the default - years ago it was `at_beginning`, rails guides only mention `beginning_of` now  - showing my age :laughing:

PR #432, opened 2022-02-22
HOTT-1295: Pin rails to 7.0

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1295

### What?

I have added/removed/altered:

- [x] Pin rails to 7.0

### Why?

I am doing this because:

- We should control more closely breaking versions of rails
- Rails do not use semver and minor versions are breaking


alexdesi (approved):
✅ it sounds good.

PR #436, opened 2022-02-24
HOTT-1293: Refresh staging database

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1283

### What?

I have added/removed/altered:

- [x] Adds rake task to clear caches after a db refresh
- [x] Refresh staging database

### Why?

I am doing this because:

- We want to make sure staging looks/acts like production when we integrate/validate changes before they go live


alexdesi (approved):
✅

PR #441, opened 2022-02-28
HOTT-1337: Obfuscate errors to clients of the backend apis

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1337

### What?

I have added/removed/altered:

- [x] Added specs of current error propagation behaviour
- [x] Altered the internal server error behaviour to not include the exception body

### Why?

I am doing this because:

- We were missing coverage of this
- We need to make sure that internal details of the service do not leak to clients

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,11 +6,13 @@ class ApplicationController < ActionController::Base
   before_action :clear_association_queries
   around_action :configure_time_machine
 
-  rescue_from Exception,                          with: :render_internal_server_error
-  rescue_from ArgumentError,                      with: :render_bad_request
-  rescue_from Sequel::RecordNotFound,             with: :render_not_found
-  rescue_from ActionController::RoutingError,     with: :render_not_found
-  rescue_from AbstractController::ActionNotFound, with: :render_not_found
+  unless Rails.application.config.consider_all_requests_local
+    rescue_from Exception,                          with: :render_internal_server_error
+    rescue_from ArgumentError,                      with: :render_bad_request
+    rescue_from Sequel::RecordNotFound,             with: :render_not_found
+    rescue_from ActionController::RoutingError,     with: :render_not_found
+    rescue_from AbstractController::ActionNotFound, with: :render_not_found
+  end
 
   def render_not_found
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -6,13 +6,11 @@ class ApplicationController < ActionController::Base
   before_action :clear_association_queries
   around_action :configure_time_machine
 
-  unless Rails.application.config.consider_all_requests_local
```
jebw, line 9: I think this will confuse matters in development mode and prevent us from seeing the normal rails error message

app/controllers/application_controller.rb
```diff
@@ -6,13 +6,11 @@ class ApplicationController < ActionController::Base
...
-    rescue_from Exception,                          with: :render_internal_server_error
-    rescue_from ArgumentError,                      with: :render_bad_request
-    rescue_from Sequel::RecordNotFound,             with: :render_not_found
-    rescue_from ActionController::RoutingError,     with: :render_not_found
-    rescue_from AbstractController::ActionNotFound, with: :render_not_found
-  end
+  rescue_from Exception,                          with: :render_internal_server_error
```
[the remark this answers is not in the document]
jebw: TBH the general rescue is wrong but thats doesn't really impact the scope of this PR

jebw (changes requested):
Looks great but dropping the `consider_all_requests_local` will confuse things for other developers

app/controllers/application_controller.rb
jebw: Isn't this `unless` to make sure these exceptions are not rescued in dev mode, so falling through to the default rails exception handler app

Without it, the dev mode will operate the same as prod mode without the red exceptions screens?

willfish:
That's what I understood, yeah. It clumsily applies to test, though, and I
wanted to make sure we had coverage of this behaviour. Possibly we could
control this more explicitly via an env var which defaults to having these
enabled for test and prod but disables for dev
(RESCUE_CONTROLLER_EXCEPTIONS or equivalent).

jebw (approved):
LGTM :raised_hands:

PR #446, opened 2022-03-01
HOTT-1368: Separate EU/UK runs

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1368

### What?

I have added/removed/altered:

- [x] Move to separate synchronizers/workers for the CDS and Taric updates
- [x] Make sure all existing tests are ported over

### Why?

I am doing this because:

- We want to run/test and make changes to each synchronizer in isolation and avoid spaghetti

### Out of scope

- Significant behavioural refactor of the synchronizers (we could reuse a lot of the rollback functionality between the two workers for example - where on the Cds rollback we're not optimally using the indexes/have slow rollbacks)
- Really anything more than extraction of the synchronizers
- I'm leaving the removal of the old synchronizer until we're totally comfortable with this running on dev/staging


spec/unit/cds_synchronizer_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        create :cds_update, example_date: pending_date, filename: "tariff_dailyExtract_v1_#{pending_date.strftime('%Y%m%d')}T123456.gzip"
+      end
+
+      # rubocop:disable RSpec/MultipleExpectations
+      it 'raises wrong sequence error and notifies Slack app' do
+        allow(SlackNotifierService).to receive(:new).and_call_original
+
+        expect { described_class.apply }.to raise_error(BaseSynchronizer::FailedUpdatesError)
+
+        expect(SlackNotifierService).to have_received(:new)
```
alexdesi, line 34: It is not `call` a class method?
In other parts, we test it this way:

```
  allow(SlackNotifierService).to receive(:call)
   ...
  expect(SlackNotifierService).to have_received(:call)
```

spec/unit/taric_synchronizer_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+
+      context 'when reindex_all_indexes arg is true' do
+        subject(:apply) { described_class.apply(reindex_all_indexes: true) }
+
+        it 'does not kick off the ClearCacheWorker' do
+          allow(Sidekiq::Client).to receive(:enqueue)
+          allow(TariffSynchronizer::BaseUpdateImporter).to receive(:perform)
+
+          apply
+        rescue StandardError
```
alexdesi, line 211: maybe we can split this into two different tests, and avoid to use  rescue inside the test?

- expect to raise StandardError
- 
- expect(Sidekiq::Client).not_to have_received(:enqueue).with(ClearCacheWorker)

spec/unit/taric_synchronizer_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+
+        allow(TradeTariffBackend).to receive(:uk?).and_return(false)
+      end
+
+      it 'raises FailedUpdatesError and notifies Slack' do
+        allow(SlackNotifierService).to receive(:new).and_call_original
+
+        expect { apply }.to raise_error(BaseSynchronizer::FailedUpdatesError)
+
+        expect(SlackNotifierService).to have_received(:new)
```
alexdesi, line 246: same a before, should we test if it calls the methods `call` (class method) ?

spec/unit/tariff_synchronizer_spec.rb
```diff
@@ -249,13 +249,13 @@
...
       it 'raises a wrong sequence error and notifies Slack app' do
-        allow(SlackNotifierService).to receive(:call)
+        allow(SlackNotifierService).to receive(:new).and_call_original
 
         expect {
           described_class.apply
         }.to raise_error(TariffSynchronizer::FailedUpdatesError)
 
-        expect(SlackNotifierService).to have_received(:call)
+        expect(SlackNotifierService).to have_received(:new)
```
alexdesi, line 258: same here

spec/unit/tariff_synchronizer_spec.rb
```diff
@@ -287,11 +287,11 @@
       let(:pending_date) { applied_date + 2.days }
 
       it 'raises wrong sequence error and notifies Slack app' do
-        allow(SlackNotifierService).to receive(:call)
+        allow(SlackNotifierService).to receive(:new).and_call_original
 
         expect { described_class.apply_cds }.to raise_error(TariffSynchronizer::FailedUpdatesError)
 
-        expect(SlackNotifierService).to have_received(:call)
+        expect(SlackNotifierService).to have_received(:new)
```
alexdesi, line 294: and here

alexdesi (approved):
Nice refactor!
Thanks to your explanation, it all seems to make sense.
As discussed it should be nice to have more instance methods and fewer class methods,
but let's not over-refactor for now.

I've added a couple of comments to improve some tests.

spec/unit/cds_synchronizer_spec.rb
willfish: I think its not anymore for consistency we always instantiate a stateless service. I mainly wanted to know it gets called and we test the call method in isolation elsewhere

spec/unit/taric_synchronizer_spec.rb
willfish: I think we've had this convo before :thinking: . So I think I'll probably completely rewrite the tests at some point so we can tackle that there

alexdesi: ok, yes, we refactored something similar before.

PR #450, opened 2022-03-01
hott-1365 Expose export in the measure serializer.

alexdesi:
### Jira link

[HOTT-1365](https://transformuk.atlassian.net/browse/HOTT-1365)

### What?
Expose export in the measure serializer.

### Why
See the ticket description.


jebw (approved):
Code looks good but interesting brakeman is failing though can't see why this would cause a failure from brakeman

alexdesi:
@jebw after rebasing brakeman passes all the checks :)

PR #452, opened 2022-03-01
HOTT-1360: Surface measure condition class

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1360

### What?

I have added/removed/altered:

- [x] Added measure condition presenter and specs
- [x] Added measure action methods and specs
- [x] Added measure condition class to declarables

### Why?

I am doing this because:

- This is required so we can more easily work with complex measure conditions in the frontend later


jebw (approved):
LGTM

PR #453, opened 2022-03-02
Exit brakeman gracefully after parser errors

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Fixed typo in brakeman config

### Why?

I am doing this because:

- ruby 6 syntax brakes the brakeman sexpression tokeniser and propagates errors since brakeman 4

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -294,7 +294,7 @@ jobs:
       - run:
           name: Inspecting with Brakeman
           when: always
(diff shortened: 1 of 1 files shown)


alexdesi (approved):
👍

jebw:
@willfish FYI - this will have disabled running brakeman? - It will exit with and exit code of 0 and the checks will show as passing even if it finds an issue?

willfish:
That's definitely not what I intended @jebw

PR #455, opened 2022-03-02
Revert "HOTT-1368: Separate EU/UK runs"

willfish:
Reverts trade-tariff/trade-tariff-backend#446


willfish:
Reverting until I've had time to fix a separate bug introduced by curb replacement.

willfish:
I've validated that this ran successfully on staging

PR #456, opened 2022-03-02
Revert "Revert "HOTT-1368: Separate EU/UK runs""

willfish:
Reverts trade-tariff/trade-tariff-backend#455


app/workers/cds_updates_worker.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      CdsSynchronizer.apply(reindex_all_indexes: true)
+    end
+  end
+
+  private
+
+  def cut_off_date_time
+    @cut_off_date_time ||= begin
+      hour, minute = CUT_OFF_TIME.split(':', 2).map(&:to_i)
```
alexdesi, line 29: maybe we can use directly an array as constant:

```
CUT_OFF_TIME = [7, 30]
```

alexdesi (approved):
It looks good to me ✅

PR #457, opened 2022-03-03
HOTT-1359: Simplify and tidy measure presenter

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1359

### What?

I have added/removed/altered:

- [x] Simplify the measure presenter constructor
- [x] Remove unused geographical areas getter, instance variable and signature param

### Why?

I am doing this because:

- I'm working on extending this presenter and it needed cleaning/side effects removing

### DIFF
--- a/app/presenters/api/v2/measures/measure_presenter.rb
+++ b/app/presenters/api/v2/measures/measure_presenter.rb
@@ -2,20 +2,12 @@ module Api
   module V2
     module Measures
       class MeasurePresenter < SimpleDelegator
-        attr_reader :measure, :duty_expression, :geographical_areas, :national_measurement_units
-
         delegate :id, to: :duty_expression, prefix: true
-        delegate :geographical_area, :geographical_area_id, to: :measure
 
-        def initialize(measure, declarable, geographical_areas = [])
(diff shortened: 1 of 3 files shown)


app/presenters/api/v2/measures/measure_presenter.rb
```diff
@@ -26,6 +18,17 @@ def vat
...
         end
 
+        def duty_expression
+          Api::V2::Measures::DutyExpressionPresenter.new(self, @declarable)
+        end
+
+        def national_measurement_units
+          @national_measurement_units ||= @declarable.national_measurement_unit_set
+                                                    &.national_measurement_unit_set_units
+                                                    &.select(&:present?)
```
alexdesi, line 28: I am aware you didn't write it, I think this method has the wrong name,

- if `present?` is the a method from Rails, it is superfluous, for example:

```
 [1,2,3,4,5].select(&:present?)
=> [1, 2, 3, 4, 5]
```

otherwise, it has the wrong name IMO.

alexdesi (approved):
It looks good ✅ 
Just comment about the `present?` method.

PR #458, opened 2022-03-03
HOTT-1359: Surface whether the cds waiver applies for the whole measure

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1359

### What?

I have added/removed/altered:

- [x] Added business logic methods for understanding whether the current measure has the cds universal waiver measure condition
- [x] Added presenter/serializer methods to pull on this new business logic

### Why?

I am doing this because:

- This is going to be used as part of the permuation UI


app/models/measure.rb
```diff
@@ -101,6 +101,10 @@ class Measure < Sequel::Model
 
   delegate :third_country?, :excise?, :vat?, :trade_remedy?, to: :measure_type, allow_nil: true
 
+  def universal_waiver_applies?
+    measure_conditions&.any?(&:universal_waiver_applies?)
```
jebw, line 105: Whats the need for the safe navigation operator? `measure_conditions` is a relationship isn't it? So should not ever be nil?

jebw (approved):
LGTM - only question is around the safe navigation operator - seems like its handling a scenario which cannot happen?

app/models/measure.rb
willfish: No need. Have removed

PR #463, opened 2022-03-08
Handle number indents reloads more wholly

willfish:
### Jira link

HOTT-<TODO>

### What?

This was prompted by https://sentry.io/organizations/engine-le/issues/3079683885/?project=5557005&referrer=slack

I have added/removed/altered:

- [x] Added number_indents override to goods nomenclature that reloads when the goods_nomenclature_indents aren't loaded properly
- [x] Removed reload_if_stale calls and method implementation

### Why?

I am doing this because:

- The reload_if_stale was a patch on a problem that was identified in a specific area of the code base.
- This problem is common to all calls to this relationship so I'm patching it globally

### Not covered

- The relationship that drives this should be a many to one that limits and orders on the dataset - I'm not touching this problem here.
- I'm not understanding/resolving the core issue that leads to this. I've had a look a couple of times and I am unable to reproduce the circumstances that lead to this - it requires a deep dive on our orm infrastructure.

### DIFF
--- a/app/lib/goods_nomenclature_mapper.rb
+++ b/app/lib/goods_nomenclature_mapper.rb
@@ -70,9 +70,6 @@ class GoodsNomenclatureMapper
   end
 
   def map_goods_nomenclatures(primary, secondary)
-    reload_if_stale(primary)
-    reload_if_stale(secondary)
-
     if (heading_map?(primary, secondary) &&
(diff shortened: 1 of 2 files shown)


jebw (approved):
Looks good - I'm trying to consider any unintended consequences of this and non springs to mind. I think its worth a run of the regression suite on staging prior to any release to prod though (or waiting for an overnight run).

PR #466, opened 2022-03-09
HOTT-1361 Add Permutations to Commodities API

jebw:
### Jira link

[HOTT-1361](https://transformuk.atlassian.net/browse/HOTT-1361)

### What?

I have added/removed/altered:

- [x] Moved the MeasureConditionPresenter to instead be a MeasureConditionClassification model, and delegated to it from MeasureCondition
- [x] Added models to represent Permutations, Groups of Permutations and a Calculator to compute the different permutations
- [x] Extended MeasureCondition to return computed permutations
- [x] Added Serializers for Groups and Permutations
- [x] Extended MeasureConditionSerializer to include the Groups and their Permutations
- [x] Feature flagged the API so its not exposed in production or staging till we refine our client

### Why?

I am doing this because:

- We want to expose a new API providing the combintations of Measure Conditions to our frontend and our users

### Have you? (optional)


app/models/measure_condition.rb
```diff
@@ -132,4 +134,16 @@ def universal_waiver_applies?
   def units
     measure_condition_components.map(&:unit)
   end
+
+  def permutation_key
+    return unless certificate_type_code || certificate_code || condition_duty_amount
```
willfish, line 139: So AFAIK the condition either has a certificate type code and certificate code _or_ a condition duty amount.

app/models/measure_condition.rb
```diff
@@ -132,4 +134,16 @@ def universal_waiver_applies?
...
+
+    "#{certificate_type_code}-#{certificate_code}-#{condition_duty_amount}"
+  end
+
+private
+
+  def classification
+    @classification ||= MeasureConditionClassification.new(self)
```
willfish, line 147: Very clean :tada:

app/models/measure_condition_permutations/calculator.rb
```diff
@@ -0,0 +1,110 @@
...
+    # There are 2 rules for calculating permutations
+    #
+    # When there are conditions with matching permutation_keys, then a single
+    # permutation group is provided with all permutations
+    #
+    # If there are no conditions with matching permutation key, then generate
+    # one group per condition code, with a separate permutation per condition
+    # within it
+
+    def initialize(measure)
```
willfish, line 12: So I'm definitely struggling to reason about this class (and I'm not convinced that's because of anything you've done)

spec/models/measure_condition_permutations/calculator_spec.rb
```diff
@@ -0,0 +1,165 @@
+RSpec.describe MeasureConditionPermutations::Calculator do
+  subject(:calculator) { described_class.new measure_with_conditions }
+
+  let(:measure) do
+    create(:measure)
+      .tap { |m| create :measure_condition, measure_sid: m.measure_sid }
+  end
+
+  describe '#measure_conditions' do
```
willfish, line 9: I think we could move this to be a public scope on another model? It seems to be someone else's api

willfish:
I think we need to have some coverage for the scenarios we've got in the ticket:

#### Scenario 1 

**Example**

_V_

- ABC
- DEF
- FOO
- QUX

**Expected Permutations**

- [ABC]
- [DEF]
- [FOO]
- [QUX]

#### Scenario 2

**Example**

_V_

- ABC

_D_

- DEF

**Expected Permutations**

- [ABC, DEF]

#### Scenario 3 

**Example**

_V_

- ABC
- DEF
- QUX

_D_

- ABC
- DEF
- FOO

**Permutations**

- [ABC]
- [DEF]
- [QUX,FOO]

#### Scenario 4 

**Example**

_V_

- ABC
- DEF
- QUX

_D_

- ABC
- DEF
- FOO

_C_

- ABC
- DEF
- FUX

**Permutations**

- [ABC, FUX]
- [DEF]
- [QUX, FOO, FUX]

willfish:
I think having the UI plumbed together might make it easier to validate the api responses. I ended up crafting my own expectations because json api (especially for large responses) can be quite fiddly

app/models/measure_condition.rb
alexdesi: The actual condition means that all `certificate_type_code, certificate_code, condition_duty_amount` must be true to return 
"#{certificate_type_code}-#{certificate_code}-#{condition_duty_amount}"
but @willfish is suggesting that  the condition should be
`if (certificate_type_code && certificate_code)  ||  condition_duty_amount`

I don't know, just thinking out loud about this 😄

app/models/measure_condition_classification.rb
```diff
@@ -0,0 +1,46 @@
+class MeasureConditionClassification
+  DOCUMENT_CLASS = 'document'.freeze
+  EXEMPTION_CLASS = 'exemption'.freeze
+  THRESHOLD_CLASS = 'threshold'.freeze
+  NEGATIVE_CLASS = 'negative'.freeze
+  UNKNOWN_CLASS = 'unknown'.freeze
```
alexdesi, line 6: what about using the magic comment at the beginning:
`# frozen_string_literal: true`

to avoid freezing every single string?

app/models/measure_condition_classification.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  EXCLUDED_DOCUMENT_CODE_CLASS_DOCUMENT_CODES = [
+    '999L', # National Document: CDS universal waiver
+    'C084', # Other certificates: Exemption by virtue of Articles 3 and 4 of regulation 2019/2122 (animals intended for scientific purposes, research and diagnostic samples)
+  ].freeze
+
+  EXEMPTION_CERTIFICATE_TYPE_CODE = 'Y'.freeze # Particular provisions
```
alexdesi, line 13: we should group this with the other const strings above.

app/models/measure_condition_classification.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def measure_condition_class
+    return THRESHOLD_CLASS if threshold_class?
+    return NEGATIVE_CLASS if negative_class?
+    return EXEMPTION_CLASS if exemption_class?
+    return DOCUMENT_CLASS if document_class?
+
+    UNKNOWN_CLASS
+  end
```
alexdesi, line 26: Just for fun: to avoid this conditional cascade we could do this:

```
def measure_condition_class
  threshold_class || negative_class ||
  exemption_class || document_class ||
  UNKNOWN_CLASS
end

def threshold_class
  @measure_condition.condition_duty_amount.present? && THRESHOLD_CLASS
end

def negative_class
  @measure_condition.measure_action.present? && 
  @measure_condition.measure_action.negative_action? &&
  NEGATIVE_CLASS
end

... 

```

also, I guess this should be related to the ask don't tell principle.

Not very opinionated, up to you.

app/models/measure_condition_permutations/calculator.rb
```diff
@@ -0,0 +1,110 @@
...
+      @measure = measure
+    end
+
+    def measure_conditions
+      @measure_conditions = @measure.measure_conditions
```
alexdesi, line 17: did you mean to memoize here? `||=`

alexdesi: also, where is this method used? I saw it only in the tests,
maybe I am missing something

app/models/measure_condition_classification.rb
willfish: I don't think we should start doing that unless we decide to do it more frequently. This would add a mix of patterns in the application which we have way too much of already

app/models/measure_condition.rb
jebw: @willfish yeh - I'd thought that rule should be `(certificate_type_code && certificate_code) || condition_duty_amount)` but this was changed at Matts request so I think its right.

@alexdesi the condition is _any_ of them, not _all_ of them - its just used for the grouping so if it ends up some like "AB--" as a permutation key thats fine - this isn't publicly exposed, its just an implementation detail

app/models/measure_condition_classification.rb
jebw: @willfish any opinions? I know you'd written this originally, I just moved its location?

app/models/measure_condition_permutations/calculator.rb
jebw: Hmm - good question, I think I decided against memoizing it but that means it doesn't need assigning to a variable - code moved around a lot, I think this was moved from the initializer where direct assignment would make sense.

I'll revisit this

app/models/measure_condition_classification.rb
willfish: I mean its conceptually totally different

app/models/measure_condition_classification.rb
willfish: So I don't have strong opinions other than I don't see it adding a lot of value in an already complex PR

app/models/measure_condition_permutations/calculator.rb
willfish: I think this is used in the serializer?

spec/serializers/api/v2/measures/measure_condition_serializer_spec.rb
```diff
@@ -1,7 +1,7 @@
 RSpec.describe Api::V2::Measures::MeasureConditionSerializer do
   subject(:serializer) { described_class.new(serializable).serializable_hash.as_json }
 
-  let(:serializable) { Api::V2::Measures::MeasureConditionPresenter.new(create(:measure_condition)) }
+  let(:serializable) { create(:measure_condition) }
```
willfish, line 4: Interesting :wave: . I guess we don't rely on any of the presenter behaviour :tada:

app/models/measure_condition_permutations/calculator.rb
jebw: Should be simpler now - split into two separate calculators

app/presenters/api/v2/measures/measure_presenter.rb
```diff
@@ -93,6 +94,16 @@ def scheme_code
           TradeTariffBackend.rules_of_origin.scheme_associations[geographical_area_id]
         end
 
+        def measure_condition_permutation_groups
+          @measure_condition_permutation_groups = \
+            MeasureConditionPermutations::Calculator.new(@measure)
```
willfish, line 99:
```suggestion
            MeasureConditionPermutations::Calculator.new(self)
```

app/services/cached_commodity_service.rb
```diff
@@ -140,7 +146,7 @@ def meursing_additional_code_id
   end
 
   def cache_key
-    "_commodity-#{commodity.goods_nomenclature_sid}-#{actual_date}-#{TradeTariffBackend.currency}-#{geographical_area_id}-#{meursing_additional_code_id}"
+    "_commodity-v#{CACHE_VERSION}-#{commodity.goods_nomenclature_sid}-#{actual_date}-#{TradeTariffBackend.currency}-#{geographical_area_id}-#{meursing_additional_code_id}"
```
willfish, line 149: Do we need this chipolata? I think we clear this nightly anyway and this would only cover us if we backed out of permutations?

willfish: At which point we could just clear the cache manually on a deploy

jebw:
Don't need to - github just updates base branch automatically

app/services/cached_commodity_service.rb
jebw: Its pretty useful to be able to just bump the cache version number when at PR invalidates it - thats pretty much how Rails default cache keys work

PR #467, opened 2022-03-09
HOTT-1396 Fix migrations on staging refresh

jebw:
### Jira link

[HOTT-1396](https://transformuk.atlassian.net/browse/HOTT-1396)

### What?

I have added/removed/altered:

- [x] Corrected the service passed into the migrations command

### Why?

I am doing this because:

- The wrong service value was being passed into the migrations command, preventing the weekly staging refresh from running the migrations

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -270,7 +270,7 @@ jobs:
             cf conduit tariff-"<< parameters.service >>"-staging-postgres -- psql $PGDATABASE < tariff-"<< parameters.service >>"-production-postgres.psql
             cf conduit tariff-"<< parameters.service >>"-staging-postgres -- psql $PGDATABASE < db/seeds/staging.sql
       - run_migrations:
-          service: staging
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -270,7 +270,7 @@ jobs:
             cf conduit tariff-"<< parameters.service >>"-staging-postgres -- psql $PGDATABASE < tariff-"<< parameters.service >>"-production-postgres.psql
             cf conduit tariff-"<< parameters.service >>"-staging-postgres -- psql $PGDATABASE < db/seeds/staging.sql
       - run_migrations:
-          service: staging
```
willfish, line 273: We should only be running staging migrations? What am I missing?

willfish (approved):

.circleci/config.yml
jebw: Yeh - that was the mistake I made - bit confusing, service is XI or UK

PR #469, opened 2022-03-10
HOTT-1400 Exclude hidden commodities from suggestions

jebw:
### Jira link

HOTT-1400

### What?

I have added/removed/altered:

- [x] Changed SuggestionsService to exclude directly hidden goods nomenclatures

### Why?

I am doing this because:

- they should not be showing up in search suggestions

### DIFF
--- a/app/services/base_suggestions_service.rb
+++ b/app/services/base_suggestions_service.rb
@@ -10,6 +10,7 @@ class BaseSuggestionsService
           .distinct
           .order(Sequel.desc(:goods_nomenclature_item_id))
           .index_by(&:goods_nomenclature_item_id)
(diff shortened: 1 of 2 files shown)


app/services/base_suggestions_service.rb
```diff
@@ -35,4 +36,8 @@ def handle_commodity_record(_commodity)
   def handle_search_reference_record(_search_reference)
     nil
   end
+
+  def hidden_goods_nomenclature_item_ids
```
alexdesi, line 40: what about matching the table name?
`def hidden_goods_nomenclature_ids`

also, it is shorter :)

alexdesi (approved):

PR #470, opened 2022-03-11
HOTT-1404 Use tagged releases

jebw:
### Jira link

[HOTT-1403](https://transformuk.atlassian.net/browse/HOTT-1403)

### What?

I have added/removed/altered:

- [x] create a release + git tag in GitHub
- [x] tag the release in DockerHub
- [x] notify the slack channel when a release happens
- [x] allow re-releasing older releases 

### Why?

I am doing this because:

- We want to be able to redeploy releases without rebuilding
- We want to build a record of what we have released and when
- We want to notify the team when changes are released, and what was changed

- Makes changes to the deployment process

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,31 +6,13 @@ orbs:
   cloudfoundry: circleci/cloudfoundry@1.0
   slack: circleci/slack@4.3.0
   queue: eddiewebb/queue@1.6.4
+  tariff: trade-tariff/trade-tariff-ci-orb@0 # can also change to @dev:<gitsha> for specific version or @dev:alpha to test dev branches
 
 commands:
-  cf_install:
-    parameters:
-      space:
-        type: string
-    steps:
-      - run:
-          name: "Setup CF CLI"
-          command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -49,7 +31,7 @@ commands:
           name: "Push new app in dark mode"
           command: |
             export DOCKER_IMAGE=tariff-backend
-            export DOCKER_TAG=<<# parameters.dev-release >>dev-<</ parameters.dev-release >>${CIRCLE_SHA1}
+            export DOCKER_TAG="<< parameters.docker_image_tag >>"
```
willfish, line 34: Had this thought when I was doing search query parser. Why not just explicitly pass it in everywhere :tada: . The old syntax is awful too

willfish (approved):

.circleci/config.yml
jebw: Yeh, probably a good idea - I'd rather limit the scope of changes in this PR though so may look to do that in a follow up PR

PR #478, opened 2022-03-16
hotfix docker-compose config

chriscoffee:
For some reason I'd excluded a change which disabled security for
opensearch as opensearch is 'better' than elasticsearch when it comes to
defaults choices. However, this gets in the way when we want to run
something locally quickly without messing about too much.

I've updated the file so that we can run the backend with docker-compose
without too much faffing around.

Signed-off-by: Chris Mills <7100370+chriscoffee@users.noreply.github.com>


alexdesi (approved):
perfect, thanks!

PR #480, opened 2022-03-18
HOTT-1411: Surface polymorphic goods nomenclature for quota search

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1411

### What?

I have added/removed/altered:

- [x] Added polymorphic goods nomenclature to measure serializer for quota definitions
- [x] Added related polymorphic methods to the goods nomenclature class

### Why?

I am doing this because:

- This is required so that we can make commodities, subheadings, headings and chapters that are attached to a measure
  link to the correct thing in the frontend

### DIFF
--- a/spec/controllers/api/v2/quotas_controller_spec.rb
+++ b/spec/controllers/api/v2/quotas_controller_spec.rb
@@ -123,17 +123,17 @@ RSpec.describe Api::V2::QuotasController, type: :controller do
                 goods_nomenclature: {
                   data: {
                     id: String,
-                    type: 'goods_nomenclature',
+                    type: 'commodity',
                   },
                 },
               },
             },
             {
               id: String,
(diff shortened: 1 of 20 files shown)


spec/controllers/api/v2/quotas_controller_spec.rb
```diff
@@ -112,6 +120,20 @@
...
+                  },
+                },
+              },
+            },
+            {
+              id: String,
+              type: 'goods_nomenclature',
+              attributes: {
+                goods_nomenclature_class: 'Commodity',
+                goods_nomenclature_id: String,
```
willfish, line 136: Add producline suffix and reflect change in goods_nomenclature_item_id

alexdesi (approved):
✅ We've reviewed the code via Zoom.

PR #486, opened 2022-03-22
HOTT-1414: Tweak tariff mash to work with polymorphic serializers

willfish:
### Jira link

### What?

The relationship method calls map for a TariffMash for non-static serializers: https://github.com/jsonapi-serializer/jsonapi-serializer/blob/master/lib/fast_jsonapi/relationship.rb#L110-L114. We want Mash to be treated like a non-enumerable object.

I have added/removed/altered:

- [x] Altered Hashie::TariffMash so that it works with polymorphic serializers

### Why?

I am doing this because:

- This is required so we can serialize ES-backed search results polymorphically

### AC

- This does not affect when An Array/Hashie::Mash::Array is encountered by the jsonapi-serializer.


jebw (approved):
LGTM

PR #489, opened 2022-03-22
HOTT-1414: Adds necessary attributes to ES-backed goods nomenclature

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added attributes to serialize ES-backed goods nomenclature with new shared jsonapi serializers
- [x] Added geographical area helper to enable storing geographical areas in ES
- [x] Added spec coverage for existing functionality

### Why?

I am doing this because:

- This is required so that ES-backed search results can be serialized using the new shared serializers
- We were missing coverage

### DIFF
--- a/app/serializers/cache/additional_code_serializer.rb
+++ b/app/serializers/cache/additional_code_serializer.rb
@@ -34,7 +34,10 @@ module Cache
             validity_end_date: measure.validity_end_date,
             goods_nomenclature_item_id: measure.goods_nomenclature_item_id,
             goods_nomenclature_sid: measure.goods_nomenclature_sid,
+            goods_nomenclature_id: measure.goods_nomenclature_sid,
             goods_nomenclature: goods_nomenclature_attributes(measure.goods_nomenclature),
+            geographical_area_id: measure.geographical_area_id,
+            geographical_area: geographical_area_attributes(measure.geographical_area),
           }
         end,
       }
--- a/app/serializers/cache/search_cache_methods.rb
+++ b/app/serializers/cache/search_cache_methods.rb
@@ -1,21 +1,34 @@
 module Cache
   module SearchCacheMethods
-    def has_valid_dates(hash, start_key = :validity_start_date, end_key = :validity_end_date)
-      hash[start_key].to_date <= as_of &&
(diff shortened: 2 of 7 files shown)


app/serializers/cache/search_cache_methods.rb
```diff
@@ -1,21 +1,34 @@
 module Cache
   module SearchCacheMethods
-    def has_valid_dates(hash, start_key = :validity_start_date, end_key = :validity_end_date)
```
willfish, line 3: This start_key/end_key behaviour was a copy/paste from a previous implementation that was used to serialize measures

app/serializers/cache/additional_code_serializer.rb
```diff
@@ -34,7 +34,10 @@ def as_json
             validity_end_date: measure.validity_end_date,
             goods_nomenclature_item_id: measure.goods_nomenclature_item_id,
             goods_nomenclature_sid: measure.goods_nomenclature_sid,
+            goods_nomenclature_id: measure.goods_nomenclature_sid,
```
alexdesi, line 37: Do we need it?
also, I am wondering why you added the`goods_nomenclature_id` here, but not in the other serializers.

spec/serializers/cache/search_cache_methods_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+  attr_reader :as_of
+
+  def initialize(as_of)
+    @as_of = as_of
+  end
+end
+
+RSpec.describe Cache::SearchCacheMethods do
+  let(:serializer) { TestSearchCacheSerializer.new(Date.parse(as_of)) }
+  let(:as_of) { '2021-01-01' }
```
alexdesi, line 13: Possibly we don't need this let.

alexdesi (approved):
it looks good 👍 
Couple of minor comments

spec/serializers/cache/search_cache_methods_spec.rb
willfish: It just sets up the default value

app/serializers/cache/additional_code_serializer.rb
willfish: Good catch

PR #493, opened 2022-03-23
HOTT-1414: Use shared serializers for certificate search results

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Use shared serializers in certificate search results
- [x] Update certificate search controller specs to reflect change
- [x] Removes unused serializers

### Why?

I am doing this because:

- This is required for polymorphic goods nomenclature being returned by certificates


jebw (commented):
Does this introduces an N+1 on the geographical areas - they weren't previously included in the output, they are now but they are but I'm not seeing any changes to eager load them?

jebw (approved):
No N+1 because results come from ES not DB :+1:

PR #494, opened 2022-03-23
HOTT-1414: Cache Commodity#declarable? result

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added caching to the result of the current Commodity being declarable
- [x] Added method/constant to define formally what grouping and non grouping productline suffixes are
- [x] Added specs for new and updated methods

### Why?

I am doing this because:

- This method is called a lot and is very expensive (we have to enumerate children and related indents. This will optimise other places we ask this question.

### DIFF
--- a/app/models/commodity.rb
+++ b/app/models/commodity.rb
@@ -18,7 +18,7 @@ class Commodity < GoodsNomenclature
   one_to_one :heading, dataset: -> {
     actual_or_relevant(Heading)
            .filter('goods_nomenclatures.goods_nomenclature_item_id LIKE ?', heading_id)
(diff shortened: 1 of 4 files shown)


jebw (approved):
LGTM

Does cache indefinitely but I guess you're relying on the nightly job binning the whole cache?

willfish:
So AFAIK this will cache for the default ttl @jebw (which is 24 hours I believe)

jebw:
@willfish Without a TTL I think it will cache indefinitely

jebw:
Just checked, you're right - its configured to default to 24 hours

PR #498, opened 2022-03-25
Handle when a measure has no matching goods nomenclature

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Added handling for footnote measures that have no goods nomenclature

### Why?

I am doing this because:

- This fails to update some footnotes in ES


jebw (approved):
LGTM

PR #499, opened 2022-03-25
HOTT-1432: Adjust quotas query to remove duplicate definitions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1432

### What?

I have added/removed/altered:

- [x] Alter quotas query to return distinct measures by order number, only
- [x] Adds coverage for duplicate measures

### Why?

I am doing this because:

- We use the measure model to find applicable measures and we do not need these to be distinct on the validity start date.
- Its possible/happens in the wild for 2 measures to have a different validity start dates despite having the same order number.
- The order number and definition are one-to-one and we only ever need one active order number measure returned to get the active definition.


jebw (approved):
LGTM

PR #500, opened 2022-03-25
HOTT-1434 Search finds subheadings

jebw:
Previously it would not find the commodity and fall back to returning the heading

### Jira link

[HOTT-1434](https://transformuk.atlassian.net/browse/HOTT-1434)

### What?

I have added/removed/altered:

- [x] Changed the Search exact search to return the subheading model for non-declarable commodities instead of returning the parent Heading

### Why?

I am doing this because:

- This means Search in the frontend takes the user to the subheading page, rather then the Commodities page which then incorrectly fallsback to the history page even though its non declarable

- This changes the behaviour of the Search API - we don't actually document this _but_ it is publicly accessible - I would argue it changes to make it more useful but one to consider prior to merging.

Previously: undeclarable commodity = return ExactMatch for heading
Now: undeclarable commodity = return ExactMatch for subheading

### DIFF
--- a/app/services/search_service/exact_search.rb
+++ b/app/services/search_service/exact_search.rb
@@ -58,7 +58,13 @@ class SearchService
 
       # NOTE: at the moment scope .declarable is not enough to
       # determine if it is really declarable or not
-      commodity.present? && commodity.declarable? ? commodity : nil
+      if commodity.blank?
+        nil
+      elsif commodity.declarable?
+        commodity
+      else
+        Subheading.actual.by_code(query).declarable.non_hidden.first
+      end
     end
 
     def find_chapter(query)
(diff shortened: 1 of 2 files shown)


app/services/search_service/exact_search.rb
```diff
@@ -58,7 +58,13 @@ def find_commodity(query)
 
       # NOTE: at the moment scope .declarable is not enough to
       # determine if it is really declarable or not
-      commodity.present? && commodity.declarable? ? commodity : nil
+      if commodity.blank?
+        nil
```
alexdesi, line 62: this looks like a good case for clause guard:

`return nil unless commodity.present?`

PR #505, opened 2022-03-29
HOTT-1414: Exclude measures without goods nomenclature code search

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added filter for measures without goods nomenclature to cache serializer
- [x] Added filter for measures without goods nomenclature to search service
- [x] Added test coverage for both of the above

### Why?

I am doing this because:

- This is required so we filter out broken measures from the additional code search

### DIFF
--- a/app/serializers/cache/additional_code_serializer.rb
+++ b/app/serializers/cache/additional_code_serializer.rb
@@ -10,12 +10,6 @@ module Cache
     end
 
     def as_json
-      measures = additional_code.measures_dataset
-                                .exclude(goods_nomenclature_item_id: nil)
-                                .to_a
-                                .select do |measure|
-        has_valid_dates(measure)
-      end
       {
         additional_code_sid: additional_code.additional_code_sid,
         code: additional_code.code,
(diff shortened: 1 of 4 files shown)


app/serializers/cache/additional_code_serializer.rb
```diff
@@ -26,21 +20,35 @@ def as_json
...
+          geographical_area_id: measure.geographical_area_id,
+          geographical_area: geographical_area_attributes(measure.geographical_area),
+        }
+      end
+    end
+
+    def measures
+      additional_code
+        .measures_dataset.eager(:goods_nomenclature)
+        .exclude(goods_nomenclature_item_id: nil)
```
willfish, line 49: Excludes measures that have no gnid

app/serializers/cache/additional_code_serializer.rb
```diff
@@ -26,21 +20,35 @@ def as_json
...
+          goods_nomenclature: goods_nomenclature_attributes(measure.goods_nomenclature),
+          geographical_area_id: measure.geographical_area_id,
+          geographical_area: geographical_area_attributes(measure.geographical_area),
+        }
+      end
+    end
+
+    def measures
+      additional_code
+        .measures_dataset.eager(:goods_nomenclature)
```
jebw, line 48: minor but can the eager be on its own line - makes it easier to spot in the future

jebw (approved):
LGTM - minor syntactic comment

PR #506, opened 2022-03-30
HOTT-1414: Handle missing goods nomenclature in footnote search

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Refactored the footnote search controller specs
- [x] Added filtering of missing footnote measure goods nomenclature in footnote search results
- [x] Added filtering of missing footnote goods nomenclature in footnote search results
- [x] Added handling for missing/broken goods nomenclature in footnote polymorphic serializer

### Why?

I am doing this because:

- This is required to avoid sentry failures for people scraping footnotes from the api when goods nomenclature associations are broken (see https://sentry.io/organizations/engine-le/issues/3139630294/?project=5557005&query=is%3Aunresolved)

### DIFF
--- a/app/serializers/cache/footnote_serializer.rb
+++ b/app/serializers/cache/footnote_serializer.rb
@@ -2,6 +2,8 @@ module Cache
   class FootnoteSerializer
     include ::Cache::SearchCacheMethods
 
+    MAX_MEASURE_THRESHOLD = 1000
+
     attr_reader :footnote, :as_of
 
     def initialize(footnote)
@@ -10,11 +12,6 @@ module Cache
     end
 
     def as_json
-      goods_nomenclatures = footnote.goods_nomenclatures.reject do |goods_nomenclature|
-        HiddenGoodsNomenclature.codes.include? goods_nomenclature.goods_nomenclature_item_id
-      end.select do |goods_nomenclature|
-        has_valid_dates(goods_nomenclature)
-      end
       footnote_attributes = {
         code: footnote.code,
         footnote_type_id: footnote.footnote_type_id,
@@ -26,12 +23,10 @@ module Cache
         goods_nomenclature_ids: goods_nomenclatures.map(&:goods_nomenclature_sid),
         goods_nomenclatures: goods_nomenclatures.map do |goods_nomenclature|
           goods_nomenclature_attributes(goods_nomenclature)
-        end
+        end,
       }
 
-      measures = footnote.measures.select { |measure| has_valid_dates(measure) }
-      extra_large_measures = measures.size >= 1000
-      unless extra_large_measures
+      unless extra_large_measures?
         footnote_attributes[:measure_ids] = measures.map(&:measure_sid)
         footnote_attributes[:measures] = measures.map do |measure|
           {
@@ -48,8 +43,34 @@ module Cache
           }
         end
       end
-      footnote_attributes[:extra_large_measures] = extra_large_measures
+      footnote_attributes[:extra_large_measures] = extra_large_measures?
       footnote_attributes
+    end
+
+    private
+
+    def goods_nomenclatures
+      @goods_nomenclatures ||= footnote.goods_nomenclatures.compact.select do |goods_nomenclature|
+        has_valid_dates(goods_nomenclature) &&
+          HiddenGoodsNomenclature.codes.exclude?(goods_nomenclature.goods_nomenclature_item_id)
+      end
+    end
+
+    def measures
+      @measures ||= footnote
+        .measures_dataset
+        .eager(:goods_nomenclature)
+        .exclude(goods_nomenclature_item_id: nil)
+        .all
+        .select do |measure|
+          has_valid_dates(measure) &&
+            measure.goods_nomenclature.present? &&
+            HiddenGoodsNomenclature.codes.exclude?(measure.goods_nomenclature_item_id)
+        end
+    end
+
+    def extra_large_measures?
+      measures.size >= MAX_MEASURE_THRESHOLD
     end
   end
 end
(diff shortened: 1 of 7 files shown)


app/serializers/cache/footnote_serializer.rb
```diff
@@ -26,12 +23,10 @@ def as_json
...
         goods_nomenclatures: goods_nomenclatures.map do |goods_nomenclature|
           goods_nomenclature_attributes(goods_nomenclature)
-        end
+        end,
       }
 
-      measures = footnote.measures.select { |measure| has_valid_dates(measure) }
-      extra_large_measures = measures.size >= 1000
-      unless extra_large_measures
+      unless extra_large_measures?
```
alexdesi, line 29: maybe name this method `large_measures?` (if it does not exist already)

willfish: I did exactly the same thing and figured it was a step too far since the response schema attribute name would have to stay the same but with a different method

app/serializers/cache/footnote_serializer.rb
```diff
@@ -48,8 +43,34 @@ def as_json
...
+          HiddenGoodsNomenclature.codes.exclude?(goods_nomenclature.goods_nomenclature_item_id)
+      end
+    end
+
+    def measures
+      @measures ||= footnote
+        .measures_dataset
+        .eager(:goods_nomenclature)
+        .exclude(goods_nomenclature_item_id: nil)
+        .all
```
alexdesi, line 64: we might want to move the first part of this, on the Footnote model

```
= footnote
        .measures_dataset
        .eager(:goods_nomenclature)
        .exclude(goods_nomenclature_item_id: nil)
        .all
```

app/services/footnote_search_service.rb
```diff
@@ -76,4 +75,12 @@ def apply_type_filter
   def apply_description_filter
     @query.push({ multi_match: { query: description, fields: %w[description], operator: 'and' } })
   end
+
+  def apply_post_fetch_filters
```
alexdesi, line 79: is this used?

spec/controllers/api/v2/footnotes_controller_spec.rb
```diff
@@ -1,108 +1,147 @@
...
+          page: Integer,
+          per_page: Integer,
+          total_count: Integer,
+        },
+      },
+    }
+  end
+
+  before do
+    measure_with_goods_nomenclature = create(:measure, :with_base_regulation, goods_nomenclature: create(:heading, :with_description))
```
alexdesi, line 109: maybe it can help, there is a trait called 
`:with_goods_nomenclature`

willfish: I wanted a heading, though

spec/controllers/api/v2/footnotes_controller_spec.rb
```diff
@@ -1,108 +1,147 @@
...
+    create(:footnote_association_goods_nomenclature, footnote:, goods_nomenclature: create(:heading, :with_description))
+    create(:footnote_association_goods_nomenclature, footnote:, goods_nomenclature: create(:chapter, :with_description))
+    create(:footnote_association_goods_nomenclature, footnote:, goods_nomenclature_sid: '9999') # Broken reference
+
+    create(:footnote_association_measure, measure: measure_with_goods_nomenclature, footnote:)
+    create(:footnote_association_measure, measure: measure_no_goods_nomenclature, footnote:)
+
+    Sidekiq::Testing.inline! do
+      TradeTariffBackend.cache_client.reindex
+      sleep(2)
```
alexdesi, line 121: what's happen if we remove the sleep?
... it is inline, should be fine 🤔

app/serializers/cache/footnote_serializer.rb
willfish: I think probably we could rejig all of the measure datasets to exclude the dead nomenclature and DRY these things up a bit/make the rest of the application safer.

app/serializers/cache/footnote_serializer.rb
```diff
@@ -48,8 +43,34 @@ def as_json
...
+        .select do |measure|
+          has_valid_dates(measure) &&
+            measure.goods_nomenclature.present? &&
+            HiddenGoodsNomenclature.codes.exclude?(measure.goods_nomenclature_item_id)
+        end
+    end
+
+    def extra_large_measures?
```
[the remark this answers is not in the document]
alexdesi: ok, doesn't matter then

spec/controllers/api/v2/footnotes_controller_spec.rb
willfish: We could add a trait :with_goods_nomenclature_heading but it seemed overkill

spec/controllers/api/v2/footnotes_controller_spec.rb
willfish: Is async AFAIK

willfish: I'll try it

willfish: Fails

app/serializers/cache/footnote_serializer.rb
```diff
@@ -48,8 +43,34 @@ def as_json
...
+      measures.size >= 1000
```
alexdesi, line 73: this should use the constant:
measures.size >  `MAX_MEASURE_THRESHOLD`

alexdesi (approved):
✅

spec/controllers/api/v2/footnotes_controller_spec.rb
alexdesi: 👍  thanks for trying 😄

PR #508, opened 2022-03-30
Silence noisy alerts in sentry

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Log missing measurement units to stdout rather than sentry
- [x] Update specs to reflect change

### Why?

I am doing this because:

- This adds very little value and gets alerted alot of old CHIEF units

### DIFF
--- a/app/models/measurement_unit.rb
+++ b/app/models/measurement_unit.rb
@@ -47,9 +47,9 @@ class MeasurementUnit < Sequel::Model
       qualifier_code = unit_key.length == 4 ? unit_key[3..] : ''
 
       if unit.present?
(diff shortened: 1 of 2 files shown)


jebw (commented):
Personally, not convinced about this change - unless we are made aware of a problem its unlikely anyone will be proactively looking for this in the logs, hence why its going to Sentry at present

A different approach would be either
a) Add the missing codes if they should be in there
b) Add an 'ignore list' of CHIEF codes we don't care about and don't need to report to sentry - I don't think there are many chief codes and we can just evolve the list over time?

PR #510, opened 2022-03-31
HOTT-1466: Fixes type on certificate list serializer

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1466

### What?

I have added/removed/altered:

- [x] Fixes type in certificate list serializer
- [x] Fixed broken coverage in certificate type controller spec
- [x] Linted and tidied up certificate controller spec

### Why?

I am doing this because:

- This is currently returning certificates and saying they're a certificate_type

### DIFF
--- a/app/controllers/api/v2/certificates_controller.rb
+++ b/app/controllers/api/v2/certificates_controller.rb
@@ -10,11 +10,6 @@ module Api
       end
 
       def index
-        certificates = Certificate.actual
-                                  .eager(:certificate_descriptions, :certificate_description_periods, :certificate_type_description)
-                                  .order(Sequel.asc(%i[certificate_type_code certificate_code]))
-                                  .all
-
         render json: Api::V2::Certificates::CertificateListSerializer.new(certificates).serializable_hash
       end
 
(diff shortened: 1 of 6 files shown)


app/controllers/api/v2/certificates_controller.rb
```diff
@@ -10,11 +10,6 @@ def search
       end
 
       def index
-        certificates = Certificate.actual
-                                  .eager(:certificate_descriptions, :certificate_description_periods, :certificate_type_description)
-                                  .order(Sequel.asc(%i[certificate_type_code certificate_code]))
-                                  .all
-
         render json: Api::V2::Certificates::CertificateListSerializer.new(certificates).serializable_hash
```
willfish, line 13: We could safely share this serializer with the search action as long as the certificate type description was being added in the cache serializer

jebw (approved):
LGTM

Only query is around coordination with frontend - presumably that will need to change in parallel because of the change of jsonapi type?

willfish:
AFAIK this isn't an api we drive in the frontend so should be safe

willfish:
I think the type hasn't affected the model in the frontend ever/is just ignored:

https://github.com/trade-tariff/trade-tariff-frontend/blob/main/app/models/certificate.rb#L3-L20

PR #511, opened 2022-03-31
Add filter for include in QuotasController.

alexdesi:
### What?
Add filter for the include param in QuotasController.
The only allowed param is `quota_balance_events`. If not set, it uses the default includes:
  quota_order_number quota_order_number.geographical_areas measures measures.goods_nomenclature measures.geographical_area

### Jira link
HOTT-1419


spec/controllers/api/v2/quotas_controller_spec.rb
```diff
@@ -164,7 +155,7 @@
           year: [
             Time.zone.today.year.to_s,
           ],
-          include: 'quota_balance_events,measures,measures.geographical_area',
+          include: 'quota_balance_events',
```
willfish, line 158: We're missing an example covering the additional includes case?

app/controllers/api/v2/quotas_controller.rb
```diff
@@ -36,25 +38,31 @@ def search_service
...
-      end
+      def valid_includes
+        return DEFAULT_INCLUDES if include_params.empty?
+
+        valid_resources = include_params.select { |resource| ALLOWED_INCLUDES.include?(resource) }
 
-      def includes
-        return provided_includes if provided_includes.present?
+        if valid_resources.length < include_params.length
+          Sentry.capture_message("Error: invalid params in 'includes': #{invalid_includes}")
```
willfish, line 47: So we're making noise for ourselves but not returning an error to the end user. Probably its easier to just return an error that specifies the user needs to use only quota_balance_events in the includes list.

alexdesi: maybe we should raise an ArgumentError, I saw we do it in a few places.

spec/controllers/api/v2/quotas_controller_spec.rb
alexdesi: the other case:
`'when not specifying an includes list in the query params'`
is already in place (and working).

spec/controllers/api/v2/quotas_controller_spec.rb
```diff
@@ -269,7 +221,7 @@
         }
       end
 
-      it 'returns rendered found quotas' do
+      it 'includes only "quota_balance_events"' do
```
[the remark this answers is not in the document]
alexdesi: changed to 'returns rendered found quotas with the allowed resources'

app/controllers/api/v2/quotas_controller.rb
```diff
@@ -36,25 +38,31 @@ def search_service
...
-      end
+      def valid_includes
+        return DEFAULT_INCLUDES if include_params.empty?
+
+        valid_resources = include_params.select { |resource| ALLOWED_INCLUDES.include?(resource) }
 
-      def includes
-        return provided_includes if provided_includes.present?
+        if valid_resources.length < include_params.length
+          raise ArgumentError, "Error: invalid params in 'includes': #{invalid_includes}"
```
willfish, line 47: Thanks!

willfish (approved):

PR #524, opened 2022-04-11
HOTT-1487: Adds plumbing for csv serializers

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1487

### What?

I have added/removed/altered:

- [x] Added /sections.csv csv endpoint and specs
- [x] Added csv serializer concern
- [x] Migrated goods nomenclature csv endpoint to using new plumbing

### Why?

I am doing this because:

- This is required as part of a wider rollout of csv endpoints


app/serializers/api/v2/csv/csv_serializer.rb
```diff
@@ -0,0 +1,56 @@
...
+              serializable.public_send(column)
+            end
+
+            acc << data_row
+          end
+        end
+
+        def serialized_csv
+          serialized = CSV.generate do |csv|
+            serializable_array.each do |row|
```
jebw, line 24: This will build up a duplicate row array (of column arrays) of all data to be exported?

```ruby
def serialized_csv
  CSV.generate do |csv|
    add_header_row(csv)

    @serializables.each do |serializable|
      add_data_row(csv, serializable)
    end
  end
end

def add_data_row(csv, serializable)
  csv << columns.map do |column|
    serializable.public_send(column)
  end
end
```

app/serializers/api/v2/csv/csv_serializer.rb
```diff
@@ -0,0 +1,56 @@
...
+              csv << row
+            end
+          end
+
+          serialized.html_safe
```
jebw, line 29: A CSV shouldn't need to be marked as `.html_safe` since its not related to HTML? This would only be necessary if its being processed by ERB ?

I think `render content: ....` will treat the supplied string as an inline erb template, which may be why this was required - if so `render plain: ` should avoid the use of ERB but using `send_data` would be simpler still.

spec/controllers/api/v2/sections_controller_index_spec.rb
```diff
@@ -0,0 +1,50 @@
+RSpec.describe Api::V2::SectionsController do
```
jebw, line 1: This veers into my personal opinion so take or ignore as appropriate.

Controller specs are deprecated and have been for a while, their suggested replacement is request specs. Could this be a request spec instead?

It is also delving into the structure of the CSV _but_ I'd think the structure of the serialized csv gets covered by the serializer spec, this would just be an additional place to update for any future serialisation changes?

To my mind a request spec cares that 'given expected params does the controller return a csv' (eg check response status, check mime type header?)

jebw (commented):
Looks like a really solid base for all the CSV endpoints

There's a few refinements suggested below - some are just my opinion but some may help with keeping things simple.

app/serializers/api/v2/csv/csv_serializer.rb
```diff
@@ -0,0 +1,56 @@
+module Api
```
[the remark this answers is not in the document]
willfish: Ah good shout, yeah.

app/serializers/api/v2/csv/section_serializer.rb
```diff
@@ -0,0 +1,16 @@
+module Api
+  module V2
+    module Csv
+      class SectionSerializer
+        include CsvSerializer
+
+        columns :id,
```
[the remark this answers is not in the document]
willfish: That's coming next, yeah. I was actually hoping to stop us using different names for things

app/serializers/api/v2/csv/csv_serializer.rb
```diff
@@ -0,0 +1,56 @@
...
+          serialized = CSV.generate do |csv|
+            serializable_array.each do |row|
+              csv << row
+            end
+          end
+
+          serialized.html_safe
+        end
+
+        def initial_rows
```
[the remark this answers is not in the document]
willfish: Well, its the header row and the containing accumulator

app/serializers/api/v2/csv/csv_serializer.rb
willfish: Ah, I suspect that's a hangover, yeah

app/controllers/api/v2/sections_controller.rb
```diff
@@ -1,10 +1,19 @@
...
+      before_action :set_sections, only: :index
+
       def index
-        @sections = Section.eager({ chapters: [:chapter_note] }, :section_note).all
+        respond_to do |format|
+          format.csv do
+            headers['Content-Type'] = 'text/csv'
+            headers['Content-Disposition'] = "attachment; filename=#{filename}.csv"
+
+            render content: Api::V2::Csv::SectionSerializer.new(@sections).serialized_csv
```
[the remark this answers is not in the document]
willfish: Could do. I'll check if the controller spec still passes.

app/serializers/api/v2/csv/csv_serializer.rb
```diff
@@ -0,0 +1,56 @@
...
+          [columns]
+        end
+
+        delegate :columns, to: :class
+
+        module ClassMethods
+          def columns(*column_list)
+            if column_list.compact.present?
+              column_list.each do |column|
+                (@columns ||= []) << column
```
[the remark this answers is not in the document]
willfish: This doesn't actually do what I want it to do re the DSL

spec/controllers/api/v2/sections_controller_index_spec.rb
willfish: Well, I have the same opinion - mainly because that's the recommendation of the rails community, but also because controller specs hide a lot of the request/response cycle from the tester. We have a mixture in the code base and I haven't made it an overt goal to actually convert all of our controller specs to request specs. It is a bit of a smell wherever we use `render_views`, too.

app/serializers/api/v2/csv/csv_serializer.rb
willfish: I'm not seeing any duplicates?

willfish: What I've done achieves this already, tbh

app/serializers/concerns/api/v2/csv/csv_serializer.rb
```diff
@@ -0,0 +1,58 @@
...
+        private
+
+        def header_row
+          column_options.map do |column_option|
+            column_option[:column_name]
+          end
+        end
+
+        module ClassMethods
+          cattr_accessor :column_options
```
jebw, line 45: I think this will mean column options end up shared across all classes including the module - I think you need class level instance accessors instead

```ruby
module CsvSerializer
  extend ActiveSupport::Concern
  
  included do
    class << self
      attr_accessor :column_options
    end
  end
end
```

app/serializers/concerns/api/v2/csv/csv_serializer.rb
```diff
@@ -0,0 +1,58 @@
...
+            end
+          end
+        end
+
+        delegate :column_options, to: :class
+
+        private
+
+        def header_row
+          column_options.map do |column_option|
```
jebw, line 39: `column_options.pluck(:column_name)` ?

app/serializers/concerns/api/v2/csv/csv_serializer.rb
```diff
@@ -0,0 +1,58 @@
...
+        module ClassMethods
+          cattr_accessor :column_options
+
+          def columns(*column_options)
+            column_options.each(&method(:column))
+          end
+
+          def column(column, options = {}, &block)
+            (self.column_options ||= []) << options.reverse_merge(column:, column_name: column, value_block: block)
+          end
```
jebw, line 53: Splitting the initialisation would avoid the `cattr_accessor` tricks (which I think will cause issues because of the confusing way `cattr_accessor` works)

```ruby
module ClassMethods
  def column(column, **options, &block)
    self.column_options << options.reverse_merge(column:, column_name: column, value_block: block)
  end

  def column_options
    @column_options ||= []
  end
end
```

app/serializers/concerns/api/v2/csv/csv_serializer.rb
jebw: Ignore this, better option in comment below

jebw (commented):
Nice use of blocks for the column definitions!

The bit I think may cause issues is the `cattr_accessor` and testing locally I think this will end up on the module (not the class including the module) - so will end up common across all classes sharing the module.

I suspect the tests still pass because each spec is pulling in the serializer definition, then testing it. And when its pulled in it overwrites the earlier definition, but that then isn't subsequently touched by the test suite.

app/serializers/concerns/api/v2/csv/csv_serializer.rb
willfish: Great idea :). Also, good spot

app/serializers/concerns/api/v2/csv/csv_serializer.rb
```diff
@@ -0,0 +1,58 @@
...
+    module Csv
+      module CsvSerializer
+        extend ActiveSupport::Concern
+
+        def initialize(serializables, options = {})
+          @serializables = serializables
+          @options = options
+        end
+
+        def serializable_array
```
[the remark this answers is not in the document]
willfish: So this is the approach jsonapi take so I'm just following that. You can get an intermediate/modifiable form and a serialized form (think serializable_hash)

alexdesi (approved):
It looks all right here ✅ 
Just a fussy comment (not blocking).

spec/serializers/api/v2/csv/section_serializer_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      )
+
+      [section]
+    end
+
+    it 'serializes correctly' do
+      expect(serializable_array).to eq(
+        [
+          %i[id numeral title position chapter_from chapter_to],
+          [18, 'XVIII', 'Optical, photographic, cinematographic, measuring', 18, nil, nil],
```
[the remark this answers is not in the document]
willfish: Done :) and agree.

PR #535, opened 2022-04-14
hott-1489 Add the new action chapter on v2/sections_controller

alexdesi:
### What?
Add the new action chapter on v2/sections_controller and support for CSV and JSON responses.

### Jira link

https://transformuk.atlassian.net/browse/HOTT-1489


spec/controllers/api/v2/sections_controller_spec.rb
```diff
@@ -1,128 +1,141 @@
...
+  # '/api/v2/sections/:id/chapters'
+  describe '#chapters' do
+    let(:section) { create(:chapter, :with_section).section }
 
-    expect(response.body).to match_json_expression pattern
+    it 'returns successfully' do
+      get :chapters, params: { id: section.id }, format: 'csv'
+
+      expect(response).to be_successful
+    end
```
alexdesi, line 139: I refactored this spec a bit,
this test here, is the only I added, to check that it responds to '/api/v2/sections/:id/chapters' .

app/controllers/api/v2/sections_controller.rb
```diff
@@ -28,6 +28,26 @@ def show
...
+        chapters = Section.where(id: params[:id])
+                          .take
+                          .chapters
+
+        respond_to do |format|
+          format.csv do
+            send_data(
+              Api::V2::Csv::ChapterSerializer.new(chapters).serialized_csv,
+              type: 'text/csv; charset=utf-8; header=present',
+              disposition: "attachment; filename=chapters-#{actual_date.iso8601}.csv",
```
willfish, line 42:
```suggestion
              disposition: "attachment; filename=section-#{params[:id]}-chapters-#{actual_date.iso8601}.csv",
```

app/controllers/api/v2/sections_controller.rb
```diff
@@ -28,6 +28,26 @@ def show
         options[:include] = [:chapters, 'chapters.guides']
         render json: Api::V2::Sections::SectionSerializer.new(@section, options).serializable_hash
       end
+
+      def chapters
```
willfish, line 32: Would it be better to make this the index of a chapters controller?

spec/controllers/api/v2/sections_controller_spec.rb
willfish: You should be fine to just use the shared examples :)

app/controllers/api/v2/sections_controller.rb
alexdesi: the ChaptersController#index is already used 🤔 
maybe passing the section id too, 
we should use 2 different routes for the same controller's action,
I guess it would complicate the action and tests quite a bit.

spec/controllers/api/v2/sections_controller_spec.rb
```diff
@@ -1,128 +1,145 @@
...
+      expect(response.body).to match_json_expression pattern
+    end
   end
 
-  it 'returns rendered records' do
-    get :index, format: :json
+  # '/api/v2/sections/:id/chapters'
+  describe '#chapters' do
+    subject(:do_request) do
+      get(:chapters, params: { id: section.id }, format: 'csv')
```
alexdesi, line 134: I  tried, but I could not reuse your test "a successful csv response", 
because I could not use a path with a resource that was not yet created: 
'/api/v2/sections/<ID>/chapters'
The internal `before` block is executed after the path was actually used.

Anyway, I preferred to keep the test simple with a bit of duplication instead of doing strange things using lazy evaluation.

app/controllers/api/v2/sections_controller.rb
willfish: So I think we need Sections::ChaptersController#index, basically, to be resourceful and to match the route /sections/1/chapters

spec/controllers/api/v2/sections_controller_spec.rb
willfish: If you look at the other examples, we pass a before callback inside of a block we pass to the shared example to achieve this

willfish: https://github.com/trade-tariff/trade-tariff-backend/blob/main/spec/requests/api/v2/chapters_controller_spec.rb#L3-L8

willfish: describe 'GET #index' do
    let(:section) { create(:section, :with_chapters) } }

    it_behaves_like 'a successful csv response', "/sections/{section.id}/chapters"
  end

willfish: Also, we probably want a request spec for consistency and current recommended practice: https://github.com/trade-tariff/trade-tariff-backend/pull/524#discussion_r848429517

alexdesi: thanks @willfish I'll try it now 👍

alexdesi:
I've made the refactor using a different controller,  which is great.
All right, the only thing is that the stared example does not work, so I had to write a small custom test.
Let's catch up tomorrow.

spec/controllers/api/v2/sections_controller_spec.rb
alexdesi: naah does not work 😄 ,
let's catch up tomorrow.

app/controllers/api/v2/sections/chapters_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+                disposition: "attachment; filename=sections-#{params[:id]}-chapters.csv",
+              )
+            end
+
+            format.any do
+              render json: Api::V2::Chapters::ChapterListSerializer.new(chapters).serializable_hash
+            end
+          end
+        end
+
```
willfish, line 24: Surprised the linter is passing with this extra space :smile_cat:

spec/controllers/api/v2/sections_controller_spec.rb
```diff
@@ -1,128 +1,130 @@
-RSpec.describe Api::V2::SectionsController, 'GET #show' do
-  let(:heading) { create :heading, :with_chapter }
-  let(:chapter) { heading.reload.chapter }
-  let(:chapter_guide) { chapter.guides.first }
-  let(:section) { chapter.section }
-  let!(:section_note) { create :section_note, section_id: section.id }
+RSpec.describe Api::V2::SectionsController do
+  # GET /api/v2/sections/:id
```
willfish, line 2: Love this refactor :tada:

spec/services/api/v2/sections/chapters_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      get(:index, params:, format: 'csv')
+
+      response
+    end
+
+    it { is_expected.to be_successful }
+
+    it { is_expected.to have_attributes media_type: /csv/ }
+
+    context "when the param 'as_of' is specified" do
```
willfish, line 18: It actually does this behaviour without the as_of

spec/services/api/v2/sections/chapters_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
+RSpec.describe Api::V2::Sections::ChaptersController do
```
willfish, line 1: I think you mean for this file to be under spec/requests/api/v2/sections/chapters_controller_spec.rb?

spec/services/api/v2/sections/chapters_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  # '/api/v2/sections/:id/chapters'
+  describe '#index', type: :controller do
```
willfish, line 4: AFAIK we should be able to just reuse the shared examples?

spec/services/api/v2/sections/chapters_controller_spec.rb
willfish: Also, I think we can move this to the shared examples, actually. Since this is actually part of the AC of all of the tickets :tada:

willfish (approved):

PR #538, opened 2022-04-19
HOTT-1518: Support text/csv in ACCEPT header

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1518

### What?

I have added/removed/altered:

- [x] Removed default configuration for api/v2 mimetype at the routing level
- [x] Added coverage in csv request specs for ACCEPT header

### Why?

I am doing this because:

- This is required so we can support the text/csv mimetype in the ACCEPT header

- We should run a full regression and leave this on staging for a bit

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -43,7 +43,7 @@ Rails.application.routes.draw do
     end
   end
 
-  namespace :api, defaults: { format: 'json' }, path: '/' do
+  namespace :api, path: '/' do
     # How (or even if) API versioning will be implemented is still an open question. We can defer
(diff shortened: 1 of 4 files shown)


alexdesi:
I thought we were going to generalise the test

`it_behaves_like 'a successful csv response'`

but I guess it's not doable, since we should make two different calls 🤔

willfish:
Yeah, I had figured it muddied the shared example too much. I can just interpolate the path as a shared example block param (see code review commit)

alexdesi (approved):
nice one, I think it's good to use two contexts inside the shared example 👍

PR #540, opened 2022-04-19
HOTT-1519 Always use latest v1 Ruby orb

jebw:
### Jira link

[HOTT-1519](https://transformuk.atlassian.net/browse/HOTT-1519)

### What?

I have added/removed/altered:

- [x] Change the CI config to always use the latest v1 version of the Ruby orb
- [x] Fixed rules of origin specs that were failing by a change to the uk schemes file

### Why?

I am doing this because:

- Orbs follow semantic versioning, so it should be safe to upgrade minor and patch releases automatically since they wont include
breaking changes.
- There are known problems with specs not being run for the old v1.4.0 version of the ruby orb we currently use.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   aws-cli: circleci/aws-cli@2.0.3
-  ruby: circleci/ruby@1.4.0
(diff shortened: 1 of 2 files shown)


alexdesi (approved):
it looks good ✅

PR #541, opened 2022-04-20
Revert "HOTT-1518: Support text/csv in ACCEPT header"

willfish:
Reverts trade-tariff/trade-tariff-backend#538


willfish:
We're just going with the mimetype suffix for now

PR #542, opened 2022-04-20
HOTT-1379 Search improvements

jebw:
### Jira link

[HOTT-1379](https://transformuk.atlassian.net/browse/HOTT-1379)

### What?

I have added/removed/altered:

- [x] Check for subheadings as well as commodities when doing an exact search
- [x] Allow matching 6 and 8 digit comm codes when searching, which will take the user to either the relevant commodity or subheading, or fall back to the heading if the comm code is not available

### Why?

I am doing this because:

- I am trying to ensure the search behaves according to the 'scenarios' document attached to the ticket

### Notes

This supercedes #500

### DIFF
--- a/app/services/search_service/exact_search.rb
+++ b/app/services/search_service/exact_search.rb
@@ -4,9 +4,9 @@ class SearchService
       @results = case query_string
                  when /^[0-9]{1,3}$/
                    find_chapter(query_string)
-                 when /^[0-9]{4,9}$/
+                 when /^[0-9]{4}$/
                    find_heading(query_string)
-                 when /^[0-9]{10}$/
+                 when /^[0-9]{5,10}$/
                    # A commodity or declarable heading may have code of
                    # 10 digits
                    find_commodity(query_string) || find_heading(query_string)
@@ -48,6 +48,7 @@ class SearchService
     end
 
     def find_commodity(query)
+      query = query + ('0' * (10 - query.length)) if query.length < 10 # normalise shortened codes
       query = SearchService::CodesMapping.check(query) || query
 
       commodity = Commodity.actual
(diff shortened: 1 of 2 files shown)


app/services/search_service/exact_search.rb
```diff
@@ -48,6 +48,7 @@ def find_heading(query)
     end
 
     def find_commodity(query)
+      query = query + ('0' * (10 - query.length)) if query.length < 10 # normalise shortened codes
```
willfish, line 51: Clean :)

app/services/search_service/exact_search.rb
```diff
@@ -58,7 +59,13 @@ def find_commodity(query)
 
       # NOTE: at the moment scope .declarable is not enough to
       # determine if it is really declarable or not
-      commodity.present? && commodity.declarable? ? commodity : nil
+      if commodity.blank?
+        nil
+      elsif commodity.declarable?
+        commodity
+      else
+        Subheading.actual.by_code(query).declarable.non_hidden.first
```
willfish, line 67: Do we want take here?

```suggestion
        Subheading.actual.by_code(query).declarable.non_hidden.take
```

willfish: Not sure how search plays with 404s

willfish: And clearly nil is being used as not found

willfish: So probably this is actually more consistent with this particular api

willfish (approved):
LGTM

app/services/search_service/exact_search.rb
jebw: No, we don't want to raise at this point - this needs to return nil - the caller uses `find_commodity || find_heading` and has its own handling for if there's no result

PR #544, opened 2022-04-22
HOTT-1490: Add new endpoint /chapters/:id/headings on V2

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1490

### What?
This PR adds a new endpoint 
`/chapters/:id/headings on V2`

and supports the JSON and CSV formats.

It is using the existing serializer for Json: `Api::V2::Headings::HeadingSerializer`
and a new serializer for CSV: `Csv::HeadingSerializer`

### DIFF
--- a/app/controllers/api/v2/chapters_controller.rb
+++ b/app/controllers/api/v2/chapters_controller.rb
@@ -49,21 +49,21 @@ module Api
         chapter_headings = chapter.headings
 
         respond_to do |format|
+          filename = "#{TradeTariffBackend.service}-chapter-#{params[:id]}-headings-#{actual_date.iso8601}.csv"
+
           format.csv do
             send_data(
               Api::V2::Csv::HeadingSerializer.new(chapter_headings).serialized_csv,
               type: 'text/csv; charset=utf-8; header=present',
-              disposition: "attachment; filename=chapter-#{params[:id]}-headings-#{actual_date.iso8601}.csv",
+              disposition: "attachment; filename=#{filename}",
             )
-          end
-
-          format.any do
-            render json: Api::V2::Headings::HeadingSerializer.new(chapter_headings).serializable_hash
           end
         end
       end
 
       private
+
+      attr_reader :chapter
 
       def find_chapter
         @chapter = Chapter.actual
@@ -73,16 +73,8 @@ module Api
         raise Sequel::RecordNotFound if @chapter.goods_nomenclature_item_id.in? HiddenGoodsNomenclature.codes
       end
 
-      attr_reader :chapter
-
       def chapter_id
-        c_id = params[:id]
-
-        if c_id.length == 2
-          "#{c_id}00000000"
-        else # it is single digit
-          "0#{c_id}00000000"
-        end
+        "#{params[:id]}00000000"
       end
     end
   end
--- a/app/controllers/api/v2/chapters_controller.rb
+++ b/app/controllers/api/v2/chapters_controller.rb
@@ -3,7 +3,7 @@ require 'goods_nomenclature_mapper'
 module Api
   module V2
     class ChaptersController < ApiController
-      before_action :find_chapter, only: %i[show changes]
+      before_action :find_chapter, only: %i[show changes headings]
 
       def index
         @chapters = Chapter.eager(:chapter_note, :goods_nomenclature_descriptions).all
@@ -21,14 +21,14 @@ module Api
       end
 
       def show
-        root_headings = GoodsNomenclatureMapper.new(@chapter.headings_dataset
+        root_headings = GoodsNomenclatureMapper.new(chapter.headings_dataset
                                                       .eager(:goods_nomenclature_descriptions,
                                                              :goods_nomenclature_indents)
                                                       .all).root_entries
 
         options = { is_collection: false }
         options[:include] = [:section, :guides, :headings, 'headings.children']
-        presenter = Api::V2::Chapters::ChapterPresenter.new(@chapter, root_headings)
+        presenter = Api::V2::Chapters::ChapterPresenter.new(chapter, root_headings)
         render json: Api::V2::Chapters::ChapterSerializer.new(presenter, options).serializable_hash
       end
 
@@ -45,6 +45,24 @@ module Api
         render json: Api::V2::Changes::ChangeSerializer.new(@changes.changes, options).serializable_hash
       end
 
+      def headings
+        chapter_headings = chapter.headings
+
+        respond_to do |format|
+          format.csv do
+            send_data(
+              Api::V2::Csv::HeadingSerializer.new(chapter_headings).serialized_csv,
+              type: 'text/csv; charset=utf-8; header=present',
+              disposition: "attachment; filename=chapter-#{params[:id]}-headings-#{actual_date.iso8601}.csv",
+            )
+          end
+
+          format.any do
+            render json: Api::V2::Headings::HeadingSerializer.new(chapter_headings).serializable_hash
+          end
+        end
+      end
+
       private
 
       def find_chapter
@@ -55,8 +73,16 @@ module Api
         raise Sequel::RecordNotFound if @chapter.goods_nomenclature_item_id.in? HiddenGoodsNomenclature.codes
       end
 
+      attr_reader :chapter
+
       def chapter_id
-        "#{params[:id]}00000000"
+        c_id = params[:id]
+
+        if c_id.length == 2
+          "#{c_id}00000000"
+        else # it is single digit
+          "0#{c_id}00000000"
+        end
       end
     end
   end
--- a/app/models/heading.rb
+++ b/app/models/heading.rb
@@ -46,8 +46,8 @@ class Heading < GoodsNomenclature
   delegate :section, :section_id, to: :chapter, allow_nil: true
 
   # See oplog sequel plugin
-  def operation=(op)
-    self[:operation] = op.to_s.first.upcase
+  def operation=(operation)
+    self[:operation] = operation.to_s.first.upcase
   end
 
   def short_code
@@ -110,14 +110,6 @@ class Heading < GoodsNomenclature
     ).where(conditions)
      .limit(TradeTariffBackend.change_count)
      .order(Sequel.desc(:operation_date, nulls: :last))
-  end
-
-  def footnote_ids
-    footnotes.map(&:footnote_id)
-  end
-
-  def commodity_ids
-    commodities.map(&:goods_nomenclature_item_id)
   end
 
   private
--- a/app/models/heading.rb
+++ b/app/models/heading.rb
@@ -10,13 +10,13 @@ class Heading < GoodsNomenclature
 
   set_primary_key [:goods_nomenclature_sid]
 
-  one_to_many :commodities, dataset: -> {
+  one_to_many :commodities, dataset: lambda {
     actual_or_relevant(Commodity)
              .filter('goods_nomenclatures.goods_nomenclature_item_id LIKE ?', heading_id)
              .where(Sequel.~(goods_nomenclatures__goods_nomenclature_item_id: HiddenGoodsNomenclature.codes))
   }
 
-  one_to_one :chapter, dataset: -> {
+  one_to_one :chapter, dataset: lambda {
     actual_or_relevant(Chapter).filter('goods_nomenclatures.goods_nomenclature_item_id LIKE ?', chapter_id)
   }
 
@@ -64,7 +64,7 @@ class Heading < GoodsNomenclature
   end
 
   def uptree
-    [self, self.chapter].compact
+    [self, chapter].compact
   end
 
   def non_grouping?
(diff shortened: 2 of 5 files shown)


app/controllers/api/v2/chapters_controller.rb
```diff
@@ -45,6 +45,24 @@ def changes
...
+          format.csv do
+            send_data(
+              Api::V2::Csv::HeadingSerializer.new(chapter_headings).serialized_csv,
+              type: 'text/csv; charset=utf-8; header=present',
+              disposition: "attachment; filename=chapter-#{params[:id]}-headings-#{actual_date.iso8601}.csv",
+            )
+          end
+
+          format.any do
+            render json: Api::V2::Headings::HeadingSerializer.new(chapter_headings).serializable_hash
```
jebw, line 61: Have you done any performance checks for this - my reading of this is

1. Load every heading under a Chapter
2. Serialize each heading
3. heading serializer loads _every_ commodity and compute `children?` for each one

In effect, heading serializer loads _every_ commodity under _every_heading for the chapter, and computes the `commodity.children?` for each one

This feels like like it will potentially take a long time to complete for some chapters - have you tried querying each of the chapters locally to see how long it takes to respond?

The alternative would be a cut down headings serializer for the chapters, which doesn't cascade into the commodities but not sure what was requested originally - there's nothing about this JSON endpoint in the ticket?

jebw (commented):
Code looks good but not sure about the performance implications of the JSON endpoint - feels like it could be very slow, very memory intensive for some Chapters?

app/models/heading.rb
```diff
@@ -10,13 +10,13 @@ class Heading < GoodsNomenclature
 
   set_primary_key [:goods_nomenclature_sid]
 
-  one_to_many :commodities, dataset: -> {
+  one_to_many :commodities, dataset: lambda {
```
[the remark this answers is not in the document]
alexdesi: yeh, that's to make Rubocop happy, not sure if I like this one, I prefer the lambda to be honest.

app/controllers/api/v2/chapters_controller.rb
```diff
@@ -55,8 +73,16 @@ def find_chapter
         raise Sequel::RecordNotFound if @chapter.goods_nomenclature_item_id.in? HiddenGoodsNomenclature.codes
       end
 
+      attr_reader :chapter
+
       def chapter_id
-        "#{params[:id]}00000000"
+        c_id = params[:id]
+
+        if c_id.length == 2
```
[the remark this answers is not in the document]
alexdesi: cool, I'll try it, thanks.

app/models/heading.rb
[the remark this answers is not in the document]
alexdesi: oh yeah, you got it, it's beacause it's multiline

app/serializers/api/v2/csv/heading_serializer.rb
```diff
@@ -0,0 +1,24 @@
...
+
+        columns :goods_nomenclature_item_id,
+                :goods_nomenclature_sid,
+                :declarable,
+                :description,
+                :description_plain,
+                :formatted_description,
+                :producline_suffix
+
+        column :leaf do |heading|
```
willfish, line 15: This should just be:

```ruby
heading.commodities.any?
```

config/routes.rb
```diff
@@ -58,10 +58,14 @@
         end
       end
 
-      resources :chapters, only: %i[index show], constraints: { id: /\d{2}/ } do
+      resources :chapters, only: %i[index show], constraints: { id: /\d{1,2}/ } do
```
willfish, line 61: I'm not sure I understand this? Chapter short codes are always 2 chars

app/controllers/api/v2/chapters_controller.rb
willfish: This is wrong. For a heading to be a leaf we just need to work out whether there are any commodities not do stuff with every commodity

willfish: Basically heading.commodities_dataset.count.positive?

config/routes.rb
```diff
@@ -58,10 +58,14 @@
...
         member do
           get :changes
         end
+
+        member do
+          get :headings
```
willfish, line 67: We could stick this in the member block above

app/controllers/api/v2/chapters_controller.rb
alexdesi: No, nothing in the ticket, that Json was not something required, so we can simplify the json, removing the commodities for example.
Not sure how to do it using the same serializer,  maybe using the option `:include`.

spec/requests/api/v2/chapters_controller_spec.rb
```diff
@@ -8,4 +8,23 @@
       end
     end
   end
+
+  describe 'GET #chapters/:id/headings' do
```
willfish, line 12:
```suggestion
  describe 'GET #headings' do
```

spec/requests/api/v2/chapters_controller_spec.rb
```diff
@@ -8,4 +8,23 @@
...
+    let(:heading) { create :heading, :with_chapter }
+    let(:chapter) { heading.reload.chapter }
+
+    let(:path) { "/chapters/#{chapter.short_code}/headings" }
+
+    context 'when request asks for JSON format' do
```
willfish, line 18: I think we don't technically need this functionality/its YAGNI. I'm tempted to say we can just remove this everywhere until we need it

app/serializers/api/v2/csv/heading_serializer.rb
```diff
@@ -0,0 +1,24 @@
...
+          heading.commodities.any? do |commodity|
+            commodity.children.any?
+          end
+        end
+
```
willfish, line 20: Linttt me :)

app/models/heading.rb
```diff
@@ -106,12 +106,20 @@ def self.changes_for(depth = 0, conditions = {})
...
       :operation_date,
       :operation,
-      Sequel.as(depth, :depth)
+      Sequel.as(depth, :depth),
     ).where(conditions)
      .limit(TradeTariffBackend.change_count)
      .order(Sequel.desc(:operation_date, nulls: :last))
   end
 
+  def footnote_ids
```
willfish, line 115: I don't understand the addition of these chipolatas

app/controllers/api/v2/chapters_controller.rb
willfish: Actually, I'm confusing myself. We should just remove this part. We don't need it

willfish: Technically these apis are all just csv

app/controllers/api/v2/chapters_controller.rb
```diff
@@ -55,8 +73,16 @@ def find_chapter
...
+          "#{c_id}00000000"
+        else # it is single digit
```
willfish, line 83: I think we should just stick with the short code format to fetch the chapter headings

spec/requests/api/v2/chapters_controller_spec.rb
alexdesi: I agree it is, also there is some concern about performances, 
so instead of improving something that is not used, I'd remove it.

app/models/heading.rb
alexdesi: it is used by the JSON serializer,
but I'll remove it since we don't need it.

app/controllers/api/v2/chapters_controller.rb
alexdesi: I've simplified this, I think it's handy to have.

If you try this URL on production:
https://staging.trade-tariff.service.gov.uk/chapters/1
you would get "not found", which is weird since chapter 01 does actually exist.

willfish: We definitely don't want to change the protocol for representations of the goods nomenclature item id to start ignoring the first 0. This is a commonly understood way of representing the chapter in the harmonised system. If we change it we're adding complexity/create a non-conformant api around what are essentially worldwide standards.

willfish: Even in the short form its actually the wrong thing to do

app/controllers/api/v2/chapters_controller.rb
alexdesi: removed 👍

config/routes.rb
alexdesi: My thinking was it could be handy to support one digit,
also, the section endpoint supports one digit too.

np, just rolled back!

app/serializers/api/v2/csv/heading_serializer.rb
```diff
@@ -0,0 +1,19 @@
...
+        columns :goods_nomenclature_item_id,
+                :goods_nomenclature_sid,
+                :declarable,
+                :description,
+                :description_plain,
+                :formatted_description,
+                :producline_suffix
+
+        column(:leaf) { |heading| heading.commodities.empty? }
```
willfish, line 15: This executes the query and instantiates all of the commodities, sadly. If you do `heading.commodities_dataset.count` you push the check down to sequel without instantiate every commodity that belongs to the heading :)

spec/requests/api/v2/chapters_controller_spec.rb
```diff
@@ -8,4 +8,15 @@
       end
     end
   end
+
+  describe 'GET #headings' do
+    let(:heading) { create :heading, :with_chapter }
+    let(:chapter) { heading.reload.chapter }
+
+    context 'when request asks for CSV format' do
```
willfish, line 16: You don't need this context AFAIK

app/serializers/api/v2/csv/heading_serializer.rb
alexdesi: yeh, it works
dataset provides also the method `empty?`, which just looks nicer: `heading.commodities_dataset.empty?` 
https://www.rubydoc.info/github/jeremyevans/sequel/Sequel%2FDataset:empty%3F

spec/requests/api/v2/chapters_controller_spec.rb
```diff
@@ -8,4 +8,13 @@
       end
     end
   end
+
+  describe 'GET #headings' do
+    let(:heading) { create :heading, :with_chapter }
+    let(:chapter) { heading.reload.chapter }
```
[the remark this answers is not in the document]
alexdesi: Yes, otherwise it is `nil`. We do this in other places, probably we can fix that by changing the factory. Something to improve, but maybe not just now.

app/serializers/api/v2/csv/heading_serializer.rb
```diff
@@ -0,0 +1,19 @@
...
+
+        columns :goods_nomenclature_item_id,
+                :goods_nomenclature_sid,
+                :declarable,
+                :description,
+                :description_plain,
+                :formatted_description,
+                :producline_suffix
+
+        column(:leaf) { |heading| heading.commodities_dataset.empty? }
```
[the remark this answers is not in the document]
willfish:
```ruby
[8] pry(main)> Heading.first(goods_nomenclature_item_id: '0501000000').commodities_dataset.empty?
=> true
[9] pry(main)> Heading.first(goods_nomenclature_item_id: '0101000000').commodities_dataset.empty?
=> false
```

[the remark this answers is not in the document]
willfish: Is probably me overoptimising at this point, maybe

app/controllers/api/v2/chapters_controller.rb
```diff
@@ -45,8 +45,24 @@ def changes
...
 
+      def headings
+        chapter_headings = chapter.headings
+
+        respond_to do |format|
+          format.csv do
+            send_data(
+              Api::V2::Csv::HeadingSerializer.new(chapter_headings).serialized_csv,
+              type: 'text/csv; charset=utf-8; header=present',
+              disposition: "attachment; filename=chapter-#{params[:id]}-headings-#{actual_date.iso8601}.csv",
```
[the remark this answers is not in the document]
alexdesi: So we want to add the service name in the filename?
we did not do yet for the other endpoint?
maybe we should do a different PR for all the CSV endpoints

spec/requests/api/v2/chapters_controller_spec.rb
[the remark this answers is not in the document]
willfish: Cool :+1:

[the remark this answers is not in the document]
willfish: You'll actually want to rebase main to get a filename check test rolling

app/controllers/api/v2/chapters_controller.rb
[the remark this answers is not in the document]
willfish: Is merged: https://github.com/trade-tariff/trade-tariff-backend/pull/546

app/serializers/api/v2/csv/heading_serializer.rb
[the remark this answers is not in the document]
alexdesi: I've done some tests, and in most of the cases, the call is less than 1 second (in my local).

Only for 3 chapters, the call takes about 2.5 seconds (chapters 28, 29 and 84).

I think it is fine.

willfish (approved):

PR #545, opened 2022-04-25
HOTT-1543: Fix misuse of section id/position

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1543

### What?

I have added/removed/altered:

- [x] Alter section factory to reflect the primary key/id on a section being the same
- [x] Alter the section model so the primary key is clear
- [x] Removed incorrect section id constraints in routes file

### Why?

I am doing this because:

- This combination of incorrect configuration led me down a bit of a rabbit hole recently with intermittently failing specs

### DIFF
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -1,6 +1,8 @@
 class Section < Sequel::Model
   extend ActiveModel::Naming
 
+  set_primary_key [:position]
+
   plugin :timestamps
   plugin :active_model
   plugin :nullable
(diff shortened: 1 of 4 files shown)


app/models/section.rb
```diff
@@ -1,6 +1,8 @@
 class Section < Sequel::Model
   extend ActiveModel::Naming
 
+  set_primary_key [:position]
```
willfish, line 4: So this affects nothing and is just me being picky about being explicit about codifying the primary key the same we do in other models

app/models/section.rb
```diff
@@ -34,6 +36,7 @@ def chapter_ids
     chapters.pluck(:goods_nomenclature_sid)
   end
 
+
```
willfish, line 39: Remove extra line

jebw (approved):
LGTM

PR #546, opened 2022-04-25
HOTT-1544: Add service to csv endpoint filenames

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1544

### What?

I have added/removed/altered:

- [x] Added specs for csv endpoint filenames
- [x] Added service to csv endpoint filenames and updated specs

### Why?

I am doing this because:

- This is required for the user to distinguish between files from the two backends

### DIFF
--- a/app/controllers/api/v2/chapters_controller.rb
+++ b/app/controllers/api/v2/chapters_controller.rb
@@ -12,7 +12,7 @@ module Api
             send_data(
               Api::V2::Csv::ChapterSerializer.new(@chapters).serialized_csv,
               type: 'text/csv; charset=utf-8; header=present',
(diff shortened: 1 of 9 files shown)


jebw (approved):
LGTM - Code change looks great but there's already quite a few places needing to be changed and potentially more in the future - is it worth standadising this in the concern? 

eg

```ruby
serializer = Api::V2::Csv::ChapterSerializer.new(@chapters)
send_data(serializer, serializer.headers)
```

willfish:
I had exactly the same thoughts, yeah.

willfish:
I'll have a look in a separate PR, etc

PR #547, opened 2022-04-25
HOTT-1505 data migrator

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Removed historical data migrations
- [x] Removed the previous data migrator code base
- [x] Added a new data migrator which just uses the default migrations functionality from `Sequel` - this should make it easier for new developers to be aware data migrations need to be idempotent
- [x] Added a trivial generator to create new data migrations
- [x] Remove the 'done' marker for data migrations when rolling back the data sync

### Why?

I am doing this because:

- The existing data migrator is disable and badly understood
- Not integrated with the Data sync process

- Changes our data migration process

### Notes

- Does not include the ability to re-run migrations when running a Taric / CDS sync - this is already quite a big PR so I'll split that to a separate PR since it involves other changes


app/lib/tariff_synchronizer.rb
```diff
@@ -206,6 +206,8 @@ def rollback(rollback_date, keep: false)
               taric_update.delete
             end
           end
+
+          DataMigration.since(date_for_rollback).delete # Requeue data migrations
```
willfish, line 210: WOW!

willfish: So I can't believe this was missed as part of the rollbacks. It's basically a large advantage of the data migrator I'd imagined

willfish (approved):
Much better

PR #548, opened 2022-04-25
HOTT-1413 Fix the description on Footnote 808

jebw:
### Jira link

[HOTT-1413](https://transformuk.atlassian.net/browse/HOTT-1413)

### What?

I have added/removed/altered:

- [x] Added a data migration to correct the html code in the description on Footnote 808

### Why?

I am doing this because:

- The description does not render correctly
 
- Automatically changes production data via Data Migrations


willfish (approved):
Does the data migrator work with rollbacks?

jebw:
**ETL Rollbacks**

Yes - it drops the entries from the `data_migrations` table, this means those data migrations will get reapplied whenever the data migrator is re-run

**Data migrator rollback**

Yes - `rake data:rollback` will do what you expect, though in this case I didn't see much point adding a down step to 're-break' the HTML

In general data migrations need to be idempotent, so this one checks for the presence of the error, and only 'corrects' the HTML if it finds it

PR #554, opened 2022-04-26
HOTT-1505 Refactor Updates Synchronize Worker

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Moved scheduling of cache invalidation out of TariffSynchronizer and into UpdatesSynchronizerWorker
- [x] Changed TariffSynchronizer specs to not mockout calling `#apply` on `TaricUpdate`/`CdsUpdate`s themselves since this actually changes the flow in the calling class that checks after whether updates marked themselves as applied.
- [x] Makes some older Rake tasks aware of CDS
- [x] Reduce log level from sidekiq to avoid the need for silencing $stdout in some specs

### Why?

I am doing this because:

- I want to add another step to the ETL process before the cache invalidation (subsequent PR)
- This new step doesn't relate to the TariffSynchronizer itself
- The ETL process is more then just the TariffSynchronizer, currently its Synchronizer + Cache Invalidation + Index rebuild + Cache rebuild
- As it stands changes to the broader ETL process order potentially require the model for `TariffSynchronizer` to be changed, even if its unrelated

- Changes the critical TariffSynchronizer process - should run overnight in Dev environment prior to merge, and checks should be made that cache invalidation and reindex were run

### DIFF
--- a/spec/support/sidekiq.rb
+++ b/spec/support/sidekiq.rb
@@ -1,0 +1,1 @@
+Sidekiq.logger.level = Logger::WARN
--- a/app/lib/tariff_synchronizer.rb
+++ b/app/lib/tariff_synchronizer.rb
@@ -97,7 +97,7 @@ module TariffSynchronizer
   end
 
   # Taric
-  def apply(reindex_all_indexes: false)
+  def apply
     check_tariff_updates_failures
     check_sequence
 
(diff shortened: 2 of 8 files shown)


willfish (changes requested):
Lets hold off on this until I've merged my separation work. I will get round to it

spec/support/sidekiq.rb
```diff
@@ -0,0 +1 @@
+Sidekiq.logger.level = 2
```
willfish, line 1:
```suggestion
Sidekiq.logger.level =  Logger::WARN
```

willfish (approved):
On second thoughts it makes more sense to ship this and I just rebase at a later date.

PR #556, opened 2022-04-27
HOTT-1540: Adds coerced_measurement_unit_code to measurement_unit file

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1540

### What?

I have added/removed/altered:

- [x] Added coerced unit to the measurement units file
- [x] Changed version to today's date to reflect change
- [x] Updated reference to file

### Why?

I am doing this because:

- This will be used to determine whether the current input unit is uniq/a duplicate in the DC


alexdesi (approved):
✅ 🚀

PR #557, opened 2022-04-28
HOTT-1379 Return empty result for unknown commodities

jebw:
Previously search would return the heading for the missing commodity,
but this places behavioural logic in the backend and removes our ability
to handle empty searches differently on the frontend.

### Jira link

[HOTT-1379](https://transformuk.atlassian.net/browse/HOTT-1379)

### What?

I have added/removed/altered:

- [x] Return an empty result for a unknown commodity, instead of its parent heading

### Why?

I am doing this because:

- Currently we return an exact match, even though its not. 
- The frontend cannot differentiate between an exact match that it should treat as canonical and redirect the user to, and an exact match that actually wasn't and the user should be given options (ie the 404 history page)
- Returning no result makes it clear we didn't find a match for that commodity

- Includes behavioural changes to the search api - this isn't a documented API but is publicly accesible so may affect anyone integrating with the search we don't know about

### DIFF
--- a/app/services/search_service/exact_search.rb
+++ b/app/services/search_service/exact_search.rb
@@ -39,16 +39,17 @@ class SearchService
     private
 
     def find_heading(query)
+      query = normalise_shortened_code(query)
       query = SearchService::CodesMapping.check(query) || query
 
       Heading.actual
-             .by_code(query)
+             .by_declarable_code(query)
              .non_hidden
              .first
     end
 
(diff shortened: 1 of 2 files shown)


willfish (approved):
Code looks fine to me. I guess outcome of chat with Matt might influence what we do with this

jebw:
After discussion, concluded this is an edge case scenario so its fine

PR #558, opened 2022-04-28
HOTT-1540: Adds original unit to the measurement_units file

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1540

### What?

I have added/removed/altered:

- [x] Added original unit values

### Why?

I am doing this because:

- So we can present them to the user on the duty calculator screen


jebw (approved):
LGTM

PR #559, opened 2022-04-29
HOTT-1462 Add endopoint to retrieve the commodities, under Admin scope.

alexdesi:
### Jira link
[HOTT-1462](https://transformuk.atlassian.net/browse/HOTT-1462)

### What?
Generate full commodities code extract in CSV format via API call, in Admin area.

### Why?
I am doing this to avoid generating the CSV using the command line Python script.

## TODO:
-[] Generate Hierarchy
-[] Add missing tests


app/controllers/api/admin/commodities_controller.rb
```diff
@@ -25,6 +37,18 @@ def commodity_code
...
+      def all_commodities
+        commodity_groups = []
+
+        # Splitting the queries this way make the execution faster
+        (0..9).each do |starting_digit|
+          commodity_groups << Sequel::Model.db.fetch('select * from public.goods_nomenclature_export_new(?, ?) order by 2, 3', "#{starting_digit}%", actual_date)
+        end
+
+        # Running the queries and merging ...'
+        commodity_groups.map(&:to_a).flatten
```
willfish, line 50:
```suggestion
        commodity_groups.flat_map(&:all)
```

app/serializers/api/admin/commodities/commodity_csv_serializer.rb
```diff
@@ -0,0 +1,54 @@
+require_relative '../../../concerns/api/v2/csv/csv_serializer'
+
+module Api
+  module Admin
+    module Commodities
+      class CommodityCsvSerializer
+        include Api::V2::Csv::CsvSerializer
+
+        column :goods_nomenclature_sid, column_name: 'SID' do |row|
```
willfish, line 9: Why do you need the block?

app/controllers/api/admin/commodities_controller.rb
```diff
@@ -25,6 +37,18 @@ def commodity_code
       def productline_suffix
         params[:id].split('-', 2)[1] || '80'
       end
+
+      def all_commodities
+        commodity_groups = []
+
+        # Splitting the queries this way make the execution faster
+        (0..9).each do |starting_digit|
+          commodity_groups << Sequel::Model.db.fetch('select * from public.goods_nomenclature_export_new(?, ?) order by 2, 3', "#{starting_digit}%", actual_date)
```
jebw, line 46: Personally I think this is clearly a complex query so should be encapsulated on a model object, either an existing one (ie `Commodity`), or a separate object if its felt its sufficient unrelated to the Commodity

We should be able to add specs against the query itself then to check it is behaving as expected

db/migrate/20220509104200_create_goods_nomenclature_export_function.rb
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  down do
+    Sequel::Model.db.run('DROP FUNCTION IF EXISTS "public.goods_nomenclature_export_new"')
+  end
+end
+
+private
+
+def function_body
```
jebw, line 14: This is in the wrong namespace - this is at the global level, not within the database migration itself

I'd suggest just inlining this, ie 

```ruby
Sequel::Model.db.run <<~SQL.freeze
```

db/structure.sql
```diff
@@ -12,21 +12,29 @@ SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
 SELECT pg_catalog.set_config('search_path', '', false);
 SET check_function_bodies = false;
+SET xmloption = content;
```
jebw, line 15: This should only be making expected changes to the `structure.sql` file - ie just the `CREATE FUNCTION` block - all the extra changes here should be dropped

db/migrate/20220509104200_create_goods_nomenclature_export_function.rb
```diff
@@ -0,0 +1,114 @@
+Sequel.migration do
+  up do
+    Sequel::Model.db.run('CREATE SCHEMA IF NOT EXISTS "public"')
```
jebw, line 3: Is this necessary - wont `public` exist already?

jebw (changes requested):
Looks really good - some minor nits

* The query should be on a model with some specs to check its behaviour
* It would be better if the migration doesn't create a method of `function_body` in the global namespace - less vital since this will only occur during migrations
* The structure.sql has some extraneous changes in it which should be removed

app/controllers/api/admin/commodities_controller.rb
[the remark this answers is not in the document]
jebw: Thats column offsets, good spot - would be more readable to use column names

app/controllers/api/admin/commodities_controller.rb
alexdesi: 👍 cool, that's a bit faster.

app/serializers/api/admin/commodities/commodity_csv_serializer.rb
alexdesi: because inside the block, we are calling
`row[:goods_nomenclature_sid]`
and not
`row.goods_nomenclature_sid`

In other words, the attributes are not present as methods (they come from the `all_commodities` query)

db/migrate/20220509104200_create_goods_nomenclature_export_function.rb
alexdesi: Yes, I'll remove it.

db/migrate/20220509104200_create_goods_nomenclature_export_function.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  down do
+    Sequel::Model.db.run('DROP FUNCTION IF EXISTS "public.goods_nomenclature_export_new"')
+  end
+end
+
+private
+
+def function_body
+  <<~SQL.freeze
```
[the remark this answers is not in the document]
alexdesi: yep, 
also I've changed the loop to iterate all the chapters (and not the starting digit)

app/controllers/api/admin/commodities_controller.rb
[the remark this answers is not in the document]
alexdesi: yes, I like it!

app/services/admin/query_all_commodities.rb
```diff
@@ -0,0 +1,19 @@
+module Admin
+  class QueryAllCommodities
+    def self.call(actual_date)
+      # Splitting the queries this way make the execution faster
```
willfish, line 4:
```suggestion
      # Splitting the queries this way makes the execution faster
```

db/migrate/20220509104200_create_goods_nomenclature_export_function.rb
```diff
@@ -0,0 +1,118 @@
...
+                tmp_nomenclature;
+        END;
+
+        $function$;
+      SQL
+    )
+  end
+
+  down do
+    Sequel::Model.db.run('DROP FUNCTION IF EXISTS "public.goods_nomenclature_export_new"')
```
willfish, line 116:
```suggestion
    Sequel::Model.db.run('DROP FUNCTION IF EXISTS "public.fetch_chapter_commodities_for_date"')
```

app/serializers/api/admin/commodities/commodity_csv_serializer.rb
```diff
@@ -0,0 +1,54 @@
...
+          row[:goods_nomenclature_sid]
```
willfish, line 10: I wonder if it would be better to return Commodity instances just to simplify this

app/serializers/api/admin/commodities/commodity_csv_serializer.rb
```diff
@@ -0,0 +1,54 @@
+require_relative '../../../concerns/api/v2/csv/csv_serializer'
```
willfish, line 1: We should add a spec for this serializer to be consistent

willfish: https://github.com/trade-tariff/trade-tariff-backend/blob/HOTT-1462-all-commodities-csv/spec/serializers/api/v2/csv/commodity_serializer_spec.rb

spec/requests/admin/commodities_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    subject(:do_request) do
+      get api_commodities_path(format: :csv)
+      response.body
+    end
+
+    before do
+      allow(TimeMachine).to receive(:at).and_call_original
+    end
+
+    it 'contains the correct headers' do
```
willfish, line 12: We should move this test to the serializer and include some data verifications from a real row, too

spec/requests/admin/commodities_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      do_request
+
+      expect(response.body).to include("SID,Commodity code,Product line suffix,Description,Start date,End date,Indentation,End line,ItemIDPlusPLS\n")
+    end
+
+    it 'calls the TimeMachine with the correct Date' do
```
willfish, line 18: This actually doesn't add value (if you remember from our conversation previously). You're also using a different mechanism for filtering based on the actual date: you pass the actual date as a parameter to the query

willfish: I'd validate that our service receives the actual date

willfish:
```ruby
it 'passes the actual date to the query service' do
  allow(Admin::QueryAllCommodities).to receive(:call)
  do_request
  expect(Admin::QueryAllCommodities).to receive(:call).with(expected_actual_date)
end
```

db/migrate/20220509104200_create_goods_nomenclature_export_function.rb
alexdesi: 👍

spec/requests/admin/commodities_controller_spec.rb
alexdesi: Yes, I remember now.
thanks for writing the test!
( I've just changed `to receive` -> `to have_received` )

app/serializers/api/admin/commodities/commodity_csv_serializer.rb
alexdesi: done

app/serializers/api/admin/commodities/commodity_csv_serializer.rb
alexdesi: I think it would not work, in this case, we are serializing an array of hashes, returned from the fetch_chapter_commodities_for_date query.

spec/support/controller_spec_helper.rb
```diff
@@ -9,9 +9,9 @@ def pagination_pattern
   end
 
   def login_as_api_user(user = User.new)
-    request.env['warden'] = double('Authenticated API User',
-                                   authenticate!: true,
-                                   authenticated?: true,
-                                   user:)
+    double('Authenticated API User',
```
alexdesi, line 12: I don't see the value of this `request.env['warden']`, so I've removed it and see what happens.

app/serializers/api/admin/commodities/commodity_csv_serializer.rb
willfish: I don't think the data structure would be the problem. Possibly the attribute names between the commodity and the function result wouldn't match up totally, though.

db/migrate/20220509104200_create_goods_nomenclature_export_function.rb
```diff
@@ -0,0 +1,118 @@
...
+                            SELECT
+                                1
+                            FROM
+                                tmp_nomenclature
+                            WHERE
+                                goods_nomenclature_item_id LIKE CONCAT(nom_record.node, '%') AND goods_nomenclature_item_id <> nom_record.goods_nomenclature_item_id) THEN
+                            UPDATE
+                                tmp_nomenclature tn
+                            SET
+                                leaf = '1'
```
willfish, line 96: I have an irrational dislike for this script I think

app/services/admin/query_all_commodities.rb
```diff
@@ -0,0 +1,19 @@
+module Admin
+  class QueryAllCommodities
+    def self.call(actual_date)
```
willfish, line 3: Out of curiosity, what's the difference in performance now that we've removed the 0...9?

willfish: Does the api return in < 60 seconds

db/migrate/20220509104200_create_goods_nomenclature_export_function.rb
```diff
@@ -0,0 +1,118 @@
...
+                    -- Raise Notice 'goods nomenclature item id %', nom_record.goods_nomenclature_item_id;
+                    /* Leaf nodes have to have pls of 80 and no children having the same nomenclature code */
+                    IF nom_record.producline_suffix = '80' THEN
+                        IF LENGTH(nom_record.node) = 10 OR NOT EXISTS (
+                            SELECT
+                                1
+                            FROM
+                                tmp_nomenclature
+                            WHERE
+                                goods_nomenclature_item_id LIKE CONCAT(nom_record.node, '%') AND goods_nomenclature_item_id <> nom_record.goods_nomenclature_item_id) THEN
```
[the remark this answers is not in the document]
willfish:
```suggestion
                                goods_nomenclature_item_id LIKE CONCAT(nom_record.node, '%') AND goods_nomenclature_item_id <> nom_record.goods_nomenclature_item_id) AND number_indents <= nom_record.number_indents THEN
```

app/services/admin/query_all_commodities.rb
alexdesi: before - using 0..9% - about 45 sec,
now - using chapter short codes% - about 26 sec
I am not sure if the improvement is due only to that change (there might be other changes we've made)

willfish: That's a good outcome. It must be that change :)

db/migrate/20220509104200_create_goods_nomenclature_export_function.rb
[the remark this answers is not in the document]
alexdesi: I'd say, let's deliver this ticket (make the CSV downloadable), and then we can focus on the query (which is ugly, I agree)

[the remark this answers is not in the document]
willfish: The problem is, we have to spin up another migration to change that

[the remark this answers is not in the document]
willfish: And it is wrong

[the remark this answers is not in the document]
willfish: Matt has ported this change over to his own copy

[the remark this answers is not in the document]
alexdesi: ok, I am changing it - there are 3 occurrence

db/migrate/20220509104200_create_goods_nomenclature_export_function.rb
```diff
@@ -0,0 +1,120 @@
...
+                    IF nom_record.producline_suffix = '80' THEN
+                        IF LENGTH(nom_record.node) = 10 OR NOT EXISTS (
+                            SELECT
+                                1
+                            FROM
+                                tmp_nomenclature
+                            WHERE
+                                goods_nomenclature_item_id LIKE CONCAT(nom_record.node, '%')
+                                    AND goods_nomenclature_item_id <> nom_record.goods_nomenclature_item_id)
+                                    AND number_indents <= nom_record.number_indents THEN
```
willfish, line 94: So it turns out that this is actually wrong for a whole bunch of commodities even with this additional check on indents.

The grouping by node heuristic is actually not a precise way of knowing whether a goods nomenclature is part of the same tree. We have to go row by row to determine that. I think this is fine/is an improvement on what was there before but we should think about this at some stage

willfish (approved):

db/structure.sql
```diff
@@ -67,14 +366,121 @@ CREATE FUNCTION public.reassign_owned() RETURNS event_trigger
...
+    SELECT
+        *
+    FROM
+        tmp_nomenclature;
+END;
+$_$;
 
-SET default_with_oids = false;
+
+SET default_tablespace = '';
```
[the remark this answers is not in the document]
willfish: Not actually touched this so no changes should have occurred

jebw (approved):
:+1: LGTM :raised_hands:

PR #566, opened 2022-05-05
HOTT-1565 add Sidekiq healthcheck

jebw:
### Jira link

[HOTT-1565](https://transformuk.atlassian.net/browse/HOTT-1565)

### What?

I have added/removed/altered:

- [x] Added `Healthcheck` model
- [x] Removed odd global initializer for REVISION constant
- [x] Refactored `HealthcheckController` to use `Healthcheck` model 
- [x] Added Workers to record health of Sidekiq
- [x] Scheduled Workers to run hourly and with higher priority then other jobs
- [x] Report sidekiq health from `Healthcheck` model

### Why?

I am doing this because:

- We want to externally monitor the health of Sidekiq

- Changes how the key `/healthcheck` end point works


alexdesi (approved):
It looks good, we've aready discussed points on slack. ✅ 

I've just added two minor comments.

willfish (approved):
This is elegant and awesome :)

PR #568, opened 2022-05-09
HOTT-1574 Dropped Sections search refs admin api

jebw:
### Jira link

[HOTT-1574](https://transformuk.atlassian.net/browse/HOTT-1574)

### What?

I have added/removed/altered:

- [x] Removed the admin api for the Search References on Sections

### Why?

I am doing this because:

- It is no longer used by the admin interface so can be removed


willfish:
Do we want to clear out some of these bits, too: https://github.com/trade-tariff/trade-tariff-backend/blob/main/app/models/search_reference.rb#L24-L25

PR #573, opened 2022-05-10
HOTT-1565 Fix for flaky spec on Healthcheck model

jebw:
### Jira link

HOTT-1565

### What?

I have added/removed/altered:

- [x] Fixed flaky spec introduced as part of the Sidekiq healthcheck work

### Why?

I am doing this because:

- flaky specs result in false negatives for the whole team

### DIFF
--- a/spec/models/healthcheck_spec.rb
+++ b/spec/models/healthcheck_spec.rb
@@ -92,6 +92,12 @@ RSpec.describe Healthcheck do
   describe '.check' do
     subject { described_class.check }
 
-    it { is_expected.to include sidekiq: false }
+    before do
+      TradeTariffBackend.redis.set \
+        described_class::SIDEKIQ_KEY,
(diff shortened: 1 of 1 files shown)


spec/models/healthcheck_spec.rb
```diff
@@ -92,6 +92,12 @@
   describe '.check' do
     subject { described_class.check }
 
-    it { is_expected.to include sidekiq: false }
+    before do
```
willfish, line 95: I don't get it but it's fine

willfish (approved):

spec/models/healthcheck_spec.rb
jebw: Just verifies that it cascades through from `Healthcheck.check` to `Healthcheck.instance.check`

jebw: The fix is because depending upon ordering, the redis key may already be set from an earlier spec - which then results in a `true` instead of a `false` - explicitly setting the key to a known value before hand solves that

PR #580, opened 2022-05-12
Converted NewsItemController to be a request spec

jebw:
### Jira link

HOTT-????

### What?

I have added/removed/altered:

- [x] Converted NewsItemController to be a request spec

### Why?

I am doing this because:

- Controller specs are deprecated and this provides a nice example of using request specs with for admin area apis


willfish (approved):
:tada

alexdesi (approved):
fantastic, thanks for this! ✅

PR #584, opened 2022-05-13
HOTT-1619 Update Section title

jebw:
### Jira link

[HOTT-1619](https://transformuk.atlassian.net/browse/HOTT-1619)

### What?

I have added/removed/altered:

- [x] Added a data migration to update the title of Section III
- [x] Fixed default ordering of Section model
- [x] Fixed default ordering of DataMigration model

### Why?

I am doing this because:

- We want to amend the title of Section III
- Previously API returned sections in the order of 'least recently modified' first - which meant the newly modified Section III was at the end, not between Section II and Section IV as it should be
- The DataMigrations also have a natural order, so we should follow that in retrieval to avoid confusing future developers - note this model isn't used by the migration system itself, its just a convenience model I've added for working with records from the Data Migrator

### DIFF
--- a/app/models/data_migration.rb
+++ b/app/models/data_migration.rb
@@ -1,6 +1,8 @@
 class DataMigration < Sequel::Model
   VERSION_FORMAT = /\A20\d{12}\z/
 
+  set_dataset order(Sequel.asc(:filename))
+
   dataset_module do
     def version(version)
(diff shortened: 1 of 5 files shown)


willfish (approved):
Makes loads of sense :+1:

PR #585, opened 2022-05-13
HOTT-1578 Fix csv header to be consistent with other csv/json.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1578

### What?
Fix csv header to be consistent with others csv/json.


willfish:
Chatted about this

PR #590, opened 2022-05-17
HOTT-1582: Tidy mappers

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1582

### What?

I have added/removed/altered:

- [x] Removes defensive failures that will be caught with unit tests
- [x] Removes 2 unused setters on the entity mappers
- [x] Lints every mapper

### Why?

I am doing this because:

- All of these seemed unnecessary to me/add misdirection to the programmer

### DIFF
--- a/app/lib/cds_importer/entity_mapper/additional_code_description_mapper.rb
+++ b/app/lib/cds_importer/entity_mapper/additional_code_description_mapper.rb
@@ -7,7 +7,7 @@ class CdsImporter
 
       self.mapping_path = 'additionalCodeDescriptionPeriod.additionalCodeDescription'.freeze
 
-      self.exclude_mapping = ['validityStartDate', 'validityEndDate'].freeze
+      self.exclude_mapping = %w[validityStartDate validityEndDate].freeze
 
       self.entity_mapping = base_mapping.merge(
         'additionalCodeDescriptionPeriod.sid' => :additional_code_description_period_sid,
@@ -15,12 +15,8 @@ class CdsImporter
         'sid' => :additional_code_sid,
         'additionalCodeType.additionalCodeTypeId' => :additional_code_type_id,
         'additionalCodeCode' => :additional_code,
-        "#{mapping_path}.description" => :description
+        "#{mapping_path}.description" => :description,
       ).freeze
-
-      self.entity_mapping_key_as_array = mapping_with_key_as_array.freeze
-
-      self.entity_mapping_keys_to_parse = mapping_keys_to_parse.freeze
     end
   end
 end
--- a/app/lib/cds_importer/entity_mapper/base_mapper.rb
+++ b/app/lib/cds_importer/entity_mapper/base_mapper.rb
@@ -28,13 +28,13 @@ class CdsImporter
         end
 
         def mapping_with_key_as_array
-          entity_mapping.keys.each_with_object({}) do |key, memo|
+          @mapping_with_key_as_array ||= entity_mapping.keys.each_with_object({}) do |key, memo|
             memo[key.split(PATH_SEPARATOR)] = entity_mapping[key]
           end
         end
 
         def mapping_keys_to_parse
-          mapping_with_key_as_array.keys.reject do |key|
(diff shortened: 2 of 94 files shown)


app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -108,42 +108,21 @@ def parse
         if mapping_path.present?
           expanded.select! { |values| values.dig(*mapping_path.split(PATH_SEPARATOR)).present? }
         end
-        expanded.map { |values| create_instance(values) }
+        expanded.map(&method(:build_instance))
```
alexdesi, line 111: I guess values are passed implicitly here 🤔

alexdesi (approved):
I looks good to me! ✅

app/lib/cds_importer/entity_mapper/base_mapper.rb
jebw: @alexdesi Yeh, works the same way as symbol to proc - symbol to proc in effect recognises its a proc and just calls `#method` on the object handed in - where as this passes in the method from the calling class

Its the equivalent of javascript's ability to use functions like variables

```javascript
function sayHello() {
  console.log("Hello world!")
}

setTimeout(sayHello, 5000)
```

app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -108,42 +108,21 @@ def parse
...
-
-      def entity_class
-        self.class.entity_class.presence || raise(ArgumentError, "entity_class has not been defined: #{self.class}")
-      end
-
-      def mapping_path
-        self.class.mapping_path.presence
-      end
-
-      def entity_mapping
```
jebw, line 132: Whilst the `delegate` approach is obviously much simpler, this will help future developers if they are getting odd behaviour whilst trying to implement new root entities?

app/lib/cds_importer/entity_mapper/additional_code_description_mapper.rb
```diff
@@ -7,20 +7,16 @@ class AdditionalCodeDescriptionMapper < BaseMapper
...
         'additionalCodeDescriptionPeriod.sid' => :additional_code_description_period_sid,
         "#{mapping_path}.language.languageId" => :language_id,
         'sid' => :additional_code_sid,
         'additionalCodeType.additionalCodeTypeId' => :additional_code_type_id,
         'additionalCodeCode' => :additional_code,
-        "#{mapping_path}.description" => :description
+        "#{mapping_path}.description" => :description,
       ).freeze
-
-      self.entity_mapping_key_as_array = mapping_with_key_as_array.freeze
```
jebw, line 21: I _think_ this change will mean the mapper keeps recomputing the entity keys. If I was to guess some of this is working around the original developer not understanding class level instance variables

If this is getting removed then `mapping_with_key_as_array` should be memoized at class level, likewise `mapping_keys_to_parse` ?

jebw (approved):
Looks good - 2 changes of note

1. It removes the debug info for the developer trying to add new root entities - the better specs probably will help with debugging these type of issues instead though
2. The 'keys' variables being dropped - seems fine but is a (probably minor) performance drop - the methods they are calling could be memoized at class level to avoid repeatedly recomputing the lists of keys

app/lib/cds_importer/entity_mapper/base_mapper.rb
willfish: It's actually super difficult to get wrong now. You add a spec with your real xml node and check it parses cleanly/produces the resource you expect it too. I'll make it even more obvious with the specs coming for insertion/soft deletion, too

app/lib/cds_importer/entity_mapper/additional_code_description_mapper.rb
willfish: I can add memoization for sure. I'd be blown away if it improved performance at all, though

PR #591, opened 2022-05-18
HOTT-1545 Return all the search ref when query letter is not provided.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1545

### What?
Return all the search ref when the letter param is not provided.

### DIFF
--- a/app/controllers/api/v2/search_references_controller.rb
+++ b/app/controllers/api/v2/search_references_controller.rb
@@ -2,15 +2,15 @@ module Api
   module V2
     class SearchReferencesController < ApiController
       def index
-        @search_references = SearchReference.for_letter(letter).by_title.all
+        search_references = SearchReference.for_letter(letter).by_title.all
 
-        render json: Api::V2::SearchReferenceSerializer.new(@search_references).serializable_hash
+        render json: Api::V2::SearchReferenceSerializer.new(search_references).serializable_hash
       end
 
       private
 
       def letter
-        params.fetch(:query, {}).fetch(:letter, 'a')
+        params.dig(:query, :letter) || ''
       end
     end
   end
--- a/app/controllers/api/v2/search_references_controller.rb
+++ b/app/controllers/api/v2/search_references_controller.rb
@@ -2,7 +2,7 @@ module Api
   module V2
     class SearchReferencesController < ApiController
       def index
-        search_references = SearchReference.for_letter(letter).by_title.all
+        search_references = SearchReference.for_letter(letter).all
 
         render json: Api::V2::SearchReferenceSerializer.new(search_references).serializable_hash
       end
--- a/spec/controllers/api/v2/search_references_controller_spec.rb
+++ b/spec/controllers/api/v2/search_references_controller_spec.rb
@@ -41,28 +41,11 @@ RSpec.describe Api::V2::SearchReferencesController do
   end
 
   context 'with no letter param provided' do
-    let(:pattern) do
-      {
-        data: [
-          {
-            id: String,
-            type: String,
-            attributes: {
-              title: String,
-              referenced_class: 'Heading',
-              referenced_id: String,
-              productline_suffix: String,
-            },
-          },
-        ],
-      }
-    end
+    it 'does not filter by letter' do
+      get :index, format: :json
 
-    it 'peforms lookup with letter A by default' do
-      get :index,
-          format: :json
-
-      expect(response.body).to match_json_expression pattern
+      ref_count = JSON.parse(response.body)['data'].count
+      expect(ref_count).to eq(3)
     end
   end
 end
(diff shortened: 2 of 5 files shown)


app/controllers/api/v2/search_references_controller.rb
```diff
@@ -2,15 +2,17 @@ module Api
   module V2
     class SearchReferencesController < ApiController
       def index
-        @search_references = SearchReference.for_letter(letter).by_title.all
+        search_references = letter.present? ? SearchReference.for_letter(letter) : SearchReference
```
alexdesi, line 5: even only this:
`search_references = SearchReference.for_letter(letter)`
would work,
but if the letter is not present it's better just to refer to `SearchReference` for better performaces.

app/controllers/api/v2/search_references_controller.rb
```diff
@@ -2,15 +2,17 @@ module Api
...
+        render json: Api::V2::SearchReferenceSerializer.new(
+          search_references.by_title.all,
+        ).serializable_hash
       end
 
       private
 
       def letter
-        params.fetch(:query, {}).fetch(:letter, 'a')
+        params.fetch(:query, {}).fetch(:letter, '')
```
willfish, line 15:
```suggestion
        params.dig(:query, :letter).presence || ''
```

app/controllers/api/v2/search_references_controller.rb
willfish: The negative branch of this isn't actually the same kind of thing and the name doesn't make sense.

I'd probably just move this and the `by_title.all` to its own method

```ruby
def search_references
  starting_query = letter.present? ? SearchReference.for_letter(letter) : SearchReference

  starting_query.by_title.all
end
```

willfish: I also think its valid that if the user does not specify a letter the for_letter method returns just a dataset

spec/controllers/api/v2/search_references_controller_spec.rb
```diff
@@ -58,11 +58,11 @@
       }
     end
 
-    it 'peforms lookup with letter A by default' do
-      get :index,
-          format: :json
+    it 'does not filter by letter' do
```
willfish, line 61: I'd surround this in a context as it doesn't make sense without one

willfish: Ignore this, just looked at the file the previous example description was adding context it shouldn't have

app/controllers/api/v2/search_references_controller.rb
alexdesi: yep, that's nicer :)

willfish (approved):

spec/controllers/api/v2/search_references_controller_spec.rb
alexdesi: it's already inside  the context 'with no letter param provided' do
what we do not need is the pattern since this is just checking the counts.

app/controllers/api/v2/search_references_controller.rb
alexdesi: I added that condition for performance reasons, but I can see that we only save a small fraction of a second, so it's not worthy.
So, let's make the model return the right search_references.

alexdesi: which is the default behaviour after all

willfish (approved):

spec/requests/api/admin/search_references_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
+RSpec.describe Api::Admin::SearchReferencesController, type: :request do
```
alexdesi, line 1: Missing request for SearchReferencesController in Admin.

Even if this is similar to the SearchReferencesController in V2,
those are in separate scopes, so I think a bit of duplication is acceptable.

spec/requests/api/admin/search_references_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      it 'does not filter by letter' do
+        do_request
+
+        search_ref_count = JSON.parse(response.body)['data'].count
+        expect(search_ref_count).to eq(3)
+      end
+    end
+  end
+
```
willfish, line 40: Am guessing this is an accident/not linted :)

spec/requests/api/admin/search_references_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
+RSpec.describe Api::Admin::SearchReferencesController, type: :request do
+  before do
+    create :search_reference, referenced: create(:heading), title: 'aa'
+    create :search_reference, referenced: create(:chapter), title: 'bb'
+    create :search_reference, referenced: create(:commodity), title: 'bb'
+  end
+
+  describe 'GET #index' do
+    subject(:do_request) do
+      authenticated_get api_search_references_path(params: query_letter, format: :json)
```
willfish, line 10: These path helpers are awful. Missing the admin scope you might confuse it with the v2/v1 apis :facepalm:

PR #593, opened 2022-05-19
HOTT-1583: Handle soft deletes of primary footnote xml node

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1583

### What?

I have added/removed/altered:

- [x] Added support for soft deletes of the primary footnote xml node
- [x] Added specs for this new behaviour
- [x] Put the behaviour behind a feature flag

### Why?

I am doing this because:

- This is required to reflect the implicit/expected behaviour of the ETL process from CDS's perspective


app/lib/cds_importer/entity_mapper.rb
```diff
@@ -32,8 +32,21 @@ def import
...
+      #
+      # This is required because CDS do not mark secondary and tertiary nested xml nodes for deletion
+      # themselves and we have to ignore importing them in the event of the primary xml node being deleted as
+      # this is managed by a separate callback process (see each primary entity mapper for what gets soft deleted).
+      def applicable_mappers_for(key, xml_node)
+        mappers = all_mappers.select { |mapper| mapper&.mapping_root == key }.sort_by(&:sort_key)
+
+        primary_mapper = mappers.first
+
+        if primary_mapper && primary_mapper.destroy_operation?(xml_node)
```
alexdesi, line 45: maybe?
`if primary_mapper&.destroy_operation?(xml_node)`

app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -57,20 +76,11 @@ def before_oplog_inserts(&block)
...
-      BASE_MAPPING = {
-        'validityStartDate' => :validity_start_date,
-        'validityEndDate' => :validity_end_date,
-        'metainfo.origin' => :national,
-        'metainfo.opType' => :operation,
-        'metainfo.transactionDate' => :operation_date,
-      }.freeze
+        def primary?
+          name.match(/\ACdsImporter::EntityMapper::(?<entity_class>.*)Mapper\z/).try(:[], :entity_class) == entity_class
+        end
```
alexdesi, line 82: this is pretty dense,
you can improve readability here, maybe splitting it into two lines?

```
entity_mapper = name.match(/\ACdsImporter::EntityMapper::(?<entity_class>.*)Mapper\z/)

entity_mapper.try(:[], :entity_class) == entity_class
```

alexdesi: it is not clear how the concept of `primary` is deduced?

app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -57,20 +76,11 @@ def before_oplog_inserts(&block)
...
-      PATH_SEPARATOR = '.'.freeze
-      METAINFO = 'metainfo'.freeze
-      BASE_MAPPING = {
-        'validityStartDate' => :validity_start_date,
-        'validityEndDate' => :validity_end_date,
-        'metainfo.origin' => :national,
-        'metainfo.opType' => :operation,
-        'metainfo.transactionDate' => :operation_date,
-      }.freeze
+        def primary?
```
alexdesi, line 80: maybe a more explicit name? primary what?

app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -48,6 +61,12 @@ def sort_key
           "#{mapping_path.to_s.length}#{name}"
         end
 
+        def destroy_operation?(xml_node)
+          xml_node.dig('metainfo', 'opType') == Sequel::Plugins::Oplog::DESTROY_OPERATION &&
+            primary? &&
+            TradeTariffBackend.handle_soft_deletes?
+        end
```
alexdesi, line 68: I am wondering if `destroy_operation?` should be inside the mapper? 🤔 
... anyway, not sure yet what would be the best place for it

app/lib/cds_importer/entity_mapper/base_mapper.rb
willfish: primary mapper

willfish: I might add a comment

app/lib/cds_importer/entity_mapper/base_mapper.rb
willfish: Yeah, that was my initial implementation and I came into callback scoping rules. I'm now passing the mapper instance to the callbacks that are interested in it

app/lib/cds_importer/entity_mapper/base_mapper.rb
willfish: Added a comment and tidied this up so the derived_entity_class is separate from the business logic method

spec/support/shared_examples/an_entity_mapper_update_operation.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.shared_examples_for 'an entity mapper update operation' do |relation|
+  it { expect { entity_mapper.import }.to change { relation::Operation.where(operation: 'U').count }.by(1) }
+end
```
alexdesi, line 3: We might merge all these entity mapper operations in just one file, and also add the 
"an xml node and mapper that are not a destroy operation"

willfish: I prefer the top level explicitness and the modular approach gives us the ability to have different setups for each depending on the mapper

alexdesi (approved):
Good one! 🎉

app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -59,27 +82,14 @@ def before_building_model(&block)
...
-      BASE_MAPPING = {
-        'validityStartDate' => :validity_start_date,
-        'validityEndDate' => :validity_end_date,
-        'metainfo.origin' => :national,
-        'metainfo.opType' => :operation,
-        'metainfo.transactionDate' => :operation_date,
-      }.freeze
-
-      def initialize(values)
-        @values = values.slice(*mapping_with_key_as_array.keys.map { |k| k[0] }.uniq)
```
willfish, line 76: This side effect in the constructor literally adds no value in terms of the built entity that results from calling the `#parse`.

I typically really don't like side effects in the constructor because it's hard to test the result of the side effect on the instance/collaborator. In this case the side effect was noop/does nothing

PR #594, opened 2022-05-20
HOTT-1567 Update Rails cache to use the redis cache store.

alexdesi:
### What?
Update Rails cache to use the Redis cache store.

### Why?
The reason of this PR is because Rails 5.2.0 (and following versions) includes a Redis cache store out of the box, so you don't need this gem anymore if you just need to store the cache in Redis.

### Jira link
https://transformuk.atlassian.net/browse/HOTT-1567


jebw (approved):
LGTM

PR #595, opened 2022-05-23
HOTT-1583: Enable soft deletion of missing footnoteDescriptionPeriod entities

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1583

### What?

I have added/removed/altered:

- [x] Added secondary soft deletion callback registration method
- [x] Added secondary soft deletion callback invocation on the footnote mapper
- [x] Added secondary soft deletion specs
- [x] Feature flag soft deletion

### Why?

I am doing this because:

- CDS send us files that do not explicitly indicate that resources should be deleted and instead are expecting us to work out what needs removing and to remove them.

### DIFF
--- a/app/lib/cds_importer/entity_mapper.rb
+++ b/app/lib/cds_importer/entity_mapper.rb
@@ -19,14 +19,14 @@ class CdsImporter
 
         instances = mapper.parse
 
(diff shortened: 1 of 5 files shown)


alexdesi (approved):
It looks good to me ✅

PR #597, opened 2022-05-24
HOTT-1599: Handle composite primary keys during missing secondary node deletion

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1599

### What?

I have added/removed/altered:

- [x] Added support for pulling out composite primary keys for secondary resources for soft deletion
- [x] Added control description period to validate what happens when we do not delete an existing period

### Why?

I am doing this because:

- This is required to support other more complex/dynamic relationships as part of measure missing secondary node soft deletion


alexdesi (approved):
it looks good ✅

PR #599, opened 2022-05-25
HOTT-1599: Handle deletions on the measure mapper

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added support for soft deletes when the measure primary is a destroy operation
- [x] Added support for missing soft deletes of the measure mapper as a primary
- [x] Added support for missing soft deletes of the measure condition mapper as a primary
- [x] Added configuration to affect mappers
- [x] Added specs to cover changes
- [x] Refactored xml entity enumeration so to avoid duplication of parse method
- [x] Added missing factory implementations
- [x] Removed some factory side effects implementations

### Why?

I am doing this because:

### DIFF
--- a/app/lib/cds_importer/entity_mapper/base_mapper.rb
+++ b/app/lib/cds_importer/entity_mapper/base_mapper.rb
@@ -54,11 +54,11 @@ class CdsImporter
 
         # Register a callback to soft delete missing entities indicated by the passed in secondary mappers
         def delete_missing_entities(*secondary_mappers)
-          before_oplog_inserts do |xml_node, _mapper_instance, model_instance|
+          before_oplog_inserts do |xml_node, _mapper_instance, primary_model_instance|
             if TradeTariffBackend.handle_soft_deletes?
               secondary_mappers.each do |secondary_mapper|
-                database_entities = database_entities_for(model_instance, secondary_mapper)
-                xml_node_entities = xml_entities_for(xml_node, secondary_mapper)
+                database_entities = secondary_mapper.database_entities_for(primary_model_instance)
+                xml_node_entities = secondary_mapper.xml_entities_for(xml_node)
                 missing_entities = database_entities - xml_node_entities
                 missing_entity_filter = secondary_mapper.missing_entity_filter_for(missing_entities)
 
@@ -130,6 +130,20 @@ class CdsImporter
           Array.wrap(entity.primary_key)
         end
 
+        def database_entities_for(primary_model_instance)
+          filter = filter_for(primary_model_instance)
+
+          entity.where(filter).pluck(*entity.primary_key).map do |composite_primary_key|
+            Array.wrap(composite_primary_key)
+          end
+        end
+
+        def xml_entities_for(xml_node)
+          xml_node_entities = new(xml_node).parse
+
+          xml_node_entities.pluck(*entity.primary_key).map(&Array.method(:wrap))
+        end
+
         protected
 
         def before_oplog_inserts(&block)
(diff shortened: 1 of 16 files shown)


app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -108,26 +108,22 @@ def missing_entity_filter_for(missing_entities)
...
         end
 
-        def relative_primary_key_paths_for_secondary_node
-          primary_key_paths_for_secondary_node.each_with_object({}) do |path, acc|
-            relative_path = path.sub("#{mapping_path}.", '')
-
-            acc[relative_path] = primary_key_mapping[path]
-          end
+        def entity_composite_primary_key
+          Array.wrap(entity.primary_key)
```
alexdesi, line 112: Nice, I did know about the wrap method. It should be about
`[ entity.primary_key ]` 
but more explicit

app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -108,26 +108,22 @@ def missing_entity_filter_for(missing_entities)
...
         end
 
-        def primary_key_paths_for_primary_node
-          primary_node_paths = primary_key_mapping.keys - primary_key_paths_for_secondary_node
+        def database_entities_for(primary_model_instance)
+          filter = filter_for(primary_model_instance)
 
-          primary_node_paths.index_with { |path| primary_key_mapping[path] }
+          entity.where(filter).pluck(*entity.primary_key).map do |composite_primary_key|
```
alexdesi, line 118: I wonder why we need to wrap after pluck 🤔

app/lib/cds_importer/entity_mapper/base_mapper.rb
willfish: It actually behaves slightly differently:

```ruby
thing = [:foo]
Array.wrap(thing) => [:foo]

thing = :foo
Array.wrap(thing) => [:foo]

thing = [:foo]
[thing] => [[:foo]]

thing = :foo
[thing] = [:foo]
```

app/lib/cds_importer/entity_mapper/base_mapper.rb
willfish: When the primary key is singular (e.g. not a composite key), pluck returns an array with a depth of 1:

```ruby
[:foo, :bar]
```

When the primary key is a composite, pluck returns an array with a depth of 2:

```ruby
[[:foo, :baz], [:bar, :qux]]
```
I wrap so can iterate regardless of what the pluck method returns

spec/lib/cds_importer/entity_mapper/measure_mapper_spec.rb
```diff
@@ -56,4 +56,187 @@
...
+      it_behaves_like 'an entity mapper destroy operation', Measure
+      it_behaves_like 'an entity mapper destroy operation', FootnoteAssociationMeasure
+      it_behaves_like 'an entity mapper destroy operation', MeasureComponent
+      it_behaves_like 'an entity mapper destroy operation', MeasureCondition
+      it_behaves_like 'an entity mapper destroy operation', MeasureConditionComponent
+      it_behaves_like 'an entity mapper destroy operation', MeasureExcludedGeographicalArea
+      it_behaves_like 'an entity mapper destroy operation', MeasurePartialTemporaryStop
+    end
+
+    context 'when there are missing secondary entities to be soft deleted' do
```
alexdesi, line 216: I don't understand this test (and the other with "... to be soft deleted" ,
What is the feature in this test that makes me think that the missing secondary entities have to be soft deleted?

spec/support/shared_examples/an_entity_mapper_missing_entity_destroy_operation.rb
```diff
@@ -23,17 +18,17 @@
...
-
-      expect(after_import_primary_keys).to include(*before_import_primary_keys)
+    # rubocop:disable RSpec/MultipleExpectations
+    it "does not hide the missing xml nodes from the #{model} view" do
+      if has_hard_deletes
+        expect { entity_mapper.import }.to change_view_count.by(-1)
+      else
+        expect { entity_mapper.import }.not_to change_view_count
+      end
     end
```
alexdesi, line 28: it looks this should be split into two different contexts:
with has hard deletes
and without hard deletes

alexdesi (approved):
it looks good to me ✅ 
just a couple of minor comments,
the tests with conditional at the end sounds are a bit strange

spec/support/shared_examples/an_entity_mapper_missing_entity_destroy_operation.rb
willfish: Sadly that's not actually possible

willfish: I'll be getting rid of this in like 2 weeks

spec/lib/cds_importer/entity_mapper/measure_mapper_spec.rb
willfish: Added some context that should help :smile:

PR #601, opened 2022-05-26
HOTT-1585: Handle additional code primary soft deletes

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added support for soft deletes when the additional code primary is a destroy operation
- [x] Added support for missing soft deletes of the additional code mapper as a primary
- [x] Added specs to cover changes
- [x] Removed some configuration that is no longer used

### Why?

I am doing this because:

### DIFF
--- a/spec/lib/cds_importer/entity_mapper/additional_code_mapper_spec.rb
+++ b/spec/lib/cds_importer/entity_mapper/additional_code_mapper_spec.rb
@@ -1,33 +1,126 @@
 RSpec.describe CdsImporter::EntityMapper::AdditionalCodeMapper do
-  it_behaves_like 'an entity mapper', 'AdditionalCode', 'AdditionalCode' do
-    let(:xml_node) do
-      {
-        'sid' => '3084',
-        'additionalCodeCode' => '169',
-        'validityEndDate' => '1996-06-14T23:59:59',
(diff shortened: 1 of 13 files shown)


spec/lib/cds_importer/entity_mapper/additional_code_mapper_spec.rb
```diff
@@ -1,33 +1,126 @@
...
+  let(:operation) { 'C' }
 
+  it_behaves_like 'an entity mapper', 'AdditionalCode', 'AdditionalCode' do
     let(:expected_values) do
       {
         validity_start_date: Time.zone.parse('1991-06-01T00:00:00.000Z'),
         validity_end_date: Time.zone.parse('1996-06-14T23:59:59.000Z'),
         national: false,
-        operation: 'U',
+        operation: 'C',
```
alexdesi, line 58: should this be:
`operation: operation`
?

alexdesi (approved):

spec/lib/cds_importer/entity_mapper/additional_code_mapper_spec.rb
willfish: chatted offline. this is basically part of removing duplication where a new default exists on the xml node

PR #602, opened 2022-05-26
HOTT-1585: Adds migration for additional code type description view language

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added migration to support uniquely identifying additional code type descriptions in the view
- [x] Added primary key for language id in oplog and model

### Why?

I am doing this because:

- This is how we uniquely identify the additional code type description when there are more than one languages we might pull out
- This was noticed as I was adding coverage for clearing missing versions of these from an additional code description type primary


alexdesi (approved):
✅

PR #603, opened 2022-05-27
HOTT-1586: Handle soft deletes for additional code type

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Handle soft deletes for additional code type
- [x] Add coverage for above

### Why?

I am doing this because:


alexdesi (commented):
👍

PR #605, opened 2022-05-30
HOTT-1586: Removes allocations

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1584

### What?

I have added/removed/altered:

- [x] Removed allocations from insert log

### Why?

I am doing this because:

- These are all coming back as 0


alexdesi (approved):
🧹🧹🧹It looks good to me!

PR #606, opened 2022-05-30
HOTT-1585: Adds show endpoint to admin updates controller

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added show endpoint to admin updates controller

### Why?

I am doing this because:

- This is required so we can show the inserts of a specific update

### DIFF
--- a/spec/controllers/api/admin/updates_controller_spec.rb
+++ b/spec/controllers/api/admin/updates_controller_spec.rb
@@ -42,4 +42,42 @@ RSpec.describe Api::Admin::UpdatesController do
       end
     end
   end
+
+  describe 'GET #show' do
+    context 'when records are present' do
+      subject(:do_request) { get :show, params: { id: update.to_param, format: :json } }
+
+      let(:pattern) do
+        {
+          data: {
+            id: String,
+            type: 'tariff_update',
+            attributes: {
+              update_type: 'TariffSynchronizer::CdsUpdate',
+              state: String,
+              filename: String,
+              created_at: String,
+            }.ignore_extra_keys!,
+          }.ignore_extra_keys!,
+        }
+      end
+
+      let(:update) { create(:cds_update) }
+
+      it { expect(do_request.body).to match_json_expression pattern }
+    end
+
+    context 'when records are not present' do
+      subject(:do_request) { get :show, params: { id: 'foo', format: :json } }
+
+      let(:pattern) do
+        {
+          error: 'not found',
+          url: 'http://test.host/admin/updates/foo',
+        }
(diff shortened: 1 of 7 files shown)


spec/unit/tariff_synchronizer/base_update_spec.rb
```diff
@@ -233,4 +233,39 @@
...
+      let(:service) { 'uk' }
+
+      it { expect { described_class.by_filename(xi_update.to_param) }.to raise_error(Sequel::RecordNotFound) }
+    end
+
+    context 'when on the uk service and fetching the uk update' do
+      let(:service) { 'uk' }
+
+      it { expect(described_class.by_filename(uk_update.to_param)).to be_a(TariffSynchronizer::CdsUpdate) }
+    end
```
alexdesi, line 257: It would be nice to make explicit that when you fetch the UK update, it'll get the GZIP file

alexdesi: and when you fetch the XI it'll use the XML.

spec/controllers/api/admin/updates_controller_spec.rb
```diff
@@ -42,4 +42,42 @@
...
+      let(:pattern) do
+        {
+          error: 'not found',
+          url: 'http://test.host/admin/updates/foo',
+        }
+      end
+
+      it { expect(do_request.body).to match_json_expression pattern }
+    end
+  end
```
alexdesi, line 82: I thought we more in favour of request spec, rather then controller.

alexdesi (commented):
it sounds good, just a couple of comments

alexdesi (approved):
✅

spec/unit/tariff_synchronizer/base_update_spec.rb
willfish: I deliberately pushed the logic down to the model so we don't have to orchestrate all of that

willfish: Misunderstood the point, here. Will have a look about making it more explicit.

spec/controllers/api/admin/updates_controller_spec.rb
willfish: I can migrate this separately

spec/unit/tariff_synchronizer/base_update_spec.rb
willfish: So thinking about  it, that should be an implementation detail that the caller does not know about.

PR #607, opened 2022-05-30
HOTT-1585: Adjust oplog row filenames to match the current update

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added support for update filenames on missing/cascading xml node deletions

### Why?

I am doing this because:

- This was missed when enumerating missing and cascading records and is essential for rollbacks to work

### DIFF
--- a/app/lib/cds_importer/entity_mapper/base_mapper.rb
+++ b/app/lib/cds_importer/entity_mapper/base_mapper.rb
@@ -54,7 +54,7 @@ class CdsImporter
 
         # Register a callback to soft delete missing entities indicated by the passed in secondary mappers
         def delete_missing_entities(*secondary_mappers)
-          before_oplog_inserts do |xml_node, _mapper_instance, primary_model_instance|
+          before_oplog_inserts do |xml_node, mapper_instance, primary_model_instance|
             if TradeTariffBackend.handle_soft_deletes?
               secondary_mappers.each do |secondary_mapper|
                 database_entities = secondary_mapper.database_entities_for(primary_model_instance)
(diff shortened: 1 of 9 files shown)


app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -206,6 +206,10 @@ def primary?
         derived_entity_class == entity_class
       end
 
+      def filename
```
willfish, line 209: Add test for this

app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -150,6 +148,8 @@ def before_building_model(&block)
         end
       end
 
+      attr_reader :xml_node
```
willfish, line 151: Add test for this

app/lib/cds_importer/record_inserter.rb
```diff
@@ -0,0 +1,56 @@
+class CdsImporter
```
willfish, line 1: Write tests for me

app/lib/cds_importer/record_inserter.rb
```diff
@@ -0,0 +1,56 @@
...
+  class RecordInserter
+    class << self
+      delegate :instrument, :subscribe, to: ActiveSupport::Notifications
+
+      def destroy_cascade_record(record, mapper, filename)
```
jebw, line 6: All 4 methods in this class take the same 'state' params (`record`, `mapper`, `filename`) - would it make more sense to store state on the instance and use instance methods in place of class methods? eg

```ruby
RecordInserter.new(record, mapper, filename).destroy_cascade_record
```

spec/integration/cds_importer/entity_mapper_spec.rb
```diff
@@ -336,7 +336,6 @@
     end
 
     it { expect { entity_mapper.import }.to change(AdditionalCode, :count).by(1) }
-    it { expect(entity_mapper.import).to eql('AdditionalCode::Operation' => 1) }
```
jebw, line 339: This is used as part of the 'check sync is still doing something' ? Is this not needed any longer?

jebw (commented):
Initial review - looks great, just needs missing specs referred to adding

Only query is the removal of insert checking - is this being checked via the instrumentation instead? (sorry - bit behind with how all these changes are working)

app/lib/cds_importer/record_inserter.rb
willfish: I had debated an instance, yeah. Seems like this module should definitely be stateful.

spec/integration/cds_importer/entity_mapper_spec.rb
willfish: Not needed anymore, no. The result of each call of each entity mapper would be propagated to the parent via the return but we now use a subscription model to subscribe/publish insertion events

PR #610, opened 2022-05-31
HOTT-1636: Remove all search references that point to heading 8803

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1636

### What?
Remove all search references that point to heading 8803

### Why?
Heading 8803 has been removed (since December 2021) and therefore the page in the admin tool to delete these is inaccessible.


lib/tasks/search_references.rake
```diff
@@ -0,0 +1,12 @@
+namespace :search_references do
+  desc 'Remove Search references whose headings are inaccessible'
+
+  task clean: :environment do
```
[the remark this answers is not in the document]
alexdesi: yes, to tests things it was easier with a task,
also, can we run a task from a scheduled job?

I'll set up the job tomorrow, what time do we want to run this?

app/workers/clear_invalid_search_references.rb
```diff
@@ -0,0 +1,14 @@
+class ClearInvalidSearchReferences
+  include Sidekiq::Worker
+
+  sidekiq_options retry: false
+
+  def perform
+    SearchReference.where(referenced_class:'Heading').each do |ref|
```
willfish, line 7: Not just headings, btw

alexdesi: Good point, but strangely, when it is not filtered by heading, the execution of the query raises an error.
The ticket only reports the error for `Heading`, so I guess it's ok to release it only for Heading for now and then iterate to fix the other kind of entries.

I've already tested it and deleted the same nine entries mentioned in the ticket.

alexdesi: 👍 as discussed, to fix this error adding a migration to delete the search references to secions.

lib/tasks/search_references.rake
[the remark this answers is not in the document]
alexdesi: Added scheduled job at 00.00, every day.

willfish (approved):

PR #611, opened 2022-05-31
HOTT-1661: Separate cascade and missing destroy behaviour

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1661

### What?

I have added/removed/altered:

- [x] Separates cascade and missing delete behaviour

### Why?

I am doing this because:

- This is required to enable these features separately


spec/lib/cds_importer/entity_mapper/base_mapper_spec.rb
```diff
@@ -229,14 +229,6 @@ def self.name
       it { is_expected.not_to be_destroy_operation }
     end
 
-    it_behaves_like 'an xml node and mapper that are not a destroy operation', 'D' do
```
willfish, line 232: The destroy_operation?/whether the feature is enabled have been totally separated, now

alexdesi (approved):
👍

PR #616, opened 2022-06-06
HOTT-1637: Renew search references from uk -> xi on a scheduled basis

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1637

### What?

I have added/removed/altered:

- [x] Renew XI search references from UK search references table

### Why?

I am doing this because:

- This is required because only UK search references are maintained and the two services should always match


.circleci/config.yml
```diff
@@ -277,6 +277,58 @@ jobs:
           service: << parameters.service >>
           environment_key: staging
 
+  renew_xi_production_search_references:
+    docker:
+      - image: cimg/ruby:3.1.2
+    steps:
+      - checkout
+      - cf_install:
+          space: 'staging'
```
willfish, line 286: I'm only refreshing from staging for now as a test case

.circleci/config.yml
```diff
@@ -512,17 +537,24 @@ workflows:
...
-            parameters:
-              service:
-                - xi
-                - uk
+          environment_key: staging
           filters:
             branches:
               only:
                 - main
+      # - refresh_staging_db:
```
willfish, line 547: I'll reinstate this after the run at the weekend

.circleci/config.yml
```diff
@@ -260,6 +260,34 @@ jobs:
...
+            sudo apt-get update -qq
+            sudo apt-get install postgresql-client
+      - run:
+          name: "Dump the uk search_references table"
+          command: |
+            cf conduit "tariff-uk-<< parameters.environment_key >>-postgres" -- pg_dump --file tariff-uk-<< parameters.environment_key >>-postgres-search_references.psql --no-acl --no-owner --clean --verbose --table search_references
+      - run:
+          name: "Restore the uk search references into xi"
+          command: |
+            cf conduit tariff-xi-<< parameters.environment_key >>-postgres -- psql $PGDATABASE < tariff-uk-<< parameters.environment_key >>-postgres-search_references.psql
```
alexdesi, line 286: do you need `--verbose` here? I saw you used it in the previous run

.circleci/config.yml
```diff
@@ -260,6 +260,34 @@ jobs:
...
+      environment_key:
+        type: string
+    docker:
+      - image: cimg/ruby:3.1.2
+    steps:
+      - checkout
+      - tariff/cf-install:
+          space: << parameters.environment_key >>
+      - run:
+          name: "Install necessary dependencies"
```
alexdesi, line 274: name: "Install Postgress Client" 

? 🤔

alexdesi (approved):
It looks good to me ✅

.circleci/config.yml
willfish: It would just create stdout mess. My validations will include verifying that the regression suite passes and that the search references between the two services match on Monday. I'll also check the number of active connections

.circleci/config.yml
willfish: I think its clear what its doing?

PR #621, opened 2022-06-07
HOTT-1570 Allow creation of sentry release to fail

jebw:
### Jira link

[HOTT-1570](https://transformuk.atlassian.net/browse/HOTT-1570)

### What?

I have added/removed/altered:

- [x] Allow Sentry release to fail

### Why?

I am doing this because:

- We've had problems with not being able to submit release info to sentry a couple of times now, once with the CLI changing and being automatically upgraded and once with an outage on their hosted service.
- Forcing the step to always return true will prevent it from blocking releasing


alexdesi (approved):
✅

PR #624, opened 2022-06-08
HOTT-1665: Adjust off-peak rules for autoscaling production

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1665

### What?

I have added/removed/altered:

- [x] Added more intelligent autoscaling policy to enable offpeak rules that are different to onpeak (06:00 to 18:00)

### Why?

We never really scale above the minimum 8 nodes during the week unless there is an exceptional event (e.g. someone scraping the api/calling an expensive endpoint across a range of different resources - iterate over all subheadings, etc).

Given this, we should be safe to change the off-peak rules such that when the number of active users diminishes we can reduce the minimum number of active nodes by 4.

We've also noticed that at weekends the number of users is about half that during peak hours (06:00 to 18:00) of the week so we should be safe to reduce the minimum number of active nodes by 2. 

Over time this should save money.

### DIFF
--- a/config/autoscaling/production-policy.json
+++ b/config/autoscaling/production-policy.json
@@ -1,5 +1,5 @@
 {
-  "instance_min_count": 8,
+  "instance_min_count": 4,
   "instance_max_count": 12,
   "scaling_rules": [
     {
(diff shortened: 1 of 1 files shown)


willfish:
After a review of the APM metrics I can see that the api requests/scraping happens off (user) peak and so there really isn't any off-peak in the sense of the backend autoscaling 

image

PR #625, opened 2022-06-13
HOTT-1628: Fixes ordering of goods nomenclature

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1628

### What?

I have added/removed/altered:

- [x] Added compound ordering to the goods nomenclature dataset
- [x] Added coverage for this new ordering behaviour

### Why?

I am doing this because:

- This is the correct ordering


alexdesi (approved):
🚀

PR #626, opened 2022-06-13
HOTT-1628: Generate materialized paths for all goods nomenclature

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1628

### What?

I have added/removed/altered:

- [x] Adds materialized path migration to goods nomenclatures oplog table and view
- [x] Adds materialised path generation worker
- [x] Adds materialised path generation service
- [x] Adds convenience methods for pulling on chapter goods nomenclatures
- [x] Adds api for retrieving parent, ancestors, siblings, children and descendants
- [x] Adds specs for the new api

### Why?

I am doing this because:

- This is required to make for a speedier hierarchy generation and was driven by upcoming changes to search indexes

### DIFF
--- a/app/models/goods_nomenclature.rb
+++ b/app/models/goods_nomenclature.rb
@@ -41,12 +41,12 @@ class GoodsNomenclature < Sequel::Model
   end
 
   one_to_one :parent, class_name: name, class: self do |_ds|
-    parent_sid = not_heading? ? path.last : chapter.goods_nomenclature_sid
+    parent_sid = !heading? ? path.last : chapter.goods_nomenclature_sid
 
     if parent_sid.present?
       GoodsNomenclature.actual.where(goods_nomenclature_sid: parent_sid)
     else
-      GoodsNomenclature.nullify
+      GoodsNomenclature.dataset.nullify
     end
   end
 
@@ -150,11 +150,7 @@ class GoodsNomenclature < Sequel::Model
 
       return class_name unless class_name == 'Commodity'
 
-      goods_nomenclature = where(goods_nomenclature_sid:).take
-
-      'Commodity' if goods_nomenclature.declarable?
-
-      'Subheading'
+      Commodity.find(goods_nomenclature_sid:).goods_nomenclature_class
     end
   end
 
@@ -198,16 +194,8 @@ class GoodsNomenclature < Sequel::Model
     !!goods_nomenclature_item_id.match(/\A\d{4}000000\z/)
   end
 
-  def not_heading?
-    !heading?
-  end
-
   def chapter?
     !!goods_nomenclature_item_id.match(/\A\d{2}00000000\z/)
-  end
-
-  def not_chapter?
-    !chapter?
   end
 
   def declarable?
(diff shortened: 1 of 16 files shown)


app/models/goods_nomenclature.rb
```diff
@@ -103,7 +150,11 @@ def goods_nomenclature_class
 
       return class_name unless class_name == 'Commodity'
 
-      Commodity.find(goods_nomenclature_sid:).goods_nomenclature_class
+      goods_nomenclature = where(goods_nomenclature_sid:).take
+
+      'Commodity' if goods_nomenclature.declarable?
```
alexdesi, line 155: maybe return here?

app/services/materialized_path_updater_service.rb
```diff
@@ -0,0 +1,60 @@
+# Generates a materialized path for all of the goods nomenclatures under a given chapter
+#
+# We accumulate updates inside of updated_goods_nomenclatures so we can execute one SQL statement for each goods nomenclature that has their path updated without slowing ourselves down with lots of queries.
```
alexdesi, line 3: this is a long line, should we split it?

spec/models/goods_nomenclature_spec.rb
```diff
@@ -384,4 +384,180 @@
...
+
+    context 'when the goods nomenclature has a non-chapter goods nomenclature item id' do
+      subject(:goods_nomenclature) { create(:goods_nomenclature, :commodity) }
+
+      it { is_expected.not_to be_chapter }
+    end
+  end
+
+  describe '#not_chapter?' do
+    context 'when the goods nomenclature has a chapter goods nomenclature item id' do
```
alexdesi, line 525: I think the description is wrong here

spec/models/goods_nomenclature_spec.rb
```diff
@@ -384,4 +384,180 @@
...
+      subject(:goods_nomenclature) { create(:goods_nomenclature, :chapter) }
```
alexdesi, line 526: is this not the same as the previous one?

spec/models/goods_nomenclature_spec.rb
```diff
@@ -384,4 +384,180 @@
...
+
+      it { is_expected.not_to be_not_chapter }
```
alexdesi, line 528: is it expected to be a chapter? 😕

app/services/materialized_path_updater_service.rb
willfish: Oh interesting question. I don't really mind with comments but can do

alexdesi (approved):
already discussed via slack 🚀 🚀 🚀

app/models/goods_nomenclature.rb
willfish: Removed this change. Coming later

spec/models/goods_nomenclature_spec.rb
willfish: This is confusing/unnecessary. Have removed

PR #634, opened 2022-06-16
Update chief_cds_guidance.json

alexdesi:
### What?

- [x] Update chief_cds_guidance.json
- [x] also remove the date from the filenam…e and added it inside the JSON,
- [x] Also fix smoketests after an update to add basic auth to the frontend 

this way we don't need to update the hardcoded file name reference.

### Jira link
https://transformuk.atlassian.net/browse/HOTT-1684


db/chief_cds_guidance.json
```diff
@@ -1,4 +1,5 @@
 {
+    "last_update": "2022-06-16",
```
alexdesi, line 2: This is to keep track of the last update, it is not used by the code.

willfish: Maybe just remove the last_update date. We know the file changed because github tells us

alexdesi: 👍

willfish (approved):

PR #636, opened 2022-06-16
HOTT-1676 Moved ES options to SearchClient

jebw:
### Jira link

HOTT-1676

### What?

I have added/removed/altered:

- [x] Moved the global search options onto the SearchClient
- [x] Default the indexes to the namespace defined on the SearchClient

### Why?

I am doing this because:

- Previously these options lacked context for which class they impacted, -locating them on the class they relate to clarifies that

- Affects search index generation, should be run overnight on Dev first

### DIFF
--- a/app/lib/trade_tariff_backend/search_client.rb
+++ b/app/lib/trade_tariff_backend/search_client.rb
@@ -5,6 +5,9 @@ module TradeTariffBackend
     # Raised if Elasticsearch returns an error from query
     QueryError = Class.new(StandardError)
 
+    cattr_accessor :server_namespace, default: 'tariff'.freeze
+    cattr_accessor :search_operation_options, default: {}
+
     attr_reader :indexed_models,
                 :index_page_size,
                 :search_operation_options,
@@ -15,7 +18,8 @@ module TradeTariffBackend
     def initialize(search_client, options = {})
       @indexed_models = options.fetch(:indexed_models, [])
       @index_page_size = options.fetch(:index_page_size, 1000)
-      @search_operation_options = options.fetch(:search_operation_options, {})
+      @search_operation_options = options.fetch(:search_operation_options,
+                                                self.class.search_operation_options)
       @namespace = options.fetch(:namespace, 'search')
 
       super(search_client)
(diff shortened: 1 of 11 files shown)


app/lib/trade_tariff_backend/search_client.rb
```diff
@@ -5,6 +5,12 @@ class SearchClient < SimpleDelegator
     # Raised if Elasticsearch returns an error from query
     QueryError = Class.new(StandardError)
 
+    cattr_accessor :server_namespace
+    self.server_namespace = 'tariff'.freeze
+
+    cattr_accessor :search_operation_options
+    self.search_operation_options = {}
```
alexdesi, line 12: same as above

spec/lib/trade_tariff_backend/search_client_spec.rb
```diff
@@ -1,4 +1,27 @@
 RSpec.describe TradeTariffBackend::SearchClient do
+  describe '.server_namespace' do
+    subject { described_class.server_namespace }
+
+    it { is_expected.to eql 'tariff-test' }
+
+    context 'when overridden' do
+      before do
+        @orig_namespace = described_class.server_namespace
```
alexdesi, line 9: maybe we could just use let
`let(orig_namespace) { described_class.server_namespace }`
to avoid using a variable ?

alexdesi (approved):
Nice refactor!
added a couple of non-blocking comments, 
✅

spec/lib/trade_tariff_backend/search_client_spec.rb
jebw: I'm not sure that would work since the let wouldn't be called until the `after` clause?

A `let!` would work but I think those are discouraged by rubocop though that may be a better approach

jebw: Used a `let` but referenced it from the `before` to ensure its assigned prior to the value being reassigned in the `before`

app/lib/trade_tariff_backend/search_client.rb
jebw: :+1: Good call

PR #637, opened 2022-06-20
HOTT-1689 Drop feature flag from Permutations API

jebw:
### Jira link

[HOTT-1689](url)

### What?

I have added/removed/altered:

- [ ] Dropped the feature flag for the Permutations API extensions

### Why?

I am doing this because:

- This has been on for a while and is a purely additive change

### DIFF
--- a/app/services/cached_commodity_service.rb
+++ b/app/services/cached_commodity_service.rb
@@ -1,5 +1,5 @@
 class CachedCommodityService
-  CACHE_VERSION = TradeTariffBackend.permutations? ? 2 : 1
+  CACHE_VERSION = 2
 
   DEFAULT_INCLUDES = [
     'section',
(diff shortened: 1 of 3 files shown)


app/services/cached_commodity_service.rb
```diff
@@ -1,5 +1,5 @@
 class CachedCommodityService
-  CACHE_VERSION = TradeTariffBackend.permutations? ? 2 : 1
+  CACHE_VERSION = 2
```
alexdesi, line 2: maybe we could even drop this cache_version? 🤔

alexdesi (approved):
It looks good ✅ 
Just a comment about CACHE_VERSION

app/services/cached_commodity_service.rb
jebw: No - the whole point of the CACHE_VERSION is it gives us an easy way to evolve the invalidate the cache contents in the future without needing to co-ordinate manual cache invalidations together with deploys - this way a developer can just bump the CACHE_VERSION number in their PR and whenever a feature which changes the API contents gets deployed and the invalidation just happens automatically

PR #643, opened 2022-06-21
HOTT-1676 allow multiple indexes per model

jebw:
### Jira link

[HOTT-1676](https://transformuk.atlassian.net/browse/HOTT-1676)

### What?

I have added/removed/altered:

- [x] Changed SearchClient to require an explicit index when indexing a single model
- [x] Changed the ElasticSearch Sequel plugin to allow specifying one or more indexes to update from a model
- [x] Changed the BuildIndexPageWorker to operate on a index instead of a model
- [x] Generalised the mechanism for excluding an index from the Search
- [x] Allow the ES Sequel extension to silently ignore when an index does not yet exist 

### Why?

I am doing this because:

- This allows us to have multiple indexes for a given model class within the new search namespace
- Those indexes can exclude themselves from Search lookups
- This will allow us to 'evolve' the index definitions by just adding additional V2/V3/V4 indexes and then dropping the old ones in subsequent PRs - the new ones will get created automatically in the overnight sync

### Have you? (optional)

- Changes the parameters contents (but not count) of the BuildIndexPageWorker, there is still compatibility with the old syntax though so any queued jobs will still complete sucessfully.

### DIFF
--- a/app/elastic_search_indexes/search_index.rb
+++ b/app/elastic_search_indexes/search_index.rb
@@ -32,4 +32,8 @@ class SearchIndex
   def goods_nomenclature?
     false
   end
+
+  def include_in_search?
+    goods_nomenclature?
+  end
 end
--- a/app/elastic_search_indexes/search_index.rb
+++ b/app/elastic_search_indexes/search_index.rb
@@ -33,7 +33,7 @@ class SearchIndex
     false
   end
 
-  def include_in_search?
-    goods_nomenclature?
+  def exclude_from_search_results?
+    !goods_nomenclature?
   end
 end
--- a/app/elastic_search_indexes/search_index.rb
+++ b/app/elastic_search_indexes/search_index.rb
@@ -9,12 +9,16 @@ class SearchIndex
     [@server_namespace, type.pluralize].join('-')
   end
 
+  def name_without_namespace
+    self.class.name.split('::').last
+  end
+
   def type
     model_class.to_s.underscore
   end
 
   def model_class
-    self.class.name.split('::').last.chomp('Index').constantize
+    name_without_namespace.chomp('Index').constantize
   end
 
   def serializer
--- a/spec/lib/trade_tariff_backend/search_client_spec.rb
+++ b/spec/lib/trade_tariff_backend/search_client_spec.rb
@@ -50,12 +50,18 @@ RSpec.describe TradeTariffBackend::SearchClient do
   end
 
   describe '#search' do
+    before { commodity } # trigger creation and indexing prior to test results
+
     let(:commodity) do
       create :commodity, :with_description, description: 'test description'
     end
 
+    let(:index_name) do
+      "#{described_class.server_namespace}-commodities"
+    end
+
     let(:search_result) do
-      TradeTariffBackend.search_client.search q: 'test', index: TradeTariffBackend.search_index_for('search', commodity).name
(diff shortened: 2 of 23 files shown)


spec/workers/build_index_page_worker_spec.rb
```diff
@@ -1,28 +1,59 @@
...
+      let(:search_result) do
+        TradeTariffBackend.search_client.search q: 'test', index: search_index.name
+      end
+
+      it 'bulk indexes all model entries' do
+        expect(search_result.hits.total.value).to be >= 1
       end
 
+      it 'has bulk indexed the expected commodity' do
+        expect(search_result.hits.hits.map(&:_source).map(&:goods_nomenclature_item_id)).to \
```
alexdesi, line 28: we could improve readability here,
I guess we can simplify with some var with meaningful names 🤔

app/elastic_search_indexes/search_index.rb
```diff
@@ -32,4 +32,8 @@ def serialize_record(record)
   def goods_nomenclature?
     false
   end
+
+  def include_in_search?
```
alexdesi, line 36: from the name, seems that `include_in_search?  ` accepts an arg

what about `is_included_in_search?`

alexdesi (commented):
Amazing work!
I've added some comments, just to make a few things more clear. 
But logic makes sense to me.

spec/lib/trade_tariff_backend/search_client_spec.rb
```diff
@@ -46,12 +46,18 @@
   end
 
   describe '#search' do
+    before { commodity } # trigger creation and indexing prior to test results
+
     let(:commodity) do
       create :commodity, :with_description, description: 'test description'
     end
```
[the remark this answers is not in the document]
jebw: I can see the improved readability but I don't like the assumption that the first commodity is the one you'd created - likely to be correct but just assuming it seems like a step backward

spec/workers/build_index_page_worker_spec.rb
```diff
@@ -1,28 +1,59 @@
...
+      end
+
+      let(:search_index) { Search::CommodityIndex.new }
+
+      let(:search_result) do
+        TradeTariffBackend.search_client.search q: 'test', index: search_index.name
+      end
+
+      it 'bulk indexes all model entries' do
+        expect(search_result.hits.total.value).to be >= 1
```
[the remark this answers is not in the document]
jebw: TBH this is an old spec which I don't much like, I think I'll drop it - it was originally all together in a single spec - the following test actually checks both expected commodities have been indexed so I'm just going to remove it

spec/workers/build_index_page_worker_spec.rb
jebw: Yeh, good call - I'll move it into a `let` to give it a name

willfish (approved):
LGTM

alexdesi (approved):
✅

alexdesi:
There are various linking errors about the description of the contexts ?
I guess we should be consistent and start the description with "when ..."

jebw:
@alexdesi Agreed but thats historical code, not anything I'm changing and I'm keen to avoid ending up rewriting all the existing non compliant specs as part of this PR so I'll leave that for now

willfish:
Need to revert this to roll a fix to ordering out to production

PR #645, opened 2022-06-22
HOTT-1662: Adds specs for rollback behaviour

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1662

### What?

I have added/removed/altered:

- [x] Added specs for rollback behaviour for cds
- [x] Added specs for rollback behaviour for taric
- [x] Added matcher for expressive rollback testing

### Why?

I am doing this because:

- This needs to be in place before I can optimise and simplify the CDS rollbacks


alexdesi (approved):
lgtm ✅

PR #646, opened 2022-06-22
HOTT-1694 Add a new service to connect with the Tariff Search Query API.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1694

### What?
Add a new service to connect with the Tariff Search Query API.

### Why?
To be able to call tokenized, spelling correct (TODO) and other NLP features that we'll add in the future.


app/services/search_query_parser.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def initialize
+    @client = Faraday.new(
+      url: API_URL,
+      headers: {'Content-Type' => 'application/json'}
+      )
+  end
+
+  def get_tokens(search_query_term)
+    api_response = @client.get("tokens/#{search_query_term}")
```
jebw, line 12: Should this be URL escaping the `search_query_term` - or does Faraday handle that?

app/services/search_query_parser.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    response_data = api_response.success? ? JSON.parse(api_response.body) : {}
```
jebw, line 14: I think we should be handling the errors within this Parser rather then relying on the consumer of `SearchQueryParser` to in turn then parse the output hash from the Parser.

I'd suggest instead including the Faraday Errors handler - any unexpected response codes and it will trigger an exception - see

https://github.com/trade-tariff/trade-tariff-frontend/blob/main/app/services/client_builder.rb#L29

The method return value could then just be `response_data` - if `success` was false then an exception would be raised so is redundant, as is `status`

spec/services/search_query_parser_spec.rb
```diff
@@ -0,0 +1,45 @@
+RSpec.describe SearchQueryParser do
```
jebw, line 1: Spec looks great but its only testing the 'happy' path at the moment. It could do with tests for behaviour under failure conditions as well, particularly for things like api requests where network issues are out of your control.

There's some testing around failure conditions in the frontend's `ApiEntity` spec

https://github.com/trade-tariff/trade-tariff-frontend/blob/main/spec/lib/api_entity_spec.rb#L100

There's also more info on the WebMock documentation - see https://github.com/bblimke/webmock

app/services/search_query_parser.rb
```diff
@@ -0,0 +1,22 @@
+class SearchQueryParser
+  API_URL = ENV['TARIFF_QUERY_SEARCH_PARSER_URL']
+
+  def initialize
+    @client = Faraday.new(
+      url: API_URL,
+      headers: {'Content-Type' => 'application/json'}
+      )
```
jebw, line 8: I think this should be outdented by 2 spaces?

jebw (changes requested):

app/services/search_query_parser.rb
```diff
@@ -0,0 +1,22 @@
+class SearchQueryParser
+  API_URL = ENV['TARIFF_QUERY_SEARCH_PARSER_URL']
+
+  def initialize
+    @client = Faraday.new(
+      url: API_URL,
+      headers: {'Content-Type' => 'application/json'}
```
jebw, line 7: Setting the header can be handled automatically by using the `:json` Request middleware

app/services/search_query_parser.rb
jebw: I've also realised the JSON parsing can be handled automatically if you configure the Farady JSON response middleware when setting up the connection?

jebw (changes requested):

app/services/search_query_parser.rb
alexdesi: yep, I've changed the way Faraday is initialized, and we don't need it anymore.

spec/services/search_query_parser_spec.rb
alexdesi: I've added an unhappy path: when the query param is 0 it return 400 (and an error is raised by Faraday).

app/services/search_query_parser.rb
alexdesi: Done 👍

app/services/search_query_parser.rb
```diff
@@ -0,0 +1,20 @@
...
+    @client = Faraday.new(API_URL) do |conn|
+      conn.request :json
+
+      conn.response :raise_error
+      conn.response :json
+    end
+  end
+
+  def get_tokens(search_query_term)
+    api_response = @client.get('tokens', q: search_query_term)
```
jebw, line 16: Is variable assignment doing much here? Could you just chain the `.body` on the end? eg

```ruby
@client.get('tokens', q: search_query_term)
       .body
```

jebw (approved):
Looks great - definitely nicer pushing more work onto Faraday middleware :+1:

One comment on the `get_tokens` method but its more personal opinion then anything else.

PR #647, opened 2022-06-23
HOTT-1697: Removes deduplication behaviour from MeasureCollection

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1697

### What?

I have added/removed/altered:

- [x] Removes `MeasureCollection#deduplicate`
- [x] Removes references to removed method

### Why?

I am doing this because:

- The implementation is preventing us from displaying MFNs for Jersey and Guernsey
- This deduplication behaviour applied to some very old CHIEF measures and is not correct for overview measures of a heading's commodities, either


jebw (approved):
LGTM - one suggestion would be to amend the commit message to include why its being dropped - may aid any archeology for future developers (obviously its in the PR itself but if the git repo changes that will be lost)

PR #649, opened 2022-06-23
HOTT-1698: Filter in only erga omnes third country overview measures

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1698

### What?

I have added/removed/altered:

- [x] Added an additional filter on the overview measures dataset that excludes non-erga-omnes third country measures
- [x] Added coverage for the overview measure filtering

### Why?

I am doing this because:

- We need to make sure that country-specific measures aren't part of the
overview of a commodity on the subheading and heading pages in the
frontend. It doesn't make sense to show these for all users on an
overview

### DIFF
--- a/spec/factories/measure_factory.rb
+++ b/spec/factories/measure_factory.rb
@@ -155,6 +155,11 @@ FactoryBot.define do
       measure_type_id { MeasureType::THIRD_COUNTRY.sample }
     end
 
+    trait :third_country_overview do
+      geographical_area_id { GeographicalArea::ERGA_OMNES_ID }
+      third_country
+    end
+
     trait :vat do
       measure_type_id { '305' }
     end
--- a/spec/models/commodity_spec.rb
+++ b/spec/models/commodity_spec.rb
@@ -180,6 +180,66 @@ RSpec.describe Commodity do
       end
     end
 
+    describe '#overview_measures' do
+      subject(:overview_measures) { commodity.overview_measures }
+
+      let(:commodity) { create(:commodity) }
+
+      before { measure }
+
+      context 'when a third country measure that is not explicitly erga omnes' do
+        let(:measure) do
+          create(
+            :measure,
+            :with_base_regulation,
+            :third_country,
+            goods_nomenclature_sid: commodity.goods_nomenclature_sid,
+          )
+        end
+
+        it { is_expected.not_to include(measure) }
+      end
+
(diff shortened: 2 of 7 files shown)


spec/factories/measure_factory.rb
```diff
@@ -155,6 +155,11 @@
       measure_type_id { MeasureType::THIRD_COUNTRY.sample }
     end
 
+    trait :third_country_overview do
+      geographical_area_id { GeographicalArea::ERGA_OMNES_ID }
+      measure_type_id { MeasureType::THIRD_COUNTRY.sample }
```
willfish, line 160:
```suggestion
      third_country
```

spec/models/commodity_spec.rb
```diff
@@ -188,21 +248,21 @@
...
       let!(:measure1)    do
         create :measure, :with_base_regulation, measure_sid: 1,
-                         measure_type_id: measure_type.measure_type_id,
-                         additional_code_type_id: nil,
-                         goods_nomenclature_sid: commodity.goods_nomenclature_sid,
-                         validity_start_date: 1.year.ago.beginning_of_day
+                                                measure_type_id: measure_type.measure_type_id,
+                                                additional_code_type_id: nil,
+                                                goods_nomenclature_sid: commodity.goods_nomenclature_sid,
+                                                validity_start_date: 1.year.ago.beginning_of_day
```
alexdesi, line 254: I am not sure if this indentation is better than the previous one,
has Rubocop complained about it?

spec/models/commodity_spec.rb
```diff
@@ -219,14 +279,14 @@
         let(:commodity1)          { create :commodity, :with_indent }
         let(:export_measure)      do
           create :measure, :with_base_regulation, measure_type_id: export_measure_type.measure_type_id,
-                           goods_nomenclature_sid: commodity1.goods_nomenclature_sid
+                                                  goods_nomenclature_sid: commodity1.goods_nomenclature_sid
```
alexdesi, line 282: ok, maybe the indentation is to align the last params, which are hash keys

alexdesi (approved):
It looks good to me ✅ 
There are lots of indentation changes,
I guess it would be nice to have it in a different commit next time (it would make it easier to review by commits)

spec/models/commodity_spec.rb
willfish: Yeah

PR #651, opened 2022-06-23
HOTT-1687 Create releases from hotfix PRs

jebw:
### Jira link

HOTT-1687

### What?

I have added/removed/altered:

- [x] Added the ability to push urgent changes to through from PR through to production

### Why?

I am doing this because:

- It avoids the need to choose between no release, reverting changes that have not yet been QAd, or requiring urgent QA decisions. 

- Changes production deployment path

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -541,6 +541,7 @@ workflows:
             branches:
               ignore:
                 - main
+                - /^hotfix\/.+/
                 - /^dependabot\/.*/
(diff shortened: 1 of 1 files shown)


chriscoffee (approved):
Allows developers to push out time critical hotfixes in the current form of the pipeline.

I've check that the custom orb supports the ability to do this, that there is documentation and that the pipeline in this form is valid.

PR #652, opened 2022-06-24
HOTT-1698: Remove broken filtering out of ancestor mfn measures

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1698

### What?

I have added/removed/altered:

- [x] Remove broken mfn filtering

### Why?

I am doing this because:

- CHIEF never supported the hierarchy so these ancestor MFNs and VAT
measures were being filtered out. CDS do support the hierarchy so this
is actually breaking (preventing people from being able to view the mfn
measure) the overview in the frontend (see 0504 for an example where
MFNs are being hidden https://dev.trade-tariff.service.gov.uk/headings/0504).


app/services/heading_service/cached_heading_service.rb
```diff
@@ -61,19 +61,15 @@ def filter_commodity_relations
...
 
         commodity.goods_nomenclature_indents.keep_if do |ident|
           has_valid_dates(ident)
         end
 
-        indent = commodity.goods_nomenclature_indents.sort_by do |ident|
+        indent = commodity.goods_nomenclature_indents.sort_by { |ident|
           Date.parse ident.validity_start_date
-        end.last
+        }.last
```
alexdesi, line 72: This is not inline, the convention is to use do/end block right?
or you wanted to inline it, which could be fine

willfish: I used rubocop to do this, actually. It wants {} over multilines when there is a method cool on the block result :man_shrugging:

alexdesi (approved):
it looks good to me 🧹

PR #653, opened 2022-06-27
Upgrade renew search reference image version

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [ ] Update search reference image version

### Why?

I am doing this because:

- So it has a new version of the psql client which will be compatible with our psql services


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ orbs:
   cloudfoundry: circleci/cloudfoundry@1.0
   slack: circleci/slack@4.3.0
   queue: eddiewebb/queue@1.6.4
-  tariff: trade-tariff/trade-tariff-ci-orb@0 # can also change to @dev:<gitsha> for specific version or @dev:alpha to test dev branches
+  tariff: trade-tariff/trade-tariff-ci-orb@0.1.6 # can also change to @dev:<gitsha> for specific version or @dev:alpha to test dev branches
```
jebw: This isn't needed and is will remove most of the maintenance advantage of using the orb approach

willfish, line 9:
```suggestion
  tariff: trade-tariff/trade-tariff-ci-orb@0 # can also change to @dev:<gitsha> for specific version or @dev:alpha to test dev branches
```

jebw (approved):
LGTM

.circleci/config.yml
willfish: Resolved in a commit which dropped this change

PR #654, opened 2022-06-27
HOTT-1646 Add Articles to Rules of Origin

jebw:
### Jira link

[HOTT-1646](https://transformuk.atlassian.net/browse/HOTT-1646)

### What?

I have added/removed/altered:

- [x] Added the article data for UK
- [x] Added an article model + serializer
- [x] Changed the RulesOfOrigin::SchemePresenter and Serializer to include the Articles
- [x] Hopefully fixed an obscure reload bug in development mode

### Why?

I am doing this because:

- We want to include the article content within the Rules of Origin wizard on the frontend

- Changes an api that is used in production but is an append only change


app/controllers/api/v2/rules_of_origin_controller.rb
```diff
@@ -15,7 +15,7 @@ def index
 
         @serializer = Api::V2::RulesOfOrigin::SchemeSerializer.new(
           presented_schemes,
-          include: %i[links proofs rules],
+          include: %i[links proofs rules articles],
```
jebw, line 18: I debated whether to make the inclusion of articles conditionalized on a query param but figured it wasn't worth the added complexity since they're all pretty short markdown documents

willfish (approved):

PR #662, opened 2022-06-28
HOTT-1702: Adds v2 search index

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1702

### What?

I have added/removed/altered:

- [x] Added v2 search index for goods nomenclatures
- [x] Plumbed together the population of this index

### Why?

I am doing this because:

- This is required so we can do some more intelligent querying of the goods nomenclature and their ancestors and is part of a wider rollout of improvements to search

### DIFF
--- a/app/elastic_search_indexes/search/goods_nomenclature_index.rb
+++ b/app/elastic_search_indexes/search/goods_nomenclature_index.rb
@@ -1,0 +1,148 @@
+module Search
+  class GoodsNomenclatureIndex < ::SearchIndex
+    def dataset
+      TimeMachine.now do
+        GoodsNomenclature.actual
+      end
+    end
+
+    def eager_load_graph
+      %i[
+        goods_nomenclature_indents
+        chapter
+        heading
+        ancestors
+      ]
+    end
+
+    def definition
+      {
+        settings: {
+          index: {
+            analysis: {
+              analyzer: {
+                english_exact: {
+                  tokenizer: 'standard',
+                  filter: %w[lowercase],
+                },
+                english: {
+                  tokenizer: 'standard',
+                  filter: %w[
+                    english_possessive_stemmer
+                    lowercase
+                    english_stop
+                    english_stemmer
+                  ],
+                },
+              },
+              filter: {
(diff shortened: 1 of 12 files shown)


app/models/goods_nomenclature.rb
```diff
@@ -27,7 +27,11 @@ class GoodsNomenclature < Sequel::Model
   one_to_one :chapter, class_name: name, class: self do |_ds|
     Chapter
       .actual
-      .where(goods_nomenclature_item_id: chapter_id)
+      .where(goods_nomenclature_item_id: chapter_code)
+  end
+
+  one_to_one :heading, class_name: name, class: self do |_ds|
+    Heading.actual.where(goods_nomenclature_item_id: heading_code)
```
willfish, line 34: Do we want to be defensive and either nullify/return self at this point?

app/elastic_search_indexes/search/goods_nomenclature_index.rb
```diff
@@ -0,0 +1,148 @@
...
+              },
+              char_filter: {
+                standardise_quotes: {
+                  type: 'mapping',
+                  mappings: [
+                    '\\u0091=>\\u0027',
+                    '\\u0092=>\\u0027',
+                    '\\u2018=>\\u0027',
+                    '\\u2019=>\\u0027',
+                    '\\u201B=>\\u0027',
```
alexdesi, line 60: Q: what is this?

app/elastic_search_indexes/search/goods_nomenclature_index.rb
```diff
@@ -0,0 +1,148 @@
...
+                  tokenizer: 'standard',
+                  filter: %w[lowercase],
+                },
+                english: {
+                  tokenizer: 'standard',
+                  filter: %w[
+                    english_possessive_stemmer
+                    lowercase
+                    english_stop
+                    english_stemmer
```
alexdesi, line 34: my understanding is that we should use a lemmatizer instead of a stemmer, which is rough  and can include similar terms but with different meaning.

PR #663, opened 2022-06-29
HOTT-1688: Update statuses of open events

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1688

### What?

I have added/removed/altered:

- [x] Altered 'Open' events to have the correct status
- [x] Updated the specs to reflect this change

### Why?

I am doing this because:

- This is to correct the returned status in the UI for the active definition


alexdesi (approved):
already discussed by slack 👍

PR #680, opened 2022-07-13
HOTT-1743: Adds heading and chapter statistics to search results

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1743

### What?

We need to return statistics for unique headings and chapters that were returned by the beta
search functionality.

This will ultimately enable us to know which headings and chapters got the best results and also
to know which headings we can filter on in the frontend UI.

I have added/removed/altered:

- [x] Adds chapter statistics serializer and specs
- [x] Adds heading statistics serializer and specs
- [x] Updates convenience short code method and specs
- [x] Adds statistics service and specs
- [x] Updates beta search service to generate stats
- [x] Update search result serializer to include stats
- [x] Adds statistics to search result model and specs
- [x] Updates serialized search result fixture to fix specs

### Why?

I am doing this because:

- This is required for some of the beta search functionality in the frontend

### DIFF
--- a/app/models/beta/search/open_search_result.rb
+++ b/app/models/beta/search/open_search_result.rb
@@ -1,7 +1,9 @@
 module Beta
   module Search
     class OpenSearchResult
-      delegate :id, to: :search_query_parser_result, prefix: true
+      delegate :id, to: :chapter_statistics, prefix: true, allow_nil: true
+      delegate :id, to: :heading_statistics, prefix: true, allow_nil: true
+      delegate :id, to: :search_query_parser_result, prefix: true, allow_nil: true
 
       attr_accessor :took,
                     :timed_out,
@@ -26,7 +28,7 @@ module Beta
           hit = Hashie::TariffMash.new
 
           hit.score = hit_result._score
-          hit.goods_nomenclature_class = hit_result._source.goods_nomenclature_class
+          hit.goods_nomenclature_class = ActiveSupport::StringInquirer.new(hit_result._source.goods_nomenclature_class)
           hit.id = hit_result._source.id
           hit.goods_nomenclature_item_id = hit_result._source.goods_nomenclature_item_id
           hit.producline_suffix = hit_result._source.producline_suffix
@@ -58,6 +60,26 @@ module Beta
 
       def total_results
         hits.count
+      end
+
+      def chapter_statistics
+        @chapter_statistics&.values || []
+      end
+
+      def chapter_statistic_ids
+        chapter_statistics.pluck(:id)
+      end
+
+      def heading_statistics
+        @heading_statistics&.values || []
+      end
+
+      def heading_statistic_ids
+        heading_statistics.pluck(:id)
+      end
+
+      def generate_statistics
+        @chapter_statistics, @heading_statistics = Api::Beta::SearchResultStatisticsService.new(hits).call
       end
     end
   end
--- a/app/services/api/beta/search_result_statistics_service.rb
+++ b/app/services/api/beta/search_result_statistics_service.rb
@@ -3,8 +3,8 @@ module Api
     class SearchResultStatisticsService
       def initialize(goods_nomenclature_hits)
         @goods_nomenclature_hits = goods_nomenclature_hits
-        @chapter_statistics = Hashie::TariffMash.new
-        @heading_statistics = Hashie::TariffMash.new
+        @chapter_statistics = {}
+        @heading_statistics = {}
       end
 
       def call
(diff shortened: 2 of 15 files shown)


app/models/beta/search/open_search_result.rb
```diff
@@ -26,7 +28,7 @@ def build_hit(hit_result)
           hit = Hashie::TariffMash.new
 
           hit.score = hit_result._score
-          hit.goods_nomenclature_class = hit_result._source.goods_nomenclature_class
+          hit.goods_nomenclature_class = ActiveSupport::StringInquirer.new(hit_result._source.goods_nomenclature_class)
```
alexdesi, line 31: Are you using StringInquirer just to avoid to do 
hit.goods_nomenclature_class == 'chapter' ?
I don't know, I prefer using just plain ruby when the gain is not much,
but I guess it's personal style

app/services/api/beta/search_result_statistics_service.rb
```diff
@@ -0,0 +1,91 @@
...
+
+        search_result_statistics
+      end
+
+      private
+
+      attr_reader :goods_nomenclature_hits
+
+      def generate_chapter_statistics
+        goods_nomenclature_hits.each do |goods_nomenclature_hit|
```
alexdesi, line 24: I am happy with shortening a var  when something it is only inside a block, and the context is clear:

`goods_nomenclature_hits.each do |gn_hit|`

I guess this is personal preference too :)

app/services/api/beta/search_result_statistics_service.rb
```diff
@@ -0,0 +1,91 @@
...
+          @chapter_statistics[goods_nomenclature_hit.chapter_id]['avg'] ||= mean_average_chapter_score_for(goods_nomenclature_hit.chapter_id)
+        end
+
+        @chapter_statistics
+      end
+
+      def generate_heading_statistics
+        accumulations = 0
+        goods_nomenclature_hits_with_headings.each do |goods_nomenclature_hit|
+          accumulations += 1
```
alexdesi, line 41: it seems this is not used

app/services/api/beta/search_result_statistics_service.rb
```diff
@@ -0,0 +1,91 @@
...
+          @heading_statistics[goods_nomenclature_hit.heading_id]['id'] ||= goods_nomenclature_hit.heading_id
+          @heading_statistics[goods_nomenclature_hit.heading_id]['description'] ||= goods_nomenclature_hit.heading_description
+          @heading_statistics[goods_nomenclature_hit.heading_id]['chapter_id'] ||= goods_nomenclature_hit.chapter_id
+          @heading_statistics[goods_nomenclature_hit.heading_id]['chapter_description'] ||= goods_nomenclature_hit.chapter_description
+          @heading_statistics[goods_nomenclature_hit.heading_id]['score'] ||= 0
+          @heading_statistics[goods_nomenclature_hit.heading_id]['score'] += goods_nomenclature_hit.score
+          @heading_statistics[goods_nomenclature_hit.heading_id]['cnt'] ||= 0
+          @heading_statistics[goods_nomenclature_hit.heading_id]['cnt'] += 1
+          @heading_statistics[goods_nomenclature_hit.heading_id]['avg'] ||= mean_average_heading_score_for(goods_nomenclature_hit.heading_id)
+          @heading_statistics[goods_nomenclature_hit.heading_id]['chapter_score'] ||= @chapter_statistics[goods_nomenclature_hit.chapter_id]['score']
```
alexdesi, line 52: maybe we could dry out the code a little bit:

``` ruby
          heading_stat ||= Hashie::TariffMash.new
          heading_stat['id'] ||= goods_nomenclature_hit.heading_id
          heading_stat['description'] ||= goods_nomenclature_hit.heading_description
          heading_stat['chapter_id'] ||= goods_nomenclature_hit.chapter_id
          heading_stat['chapter_description'] ||= goods_nomenclature_hit.chapter_description
          heading_stat['score'] ||= 0
          heading_stat['score'] += goods_nomenclature_hit.score
          heading_stat['cnt'] ||= 0
          heading_stat['cnt'] += 1
          heading_stat['avg'] ||= mean_average_heading_score_for(goods_nomenclature_hit.heading_id)
          heading_stat['chapter_score'] ||= @chapter_statistics[goods_nomenclature_hit.chapter_id]['score']

          @heading_statistics[goods_nomenclature_hit.heading_id] = heading_stat
          heading_stat = nil
```

app/services/api/beta/search_result_statistics_service.rb
```diff
@@ -0,0 +1,91 @@
...
+          goods_nomenclature_hit.chapter_id == chapter_id
+        end
+      end
+
+      def mean_average_scores
+        scores = []
+        count = 0
+
+        goods_nomenclature_hits_with_headings.each do |goods_nomenclature_hit|
+          if yield goods_nomenclature_hit
```
alexdesi, line 81: it's not clear to me what the if is actually checking, maybe that 
```
if  yield(goods_nomenclature_hit) 
```
is not nil? (or present?)

app/services/api/beta/search_result_statistics_service.rb
```diff
@@ -0,0 +1,91 @@
...
+            scores << goods_nomenclature_hit.score
+            count += 1
+          end
+        end
+
+        scores.inject(0, :+) / count
```
alexdesi, line 87: is not count equal to scores.size ? 🤔

app/services/api/beta/search_result_statistics_service.rb
willfish: Oh, this was me debugging a weird issue with Mashie. I'll remove it

app/services/api/beta/search_result_statistics_service.rb
willfish: Could do, yeah

app/services/api/beta/search_result_statistics_service.rb
willfish: Well, the result of yield is always a boolean. So equivalent would be
```ruby
 if (yield goods_nomenclature_hit == true)
```

app/services/api/beta/search_result_statistics_service.rb
willfish: We could get rid of that, yeah. I guess its more about readability at this point as adding 1 to the count each time doesn't really costs us anything

```ruby
scores.inject(0, :+) / scores.size
```

app/models/beta/search/open_search_result.rb
willfish: I basically added it as I found it more expressive/could see uses for it later :shrug: . I guess we do do this in a couple of places in the backend (quota search and in some tests)

alexdesi (approved):
✅

PR #681, opened 2022-07-14
HOTT-1755: Full-text ancestor search

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1755

### What?

We need to be able to search the ancestor tree for better search recall (especially where the leaf node has a description of 'Other').

We also need the search indexes to be more efficient - not inflexively creating inverted-/keyword- indexes when they're not needed.

I have added/removed/altered:

- [x] Removes broken exact search - removing unused inverted/keyword indexes
- [x] Adds full-text-searchable ancestor tree
- [x] Adds support for ancestor tree search to goods_nomenclature_query
- [x] Adds support for ancestor tree search to serializer
- [x] Adds support for ancestor tree to search service

### Why?

I am doing this because:

- Full-text-search with multi-match won't work with nested documents
- We were arbitrarily creating keyword- and inverted- indexes when we do not need them


app/serializers/search/goods_nomenclature_serializer.rb
```diff
@@ -79,5 +90,57 @@ def validity_start_date
...
+
+    def ancestor_11_description_indexed
+      ancestors[10].try(:[], :description_indexed)
+    end
+
+    def ancestor_12_description_indexed
+      ancestors[11].try(:[], :description_indexed)
+    end
+
+    def ancestor_13_description_indexed
```
willfish, line 142: This number comes from the largest number of ancestors of any commodity in the database

The reality is that the inverted index for this particular schema property will be very sparsely populated

app/serializers/search/goods_nomenclature_serializer.rb
```diff
@@ -79,5 +90,57 @@ def validity_start_date
...
+      ancestors[12].try(:[], :description_indexed)
+    end
```
alexdesi, line 144: I hate to say it, but this could be a case where
`define_method`
could be beneficial :)

willfish: The thing is, it doesn't actually add much - in the sense that I have to manually know the largest number of ancestors and statically define them in the index anyway. I'd say this isn't a good use of that. I'd probably use define_method if I dynamically knew something

willfish: I think it doesn't add much but I can do

alexdesi (approved):
lgtm ✅

PR #686, opened 2022-07-18
HOTT-1763: Adds synonym config file for opensearch beta index

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1763

### What?

I have added/removed/altered:

- [x] Added synonym config file
- [x] Added docker-compose volume mount
- [x] Added working example TODO to index for consuming file

### Why?

I am doing this because:

- This will soon be plumbed together to pull config files into AWS as part of our deployment pipeline

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -45,6 +45,7 @@ services:
       test: curl -s http://localhost:9200/_cluster/health | grep -vq '"status":"red"'
     volumes:
     - hmrc-os:/usr/share/opensearch/data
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -45,6 +45,7 @@ services:
       test: curl -s http://localhost:9200/_cluster/health | grep -vq '"status":"red"'
     volumes:
     - hmrc-os:/usr/share/opensearch/data
+    - ./config/opensearch/synonyms_generic.txt:/usr/share/opensearch/config/synonyms_generic.txt
```
willfish, line 48: I've populated and tested search using these synonyms and they work

PR #692, opened 2022-07-20
Removed references to top in originprocesses.md for each country

maverickvi:
### Jira link

[HOTT-1716](https://transformuk.atlassian.net/browse/HOTT-1716)

### What?

I have added/removed/altered:

- [ ] Removed references to top in originprocesses.md for each country

### Why?

I am doing this because:

- We will have more freedom when placing a go to top link on our front end and our API will return a cleaner response meaning API users won't have to deal with references to {{ top }}


alexdesi (approved):
It looks good to me ✅

PR #693, opened 2022-07-21
Hott 1701 feature flag new rules

jebw:
### Jira link

HOTT-1701

### What?

I have added/removed/altered:

- [x] Added a feature flag to hide the new API data
- [x] Removed the `original` field from the v2 rules

### Why?

I am doing this because:

- The API is still evolving so should be hidden until we're happy with it
- The original field is not currently needed and so keep the API to the minimum for now

### Have you? (optional)

- We've not yet deployed this api so it should not affect anything


jebw:
Closing because it was decided to not feature flag this

PR #694, opened 2022-07-21
HOTT-1768: Persist facet classifications

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1768

### What?

I have added/removed/altered:

- [x] Add namespace to facet configuration model
- [x] Adds keyword filter properties to beta search index
- [x] Adds heading_facet_mappings.csv file
- [x] Adds phrase token generator service
- [x] Adds default filters to goods_nomenclature_query
- [x] Handle heading short code for chapters
- [x] Updates converter service to include heading mappings
- [x] Adds facet classification model to encapsulate classifications
- [x] Updates ES serializer to persist classifications
- [x] Adds comment to token generator

### Why?

I am doing this because:

- This is required so we can dynamically query facet category classifications


app/models/beta/search/goods_nomenclature_query.rb
```diff
@@ -22,20 +26,22 @@ class GoodsNomenclatureQuery
...
         query.adjectives = search_query_parser_result.adjectives
+        query.filters = filters
 
         query
       end
 
       def query
         candidate_query = { query: { bool: {} } }
 
+        candidate_query[:query][:bool][:filter] = filter_part if filters.any?
```
willfish, line 44: This implementation is coming in the next PR

app/models/beta/search/facet_classification.rb
```diff
@@ -0,0 +1,64 @@
...
+              matching_facet_classifications = word_classifications[token] || {}
+
+              matching_facet_classifications.slice(*applicable_facet_classifiers).each do |facet, classification|
+                classifications[facet] ||= {}
+
+                # We always want the most precise classifications
+                #
+                # In practice this means we bail extending classifications if the lowest goods_nomenclature
+                # node in the tree has found any for the given facet category
+                next if classifications[facet].except(gn.goods_nomenclature_sid).any?
```
alexdesi, line 24: I guess we could create a private method with a meaningful name for this?

def  classification_for_facet_not_empty?  (I actually don't like much this name 🤔 )
  classifications[facet].except(gn.goods_nomenclature_sid).any?
end

app/models/beta/search/facet_classification.rb
```diff
@@ -0,0 +1,64 @@
...
+
+                classifications[facet][gn.goods_nomenclature_sid] ||= Set.new
```
alexdesi, line 26: good to use standard API like set 👍

app/models/beta/search/facet_classification.rb
```diff
@@ -0,0 +1,64 @@
...
+          end
+
+          # Facet category classifications can only belong to one goods_nomenclature
+          classifications = classifications.transform_values(&:values).transform_values(&:first)
+
+          facet_classification = new
+
+          facet_classification.classifications = classifications
+
+          facet_classification
```
alexdesi, line 39: maybe removing the reassignment? 

```ruby
facet_classification = new
facet_classification.classifications = classifications.transform_values(&:values).transform_values(&:first)
facet_classification
```

app/models/beta/search/facet_classification.rb
```diff
@@ -0,0 +1,64 @@
...
+                next if classifications[facet].except(gn.goods_nomenclature_sid).any?
+
+                classifications[facet][gn.goods_nomenclature_sid] ||= Set.new
+                classifications[facet][gn.goods_nomenclature_sid] << classification
+              end
+            end
+          end
+
+          # Facet category classifications can only belong to one goods_nomenclature
+          classifications = classifications.transform_values(&:values).transform_values(&:first)
```
alexdesi, line 33: maybe we can just use one `transform_values` instead of two in cascade?

app/models/beta/search/goods_nomenclature_query.rb
```diff
@@ -22,20 +26,22 @@ class GoodsNomenclatureQuery
...
         candidate_query[:query][:bool][:must] = must_part if must_part.any?
         candidate_query[:query][:bool][:should] = should_part if should_part.any?
```
alexdesi, line 46: do we need the ifs here?
maybe it's fine to have 

`candidate_query[:query][:bool][:filter] = nil`

same for must and should

alexdesi (commented):
added few minor comments

app/models/beta/search/goods_nomenclature_query.rb
willfish: This is actually something required by the ES api

alexdesi (approved):
lgtm ✅

PR #698, opened 2022-07-25
HOTT-1774: Adds spec and handling for missing memberships

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1774

### What?

I have added/removed/altered:

- [x] Added handling for missing geographicalAreaMembership nodes in mapper model callback
- [x] Moved all specs relating to the geographicalAreaMembership mapper callbacks into the mapper spec file
- [x] Added coverage for missing geographicalAreaMembership nodes

### Why?

I am doing this because:

- This is required because the existing geographicalAreaMembership import process is broken as of the file tariff_dailyExtract_v1_20220721T235959.gzip
- We shouldn't explode in these scenarios (current behaviour) and should still be able to import quotas, etc.


alexdesi (approved):
✅

PR #704, opened 2022-07-27
HOTT-1741: Enable support for static and dynamic search filters

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1741

### What?

This PR enables dynamic es filtering based on supported keyword indexes (mainly for facets but also we support a bunch of other static properties, too).

It enables calling the beta search api like this:

```
GET /api/beta/search?q=horses&filter[animal_state]=live&filter[goods_nomenclature_class]=Commodity
```

Which can be translated to fetch all goods nomenclature that multi match the query 'horses' and return only hits that have the `live` animal state and which are a Commodity 

I have added/removed/altered:

- [x] Adds filter generator service and specs
- [x] Enable filter passing to the beta search controller
- [x] Adds search configuration helpers to enable shared configuration
- [x] Update configuration initializer to preload all filters
- [x] Updates search service and specs to pass filters
- [x] Promote heading_id and chapter_id to have keyword indexes
- [x] Adds dynamic and static filters to goods_nomenclature_query

### Why?

I am doing this because:

- This is required so users can further refine their search results based on dynamic facet filters and static filters for all of the available keyword-indexed properties.


app/controllers/api/beta/search_controller.rb
```diff
@@ -7,7 +7,6 @@ class SearchController < ApiController
         :heading_statistics,
         :chapter_statistics,
         :guide,
-        'facet_filter_statistics',
```
willfish, line 10: This actually isn't required/is implicit when we added `'facet_filter_statistics.facet_classification_statistics'` as an include

app/elastic_search_indexes/search/goods_nomenclature_index.rb
```diff
@@ -83,6 +83,8 @@ def definition
             goods_nomenclature_class: { type: 'keyword' },
             goods_nomenclature_item_id: { type: 'text' },
             producline_suffix: { type: 'keyword' },
+            chapter_id: { type: 'keyword' },
```
willfish, line 86: Since its a requirement to filter these now we need to make sure they are backed by keyword indexes

javier-transform (approved):

PR #717, opened 2022-08-02
HOTT-1759 - Permit intercept messages to be used on search terms

javier-transform:
### Jira link

[HOTT-1759](https://transformuk.atlassian.net/browse/HOTT-1759)

### What?

I have added/removed/altered:
 - [x] A Model named `intercept_messages`
 - [x]  A serializer named `intercept_messages_serializer.rb`
 - [x] Added test cases for this scenario
 - [x] Populated en.yml with valid keys to test the messages
### Why?

I am doing this because:

- To cater ticket requirements

### Have you? (optional)

### DIFF
--- a/app/models/beta/search/intercept_message.rb
+++ b/app/models/beta/search/intercept_message.rb
@@ -12,8 +12,8 @@ module Beta
 
         result = new
 
-        result.term = (I18n.t "#{search_query}.title")
-        result.message = (I18n.t "#{search_query}.message")
+        result.term = I18n.t "#{search_query}.title"
+        result.message = I18n.t "#{search_query}.message"
 
         result
       end
--- a/app/models/beta/search/intercept_message.rb
+++ b/app/models/beta/search/intercept_message.rb
@@ -8,12 +8,14 @@ module Beta
       attr_accessor :intercept_message, :term, :message
 
       def self.build(search_query)
-        return unless search_query.eql?((I18n.t "#{search_query}.title"))
+        query = search_query.downcase
+
+        return unless query.eql?((I18n.t "#{query}.title"))
 
         result = new
 
-        result.term = I18n.t "#{search_query}.title"
-        result.message = I18n.t "#{search_query}.message"
+        result.term = I18n.t "#{query}.title"
+        result.message = I18n.t "#{query}.message"
 
         result
       end
--- a/app/models/beta/search/intercept_message.rb
+++ b/app/models/beta/search/intercept_message.rb
@@ -1,0 +1,22 @@
+module Beta
+  module Search
+    class InterceptMessage
+      include ContentAddressableId
+
+      content_addressable_fields :intercept_message
+
+      attr_accessor :intercept_message, :term, :message
+
+      def self.build(search_query)
+        return unless search_query.eql?((I18n.t "#{search_query}.title"))
+
+        result = new
+
+        result.term = (I18n.t "#{search_query}.title")
+        result.message = (I18n.t "#{search_query}.message")
+
+        result
+      end
+    end
+  end
+end
--- a/app/models/beta/search/intercept_message.rb
+++ b/app/models/beta/search/intercept_message.rb
@@ -3,19 +3,19 @@ module Beta
     class InterceptMessage
       include ContentAddressableId
 
-      content_addressable_fields :intercept_message
+      content_addressable_fields :term, :message
 
-      attr_accessor :intercept_message, :term, :message
+      attr_accessor :term, :message
 
       def self.build(search_query)
         query = search_query.downcase
 
-        return unless query.eql?((I18n.t "#{query}.title"))
+        return unless query.eql?(I18n.t("#{query}.title"))
 
         result = new
 
-        result.term = I18n.t "#{query}.title"
-        result.message = I18n.t "#{query}.message"
+        result.term = I18n.t("#{query}.title")
+        result.message = I18n.t("#{query}.message")
 
         result
       end
--- a/app/models/beta/search/open_search_result.rb
+++ b/app/models/beta/search/open_search_result.rb
@@ -141,6 +141,8 @@ module Beta
       def intercept_message
         searched_term = search_query_parser_result.corrected_search_query
 
+        return unless searched_term.eql?((I18n.t "#{searched_term}.title"))
+
         {
           term: (I18n.t "#{searched_term}.title"),
           message: (I18n.t "#{searched_term}.message"),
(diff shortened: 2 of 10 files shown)


spec/models/beta/search/open_search_result_spec.rb
```diff
@@ -266,6 +266,15 @@
...
 
+  describe '#intercept_message' do
+    subject(:result) { build(:search_result) }
+
+    it 'will match the object in the yaml file' do
+      searched_term = result.search_query_parser_result.corrected_search_query
+      expect(searched_term).to eq((I18n.t "#{searched_term}.title"))
+    end
+  end
+
```
alexdesi, line 277: should we add the context where searched_term is not present?
Is it a common scenario, which we can handle returning nil
or is it a rare scenario, which should be handled by exception? 🤔

javier-transform: If the searched term is not present then we currently get the message and title that it does not exist in the `yaml` file. But we can tweak it as you've mentioned and return nil instead, if it doesn't exist over there. I'll modify this edge case and ping you shortly :)

app/models/beta/search/open_search_result.rb
```diff
@@ -137,6 +137,17 @@ def redirect!
...
         @redirect
       end
+
+      def intercept_message
+        searched_term = search_query_parser_result.corrected_search_query
+
+        return unless searched_term.eql?((I18n.t "#{searched_term}.title"))
+
+        {
+          term: (I18n.t "#{searched_term}.title"),
```
willfish, line 147: I don't think you need to change the evaluation rules with these extra (). Essentially `I18n.t "#{searched_term}.title"` will return the thing you want.

willfish:
So I'm wondering about a few things architecturally:

- Do we want to store the intercept messages in a separate table and have them managed via some initial seeds. I guess my thoughts on why this is desirable is because this would: 
  - be consistent with what we do for search references
  - give you the model you want for serializing InterceptMessage instances
  - of all of the functionality we've delivered around search, these are the most intriguing to HMRC and could help get stakeholders managing other aspects of the admin UI
  - our users are going to potentially have thousands of these that they want to manage
- Are we 100% sure it makes sense actually doing a search when we want to give people specific guidance on a search term? even the name InterceptMessage makes me think we've stopped something from happening

app/models/beta/search/intercept_message.rb
```diff
@@ -0,0 +1,22 @@
+module Beta
+  module Search
+    class InterceptMessage
+      include ContentAddressableId
+
+      content_addressable_fields :intercept_message
+
+      attr_accessor :intercept_message, :term, :message
```
willfish, line 8:
```suggestion
      attr_accessor :term, :message
```

app/models/beta/search/intercept_message.rb
```diff
@@ -0,0 +1,22 @@
+module Beta
+  module Search
+    class InterceptMessage
+      include ContentAddressableId
+
+      content_addressable_fields :intercept_message
```
willfish, line 6:
```suggestion
      content_addressable_fields :term, :message
```

spec/models/beta/search/open_search_result_spec.rb
```diff
@@ -266,6 +266,20 @@
     it { expect { result.redirect! }.to change(result, :redirect?).from(nil).to(true) }
   end
 
+  describe '#intercept_message' do
+    subject(:result) { build(:search_result) }
+
+    it 'will match the object in the yaml file' do
+      searched_term = result.search_query_parser_result.corrected_search_query
+      expect(searched_term).to eq((I18n.t "#{searched_term}.title"))
```
willfish, line 274:
```suggestion
      expect(searched_term).to eq(I18n.t "#{searched_term}.title")
```

spec/models/beta/search/open_search_result_spec.rb
```diff
@@ -266,6 +266,20 @@
...
+    end
+
+    it 'will not match the object in the yaml file' do
+      searched_term = 'random_string'
+      expect(searched_term).not_to eq((I18n.t "#{searched_term}.title"))
```
willfish, line 279:
```suggestion
      expect(searched_term).not_to eq(I18n.t "#{searched_term}.title")
```

app/models/beta/search/intercept_message.rb
```diff
@@ -0,0 +1,22 @@
...
+
+      attr_accessor :intercept_message, :term, :message
+
+      def self.build(search_query)
+        return unless search_query.eql?((I18n.t "#{search_query}.title"))
```
willfish, line 11: Should we downcase the input search query?

willfish (approved):
LGTM. I do think we'll want to downcase the search query when we build the intercept message

willfish:
For anyone coming to this PR in the future, we ended up deciding to deliver without the database persistence/admin api for now as we have other things to unpick with GDS SSO in the admin area that are prerequisites to wider adoption of that application

app/models/beta/search/intercept_message.rb
javier-transform: We need the :intercept_message field here otherwise we get this error from the concern 
image

spec/models/beta/search/open_search_result_spec.rb
javier-transform: Yea, i didn't want to add these braces too but apparently rubocop picks it up as an offence so that's why i added them.
image

app/models/beta/search/intercept_message.rb
willfish: :eyes:

spec/models/beta/search/open_search_result_spec.rb
willfish: So it's probably complaining about I18n.t("#{searched_term}.title")

willfish: ****
```suggestion
      expect(searched_term).to eq(I18n.t("#{searched_term}.title"))
```

spec/models/beta/search/open_search_result_spec.rb
willfish:
```suggestion
      expect(searched_term).not_to eq(I18n.t("#{searched_term}.title"))
```

app/models/beta/search/intercept_message.rb
javier-transform: You're right, i missed that. We do need to do that. Updating it with a short commit

PR #722, opened 2022-08-03
HOTT-1831: Exclude presentational headings from search

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1831

### What?

I have added/removed/altered:

- [x] Exclude presentational headings from goods nomenclature index
- [x] Adds spec to show query skips grouping headings

### Why?

I am doing this because:

- This is all required so we do not add grouping headings (which are only presentational) to the search results

### DIFF
--- a/app/elastic_search_indexes/search/goods_nomenclature_index.rb
+++ b/app/elastic_search_indexes/search/goods_nomenclature_index.rb
@@ -2,13 +2,10 @@ module Search
   class GoodsNomenclatureIndex < ::SearchIndex
     def dataset
       TimeMachine.now do
-        GoodsNomenclature.actual
+        GoodsNomenclature.actual.exclude(
+          "goods_nomenclatures.goods_nomenclature_item_id LIKE '____000000' AND producline_suffix != '80'",
+        )
       end
-    end
-
-    # Skips headings that are only used for display purposes
-    def skip?(goods_nomenclature)
-      goods_nomenclature.heading_grouping?
     end
 
     def eager_load_graph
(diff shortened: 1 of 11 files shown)


app/elastic_search_indexes/search/goods_nomenclature_index.rb
```diff
@@ -6,6 +6,13 @@ def dataset
       end
     end
 
+    # Headings are grouped by other grouping headings and we do not want
+    # these grouping headings in the search results since they're mainly for
+    # display purposes.
```
alexdesi, line 11: can we simplify the comment?

app/workers/build_index_page_worker.rb
```diff
@@ -21,6 +21,8 @@ def perform(index_namespace, index_name, page_number, page_size)
 
   def serialize_for(operation, index, entries)
     entries.each_with_object([]) do |model, memo|
+      next if index.skip?(model)
+
```
alexdesi, line 25: I was thinking we could use a "Tell Don’t Ask" approach.

instead of asking what we should skip,
we could tell what we want:

```ruby
get_ungrupped_headings(entries) do |model, memo| 
  # same stuff as before
end

def get_ungrupped_headings # I am sure you can find a better name :) 
   # filter out the grupped ones
end
```

app/elastic_search_indexes/search/goods_nomenclature_index.rb
willfish: Please add a suggestion :)

app/workers/build_index_page_worker.rb
willfish: So I'd need the dataset to do a union on headings that we want plus everything else. In this case, asking is actually easier

alexdesi (approved):
✅

PR #724, opened 2022-08-05
HOTT-1759 - Adds a migration to fix description for a couple of comm codes

javier-transform:
### Jira link

[HOTT-1795](https://transformuk.atlassian.net/browse/HOTT-1795)

### What?

I have added/removed/altered:

- [x] Added migration to fix com codes for the following ids (154585, 157423)

### DIFF
--- a/db/data_migrations/2022080530154709_fix_broken_comm_code_descriptions.rb
+++ b/db/data_migrations/2022080530154709_fix_broken_comm_code_descriptions.rb
@@ -1,8 +1,0 @@
-Sequel.migration do
-  up do
-    Sequel::Model.db[:goods_nomenclature_descriptions_oplog].where(id: 154_585).update(description: 'Tunas (of the genus Thunnus), skipjack tuna (stripe-bellied bonito) (Katsuwonus pelamis), excluding edible fish offal of subheadings 0302 91 to 0302 99')
-    Sequel::Model.db[:goods_nomenclature_descriptions_oplog].where(id: 157_423).update(description: 'Four-stroke petrol engines of a cylinder capacity of not more than 250 cm3 for use in the manufacture of garden equipment of heading 8432, 8433, 8436 or 8508')
-  end
-
-  down {}
-end
--- a/db/data_migrations/2022080530154709_fix_broken_comm_code_descriptions.rb
+++ b/db/data_migrations/2022080530154709_fix_broken_comm_code_descriptions.rb
@@ -1,9 +1,8 @@
 Sequel.migration do
+  up do
+    Sequel::Model.db[:goods_nomenclature_descriptions_oplog].where(id: 154_585).update(description: 'Tunas (of the genus Thunnus), skipjack tuna (stripe-bellied bonito) (Katsuwonus pelamis), excluding edible fish offal of subheadings 0302 91 to 0302 99')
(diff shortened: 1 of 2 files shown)


db/data_migrations/2022080530154709_fix_broken_comm_code_descriptions.rb
```diff
@@ -0,0 +1,8 @@
+Sequel.migration do
+  up do
+    Sequel::Model.db[:goods_nomenclature_descriptions_oplog].where(id: 154_585).update(description: 'Tunas (of the genus Thunnus), skipjack tuna (stripe-bellied bonito) (Katsuwonus pelamis), excluding edible fish offal of subheadings 0302 91 to 0302 99')
```
willfish, line 3: I think you'll want the primary key (not labelled id) which is goods_nomenclature_description_period_sid.

Probably better to always use the table as configured by the model, too:

```ruby
description = GoodsNomenclatureDescription.where(goods_nomenclature_description_period_sid: 154_585)
description.update(description: 'Tunas (of the genus Thunnus), skipjack tuna (stripe-bellied bonito) (Katsuwonus pelamis), excluding edible fish offal of subheadings 0302 91 to 0302 99')
```

willfish: If you really want to update the full history of descriptions that match that description sid, then you'll want to use the Operation class:

```ruby
description = GoodsNomenclatureDescription::Operation.where(goods_nomenclature_description_period_sid: 154_585)
description.update(description: 'Tunas (of the genus Thunnus), skipjack tuna (stripe-bellied bonito) (Katsuwonus pelamis), excluding edible fish offal of subheadings 0302 91 to 0302 99')
```

javier-transform: Don't think we need the operation class for this purpose

db/data_migrations/20220805304709_fix_broken_comm_code_descriptions.rb
```diff
@@ -0,0 +1,10 @@
+Sequel.migration do
+  up do
+    tunas = GoodsNomenclatureDescription.where(goods_nomenclature_description_period_sid: 154_585)
+    tunas.update(description: 'Tunas (of the genus Thunnus), skipjack tuna (stripe-bellied bonito) (Katsuwonus pelamis), excluding edible fish offal of subheadings 0302 91 to 0302 99')
```
alexdesi, line 4: this should be fine,
please could you double-check that there is not any issue in case `tunas` is empty?

alexdesi (approved):
✅ it looks fine, but before merging it check that it still works even when the were returns nothing

db/data_migrations/20220805304709_fix_broken_comm_code_descriptions.rb
```diff
@@ -0,0 +1,10 @@
...
+    engines = GoodsNomenclatureDescription.where(goods_nomenclature_description_period_sid: 157_423)
+    engines.update(description: 'Four-stroke petrol engines of a cylinder capacity of not more than 250 cm3 for use in the manufacture of garden equipment of heading 8432, 8433, 8436 or 8508')
```
alexdesi, line 6: and here

PR #727, opened 2022-08-08
HOTT-1833 adds endline and declarable status to goods nomen clature

javier-transform:
### Jira link

[HOTT-1833](https://transformuk.atlassian.net/browse/HOTT-1833)

### What?

I have added/removed/altered:

- [x] adds endline and declarable status to goods nomen clature serializer

### DIFF
--- a/spec/serializers/api/beta/commodity_serializer_spec.rb
+++ b/spec/serializers/api/beta/commodity_serializer_spec.rb
@@ -67,6 +67,8 @@ RSpec.describe Api::Beta::CommoditySerializer do
             search_references: '',
             validity_start_date: '1972-01-01T00:00:00Z',
             validity_end_date: nil,
+            end_line: nil,
+            declarable?: false,
             chapter_id: '51',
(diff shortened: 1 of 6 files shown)


spec/serializers/api/beta/commodity_serializer_spec.rb
```diff
@@ -67,6 +67,8 @@
             search_references: '',
             validity_start_date: '1972-01-01T00:00:00Z',
             validity_end_date: nil,
+            end_line: nil,
+            declarable?: false,
```
alexdesi, line 71: I think this should not always be false.
of course, there are declarable commodities ... or am I missing something? 🤔

javier-transform: Yes it's not always false, here is the method that checks for declarable status image

And we've populated the test cases for the declarable status in the goods_nomenclature_spec.rb here
image

alexdesi (approved):
it looks good ✅

PR #733, opened 2022-08-15
HOTT-1870: Expose Origin reference document attributes on ROO API

maverickvi:
### Jira link

[HOTT-1870](https://transformuk.atlassian.net/browse/HOTT-1870)

### What?

I have added/removed/altered:

- [ ] Exposed Origin reference document attributes on Rules of origin API

### Why?

I am doing this because:

- We need to access Origin reference document attributes on the front end

### Have you? (optional)

### DIFF
--- a/app/models/rules_of_origin/origin_reference_document.rb
+++ b/app/models/rules_of_origin/origin_reference_document.rb
@@ -1,14 +1,13 @@
-# frozen_string_literal: true
-
 module RulesOfOrigin
   class OriginReferenceDocument
     include ActiveModel::Model
+    include ContentAddressableId
 
     attr_accessor :ord_title, :ord_version, :ord_date, :ord_original
-    attr_writer :id
 
-    def id
-      @id = 'origin_reference_document_id'
-    end
+    content_addressable_fields :ord_title,
+                               :ord_version,
+                               :ord_date,
+                               :ord_original
   end
 end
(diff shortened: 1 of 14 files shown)


javier-transform (approved):
LGTM

app/models/rules_of_origin/origin_reference_document.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
```
willfish, line 1: You shouldn't need this - I suspect the wrong local rubocop config is being used

app/models/rules_of_origin/origin_reference_document.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module RulesOfOrigin
+  class OriginReferenceDocument
+    include ActiveModel::Model
+
+    attr_accessor :ord_title, :ord_version, :ord_date, :ord_original
+    attr_writer :id
+
+    def id
+      @id = 'origin_reference_document_id'
```
willfish, line 11: I suspect you want to use the ContentAddressableId module and specify the fields that make this entity unique

app/models/rules_of_origin/origin_reference_document.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module RulesOfOrigin
+  class OriginReferenceDocument
+    include ActiveModel::Model
+
+    attr_accessor :ord_title, :ord_version, :ord_date, :ord_original
+    attr_writer :id
```
willfish, line 8: You don't want this

willfish (approved):

PR #753, opened 2022-08-26
HOTT-1928: Handle rollover in patched taric downloader

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1928

### What?

The initial (and correct) version of the taric file downloader would determine what file to download for a given date (we call this the issue date) by calling an api with the current days date string.

Sadly this date → file api started giving false negative responses (404s). Calling the file api with the expected next sequence of filename worked but asking for the filename on a given issue date didn’t meaning the ETL process for taric was just blocked.

We fixed this by just checking the next sequence we can expect against the filename api using a patched version of the downloader module using a simple factory pattern.

The patched downloader is not aware of the standard of resetting the sequence when the year changes and sadly we don’t have a 1->1 connection between the issue date and the filename so are unable to predict based on the issue date what the next file is going to be when the year changes.

I have added/removed/altered:

- [x] Adds next rollover update method
- [x] Consume new taric rollover method

### Why?

I am doing this because:

- This patched downloader appears to be the downloader we're going to use after the new year and it would stop working if we didn't add this

### DIFF
--- a/app/lib/tariff_synchronizer/taric_update.rb
+++ b/app/lib/tariff_synchronizer/taric_update.rb
@@ -94,13 +94,31 @@ module TariffSynchronizer
       )
     end
 
+    def next_rollover_update
+      self.class.new(
+        filename: next_update_sequence_rollover_update_filename,
+        issue_date: next_update_rollover_issue_date,
+      )
+    end
+
     def next_update_sequence_update_filename
       "#{next_update_issue_date.iso8601}_#{next_update_sequence_url_filename}"
+    end
+
+    def next_update_sequence_rollover_update_filename
+      "#{next_update_rollover_issue_date.iso8601}_#{next_update_sequence_rollover_url_filename}"
     end
 
     def next_update_sequence_url_filename
       padded_sequence = next_update_sequence.to_s.rjust(3, '0')
       year = next_update_issue_date.strftime('%y')
+
+      "TGB#{year}#{padded_sequence}.xml"
+    end
+
+    def next_update_sequence_rollover_url_filename
+      padded_sequence = NEW_YEAR_STARTING_SEQUENCE_NUMBER.to_s.rjust(3, '0')
+      year = next_update_rollover_issue_date.strftime('%y')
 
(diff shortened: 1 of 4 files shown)


spec/unit/tariff_synchronizer/taric_update_spec.rb
```diff
@@ -173,6 +173,20 @@
     end
   end
 
+  describe '#next_update' do
+    subject(:next_update) { build(:taric_update, :applied, example_date: Date.parse('2022-08-25')).next_update }
+
+    it { is_expected.to have_attributes(issue_date: Time.zone.today, filename: '2022-08-26_TGB22238.xml') }
```
alexdesi, line 179: Is this test going to pass tomorrow 🤔 ?
(maybe, I just don't trust much test with variable time)

app/lib/tariff_synchronizer/taric_update.rb
```diff
@@ -94,17 +94,35 @@ def next_update
...
     def next_update_sequence_url_filename
       padded_sequence = next_update_sequence.to_s.rjust(3, '0')
       year = next_update_issue_date.strftime('%y')
 
       "TGB#{year}#{padded_sequence}.xml"
     end
 
+    def next_update_sequence_rollover_url_filename
+      padded_sequence = NEW_YEAR_STARTING_SEQUENCE_NUMBER.to_s.rjust(3, '0')
+      year = next_update_rollover_issue_date.strftime('%y')
```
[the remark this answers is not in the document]
willfish: This doesn't have the same effect:
```ruby
[1] pry(main)> Time.zone.today.year
=> 2022
[2] pry(main)> Time.zone.today.strftime('%y')
=> "22"

```

spec/unit/tariff_synchronizer/taric_update_spec.rb
willfish: It will now, yah

alexdesi (approved):
🚢 ✅

PR #756, opened 2022-08-30
HOTT-1918: Support groups in excluded geographical areas

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1918

### What?

I have added/removed/altered:

- [x] Add support for handling groups in measure geographical area exclusions
- [x] Adds spec coverage for new behaviour

### Why?

I am doing this because:

- A customer noticed that this wasn't supported in the tariff
- On a side note we're not actually setting the group correctly for the EU in the measure excluded geographical areas (this will need to be fixed separately)

### DIFF
--- a/app/models/geographical_area.rb
+++ b/app/models/geographical_area.rb
@@ -2,6 +2,7 @@ class GeographicalArea < Sequel::Model
   COUNTRIES_CODES = %w[0 2].freeze
   AREAS_CODES = %w[0 1 2].freeze
   ERGA_OMNES_ID = '1011'.freeze
+  REFERENCED_GEOGRAPHICAL_AREAS = { 'EU' => '1013' }.freeze
 
   plugin :time_machine
   plugin :oplog, primary_key: :geographical_area_sid
@@ -49,7 +50,11 @@ class GeographicalArea < Sequel::Model
   end
 
   def candidate_excluded_geographical_area_ids
-    @candidate_excluded_geographical_area_ids ||= included_geographical_areas.pluck(:geographical_area_id) << geographical_area_id
+    @candidate_excluded_geographical_area_ids ||= contained_geographical_areas.pluck(:geographical_area_id) << geographical_area_id
+  end
+
+  def referenced
+    self.class.where(geographical_area_id: referenced_id).actual.first
   end
 
   def contained_geographical_area_ids
@@ -83,5 +88,11 @@ class GeographicalArea < Sequel::Model
 
   def id
     geographical_area_id
+  end
+
+  private
+
+  def referenced_id
+    REFERENCED_GEOGRAPHICAL_AREAS[geographical_area_id]
   end
 end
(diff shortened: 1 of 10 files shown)


app/models/measure.rb
```diff
@@ -413,8 +409,15 @@ def order_number
     end
   end
 
-  def relevant_for_country?(country_id)
-    return false if measure_excluded_geographical_areas.map(&:excluded_geographical_area).include?(country_id)
+  def relevant_for_country?(country)
+    country_and_memberships = country.included_geographical_areas.pluck(:geographical_area_id) << country.geographical_area_id
```
willfish, line 413: Push this into a method on geographical area

app/models/geographical_area.rb
```diff
@@ -80,4 +89,10 @@ def areas
   def id
     geographical_area_id
   end
+
+  private
+
+  def referenced_id
```
willfish, line 95: We could do a migration to enable a reference but I'd rather hard code this in a hash for now as it very much could be yagni

app/models/rules_of_origin/scheme.rb
```diff
@@ -43,7 +43,7 @@ def proofs
     end
 
     def articles
-      @articles ||= Article.for_scheme(self)
+      @articles ||= RulesOfOrigin::Article.for_scheme(self)
```
willfish, line 46: This fixes intermittent zeitwerk loading issues

spec/factories/geographical_area_factory.rb
```diff
@@ -8,14 +8,6 @@
     validity_start_date   { 3.years.ago.beginning_of_day }
     validity_end_date     { nil }
 
-    trait :fifteen_years do
```
willfish, line 11: These are unused

app/services/meursing_measure_finder_service.rb
```diff
@@ -28,6 +28,6 @@ def call
   attr_reader :root_measure, :additional_code_id
 
   def relevant_for_country?(measure)
-    measure.relevant_for_country?(root_measure.geographical_area_id)
+    measure.relevant_for_country?(root_measure.geographical_area)
```
jebw, line 31: This shifts from using an attribute on `root_measure` to a record lookup - is this called in a loop anyway, ie will it introduce an N+1?

app/models/geographical_area.rb
```diff
@@ -48,6 +49,14 @@ def geographical_area_description
     ds.with_actual(GeographicalAreaMembership).order(Sequel.asc(:geographical_area_id))
   end
 
+  def candidate_excluded_geographical_area_ids
+    @candidate_excluded_geographical_area_ids ||= contained_geographical_areas.pluck(:geographical_area_id) << geographical_area_id
```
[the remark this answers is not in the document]
willfish: Ah good spot

app/services/meursing_measure_finder_service.rb
willfish: Resolved this now :)

PR #758, opened 2022-09-01
HOTT-1832 Fix quota search pagination

jebw:
### Jira link

HOTT-1832

### What?

I have added/removed/altered:

- [x] Added specs to check behaviour when measures which are not end dated reference quota definitions which are end dated
- [x] Changed how quota definitions are retrieved to exclude them from the query, rather then removing blanks post factom with `.compact`

### Why?

I am doing this because:

- If a Measure references a QuotaDefinition, and the Measure is not end dated but the quota definition is - then the scope will still include those quota definitions. Subsequent retrieval via association will return `nil` results for these quota definitions which were previously then being excluded via `.compact`
- This works but leaves a `COUNT(*)` on the original scope returning a different value to the number of records the service will actually return - leading to incorrect pagination
- By excluding these end-dated quota definitions from the original query and removing the use of `.compact` the `COUNT(*)` on the scope is correct, and so is the pagination. There shouldn't be any actual change to the quota definitions returned by the service.

- Changes behaviour of quota definitions search.

### DIFF
--- a/spec/services/quota_search_service_spec.rb
+++ b/spec/services/quota_search_service_spec.rb
@@ -171,7 +171,7 @@ RSpec.describe QuotaSearchService do
     context 'when a quota definition is end dated' do
       before do
         # Modifying records directly because oplog plugin doesn't support dataset CRUD operations
-        QuotaDefinition.dataset.each do |qd|
+        QuotaDefinition.each do |qd|
           qd.validity_end_date = Date.yesterday
           qd.save
         end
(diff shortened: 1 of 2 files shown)


spec/services/quota_search_service_spec.rb
```diff
@@ -167,6 +167,42 @@
         expect(service.call).to eq([quota_definition2])
       end
     end
+
+    context 'with end dated quota_definitions' do
```
willfish, line 171:
```suggestion
    context 'when a quota definition is end dated' do
```

spec/services/quota_search_service_spec.rb
```diff
@@ -167,6 +167,42 @@
...
+      before do
+        # Modifying records directly because oplog plugin doesn't support dataset CRUD operations
+        QuotaDefinition.dataset.each do |qd|
```
willfish, line 174:
```suggestion
        QuotaDefinition.each do |qd|
```

willfish (changes requested):
Seems fine to me

willfish (approved):

PR #759, opened 2022-09-02
HOTT-1873 Expose cumulation_methods on backend RoO API

maverickvi:
### Jira link

[HOTT-<TODO>](https://transformuk.atlassian.net/browse/HOTT-1873)

### What?

I have added/removed/altered:

- [ ] Exposed cumulation_methods on backend RoO API

### Why?

I am doing this because:

- We need cumulation_methods data on frontend

### Have you? (optional)

Screenshot 2022-09-02 at 15 21 16

### DIFF
--- a/spec/factories/rules_of_origin/scheme_factory.rb
+++ b/spec/factories/rules_of_origin/scheme_factory.rb
@@ -7,6 +7,7 @@ FactoryBot.define do
     countries               { %w[FR ES IT DE] }
     footnote                { 'This scheme may be expanded in the future' }
     unilateral              { nil }
+    cumulation_methods      { { 'bilateral' => { 'countries' => %w[GB CA] }, 'extended' => { 'countries' => %w[EU AD] }, 'diagonal' => { 'countries' => %w[EU AD] } } }
 
     trait :with_links do
       links { attributes_for_list :rules_of_origin_link, 2 }
--- a/spec/models/rules_of_origin/scheme_spec.rb
+++ b/spec/models/rules_of_origin/scheme_spec.rb
@@ -16,6 +16,7 @@ RSpec.describe RulesOfOrigin::Scheme do
     it { is_expected.to respond_to :notes }
     it { is_expected.to respond_to :proofs }
     it { is_expected.to respond_to :ord }
+    it { is_expected.to respond_to :cumulation_methods }
   end
 
   describe '#links=' do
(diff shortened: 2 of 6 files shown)


spec/factories/rules_of_origin/scheme_factory.rb
```diff
@@ -7,6 +7,7 @@
     countries               { %w[FR ES IT DE] }
     footnote                { 'This scheme may be expanded in the future' }
     unilateral              { nil }
+    cumulation_methods      { { 'bilateral' => { 'countries' => %w[GB CA] }, 'extended' => { 'countries' => %w[EU AD] }, 'diagonal' => { 'countries' => %w[EU AD] } } }
```
jebw, line 10: This would be easier to read/modify as multiple lines rather then a long single line, eg

```ruby
cumulation_methods do
  {
    'bilateral' => %w[GB CA]
  }
end
```

The nesting under countries can be dropped because the assignment method handles both nested and unnested so this will remove one (minor) complexity from the test data

spec/models/rules_of_origin/scheme_spec.rb
```diff
@@ -39,6 +40,22 @@
...
+    before { instance.cumulation_methods = data }
+
+    let(:instance) { described_class.new }
+
+    let(:expected_response) { { 'bilateral' => %w[GB CA], 'extended' => %w[EU AD], 'diagonal' => %w[EU AD] } }
+
+    let(:data) do
+      { 'bilateral' => { 'countries' => %w[GB CA] }, 'extended' => { 'countries' => %w[EU AD] }, 'diagonal' => { 'countries' => %w[EU AD] } }
+    end
+
```
jebw, line 55: You could maybe add an additional context for the 'unnested' version, eg

```
context 'with direct countries list' do
  let(:data) { { 'bilateral' => %w[GB CA] } }

  it { is_expected to include 'bilateral' => %w[GB CA] }
end

File: spec/models/rules_of_origin/scheme_spec.rb [hunk 3]
```diff
@@ -39,6 +40,22 @@
...
+    it { is_expected.to eq expected_response }
```
jebw, line 56: testing Hash's with `eq` can be error prone, since Ruby 1.9 Hashes have maintained order, hence I think

```ruby
expect({ 'one' => 1, 'two' => 2 }).to eq({'two' => 2, 'one' => 1})
```

will actually fail even though you expect it to pass

I know in general RSpec recommends use of the `includes` matches

jebw (approved):
Looks good, some minor suggestions around test code

The removal of the `country_code` key from the schemes file isn't a problem because its unused

PR #762, opened 2022-09-06
HOTT-1943: Filter countries consistently

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1943

### What?

I have added/removed/altered:

- [x] Support country filtering in measure collection
- [x] Move to filtering using updated measure collection

### Why?

I am doing this because:

- This is required to make sure we support the same behaviour between headings and commodities

### DIFF
--- a/app/models/measure_collection.rb
+++ b/app/models/measure_collection.rb
@@ -26,7 +26,6 @@ class MeasureCollection < SimpleDelegator
                  @measures.select do |measure|
                    measure.relevant_for_country?(filtering_country.geographical_area_id)
                  end
-
                else
                  @measures
                end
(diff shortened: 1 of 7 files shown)


app/models/measure_collection.rb
```diff
@@ -1,13 +1,42 @@
...
+      new(@measures, @filters)
     end
   end
+
+  def apply_country_filter
+    measures = if @filters[:geographical_area_id].present?
+                 @measures.select do |measure|
+                   measure.relevant_for_country?(filtering_country.geographical_area_id)
+                 end
+
```
alexdesi, line 29: empty line

alexdesi (approved):
✅ 🤯 ✅ 🤯 ✅ 🤯 ✅ 🤯

PR #763, opened 2022-09-08
HOTT-1658: Optimise heading cache performance

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Adds migration for heading and chapter short code to view
- [x] Increase heading batch size to 400
- [x] Make Chapter#headings eager loadable
- [x] Make Commodity#chapter and Commodity#heading eager loadable
- [x] Make Heading#commodities eager loadable
- [x] Make Heading#chapter eager loadable
- [x] Adds an eager load graph for cached heading index
- [x] Eager load overview measure associations
- [x] Make use of short code columns
- [x] Exclude non-persisted values from inserts
- [x] Removes dead methods

### Why?

I am doing this because:

- Building the heading indexes currently takes days (totally unnacceptable from a team responsiveness perspective).
- Devs validating changes to endpoints that use the heading cache are also massively affected by the slowness
- Faster ETL finish would mean we could push the ETL process later without affecting live users

### DIFF
--- a/app/models/goods_nomenclature.rb
+++ b/app/models/goods_nomenclature.rb
@@ -190,19 +190,15 @@ class GoodsNomenclature < Sequel::Model
   end
 
   def heading_code
-    heading_short_code + '0' * 6 unless chapter?
+    if heading_short_code
+      heading_short_code + '0' * 6
+    end
   end
 
   def chapter_code
-    chapter_short_code + '0' * 8
-  end
-
-  def heading_short_code
-    goods_nomenclature_item_id.first(4) unless chapter?
-  end
-
-  def chapter_short_code
-    goods_nomenclature_item_id.first(2)
+    if chapter_short_code
+      chapter_short_code + '0' * 8
+    end
   end
 
   def code
--- a/db/migrate/20220908105343_adds_heading_and_chapter_id_to_goods_nomenclatures_view.rb
+++ b/db/migrate/20220908105343_adds_heading_and_chapter_id_to_goods_nomenclatures_view.rb
@@ -1,0 +1,68 @@
+Sequel.migration do
+  up do
+    run %{
+      CREATE OR REPLACE VIEW public.goods_nomenclatures
+      AS SELECT goods_nomenclatures1.goods_nomenclature_sid,
+          goods_nomenclatures1.goods_nomenclature_item_id,
+          goods_nomenclatures1.producline_suffix,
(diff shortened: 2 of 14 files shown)


app/models/goods_nomenclature.rb
```diff
@@ -190,19 +190,15 @@ def chapter_id
   end
 
   def heading_code
-    heading_short_code + '0' * 6 unless chapter?
```
alexdesi, line 193: is the check redundant?

db/migrate/20220908105343_adds_heading_and_chapter_id_to_goods_nomenclatures_view.rb
```diff
@@ -0,0 +1,68 @@
...
+          goods_nomenclatures1.operation,
+          goods_nomenclatures1.operation_date,
+          goods_nomenclatures1.filename,
+          goods_nomenclatures1.path,
+          (
+              CASE WHEN "goods_nomenclatures1"."goods_nomenclature_item_id" LIKE '__00000000' THEN
+                  NULL
+              ELSE
+              LEFT ("goods_nomenclatures1"."goods_nomenclature_item_id",
+                  4)
```
alexdesi, line 21: maybe on the same line?
`("goods_nomenclatures1"."goods_nomenclature_item_id", 4)`

db/migrate/20220908105343_adds_heading_and_chapter_id_to_goods_nomenclatures_view.rb
```diff
@@ -0,0 +1,68 @@
...
+              END) AS "heading_short_code",
+          (
+              LEFT ("goods_nomenclatures1"."goods_nomenclature_item_id",
+                  2)
```
alexdesi, line 25: `("goods_nomenclatures1"."goods_nomenclature_item_id", 2)`

willfish: I actually linted this using a standard sql formatter :)

app/models/goods_nomenclature.rb
willfish: There are circumstances when the short code aren't present so I did need to check it, yeah. Basically when the current goods nomenclature is a chapter - but I've moved the decision making to the view

alexdesi (approved):
✅ We chatted about this, and it works correctly and it's faster

PR #764, opened 2022-09-09
HOTT-1658: Adds fast_declarable?

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Move to using fast declarable to avoid using the goods nomenclature mapper

### Why?

I am doing this because:

- This means we can take advantage of the materialized path to work out whether or not a commodity is declarable

### DIFF
--- a/app/elastic_search_indexes/cache/heading_index.rb
+++ b/app/elastic_search_indexes/cache/heading_index.rb
@@ -1,18 +1,23 @@
 module Cache
   class HeadingIndex < ::Cache::CacheIndex
+    def dataset
+      TimeMachine.now do
+        super.actual
+      end
+    end
+
     def eager_load_graph
       [
         {
-          commodities: [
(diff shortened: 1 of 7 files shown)


alexdesi (approved):
✅

app/elastic_search_indexes/cache/heading_index.rb
```diff
@@ -1,18 +1,23 @@
...
+      end
+    end
+
     def eager_load_graph
       [
         {
-          commodities: [
-            :goods_nomenclature_descriptions,
-            :goods_nomenclature_indents,
-            { heading: [:commodities] }, # TODO: We should be able to replace loading the heading and its commodities with the materialized path concept and avoid eager loading for the goods nomenclature mapper
```
willfish, line 9: Fast declarable does not ask for the heading and its commodities for the goods nomenclature mapper anymore so this is now redundant

PR #765, opened 2022-09-09
HOTT-1947 Fix description for coffins

jebw:
### Jira link

HOTT-1947

### What?

I have added/removed/altered:

- [x] Added a data migration to set the description on coffins

### Why?

I am doing this because:

- The description is missing from the original data

- Includes changes to production data

### Before data migration

Screenshot from 2022-09-09 11-17-44

### After data migration

Screenshot from 2022-09-09 11-34-01

### DIFF
--- a/db/data_migrations/20220909093819_fix_description_for_coffins.rb
+++ b/db/data_migrations/20220909093819_fix_description_for_coffins.rb
@@ -1,0 +1,42 @@
+Sequel.migration do
+  # IMPORTANT! Data migrations should be Idempotent, they may get re-run as part
+  # of data rollbacks
+
+  up do
+    if TradeTariffBackend.uk?
+      Sequel::Model.db[:goods_nomenclature_descriptions_oplog]
+        .where(
+          goods_nomenclature_description_period_sid: 155_818,
+          goods_nomenclature_sid: 107_510,
+          language_id: 'EN',
+          goods_nomenclature_item_id: '4421200000',
+          productline_suffix: '80',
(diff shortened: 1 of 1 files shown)


db/data_migrations/20220909093819_fix_description_for_coffins.rb
```diff
@@ -0,0 +1,42 @@
+Sequel.migration do
+  # IMPORTANT! Data migrations should be Idempotent, they may get re-run as part
+  # of data rollbacks
+
+  up do
+    if TradeTariffBackend.uk?
+      Sequel::Model.db[:goods_nomenclature_descriptions_oplog]
```
willfish, line 7:
```suggestion
      GoodsNomenclatureDescription::Operation.where(goods_nomenclature_sid: 107510, goods_nomenclature_description_period_sid: 155818, filename: 'tariff_dailyExtract_v1_20220816T235959.gzip').update(description: 'Coffins')
```

jebw: This is a concious choice - I'm avoiding loading the data model itself - generally in migrations its best to keep as low level as possible and just stick to a rubyized SQL equivalent

willfish (approved):
I personally don't like duplicating the table name everywhere, but that's personal preference and definitely not blocking. I also think we're being overly defensive by setting all of the attributes to exactly as we expect them to appear. But, again, preference

maverickvi (approved):

maverickvi:
Looks good to me 👍

PR #766, opened 2022-09-09
HOTT-1658: Removes export refund nomenclature query

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Removed unused export refund nomenclature query from measures association

### Why?

I am doing this because:

- This hasn't been supported functionality since the end of 2016 and complicates the measures association


alexdesi (approved):
✅

PR #768, opened 2022-09-09
HOTT-1948 Fixed description of chemical mixtures

jebw:
### Jira link

HOTT-1948

### What?

I have added/removed/altered:

- [x] Added a data migration to correct the description of chemical mixtures

### Why?

I am doing this because:

- The description in the upstream data was blank

- Includes migratory actions which will affect commodity data

### Before

Screenshot from 2022-09-09 12-43-16

### After

Screenshot from 2022-09-09 14-24-52


willfish (approved):
LGTM

PR #769, opened 2022-09-10
HOTT-1658: Adds materialized measures with effective start/end dates

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added materialized view to enable precaching of measures with their real end dates
- [x] Updated measure model to point at this view
- [x] Removed now-unused declarable scopes
- [ ] Added support for materialized views to the oplog plugin
- [ ] Updated specs to reflect these changes

### Why?

I am doing this because:

- We're spending a lot of time loading measures even when filtering to a small list of measures
- This is a big blocker in enabling us to move towards eager loading goods nomenclature measures which will massively speed up measure loading as part of various workflows

### DIFF
--- a/app/lib/sequel/plugins/oplog.rb
+++ b/app/lib/sequel/plugins/oplog.rb
@@ -6,27 +6,18 @@ module Sequel
       DESTROY_OPERATION = 'D'.freeze
 
       def self.configure(model, options = {})
-        model.define_singleton_method(:oplog_options) do
-          options
-        end
-
         model_primary_key = options.fetch(:primary_key, model.primary_key)
         oplog_table_name = options.fetch(:oplog_table_name, :"#{model.table_name}_oplog")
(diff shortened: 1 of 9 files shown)


app/lib/sequel/plugins/oplog.rb
```diff
@@ -7,12 +7,17 @@ module Oplog
 
       def self.configure(model, options = {})
         model_primary_key = options.fetch(:primary_key, model.primary_key)
+        oplog_table_name = options.fetch(:oplog_table_name, :"#{model.table_name}_oplog")
```
willfish, line 10: Supply the view and we know the oplog

willfish:
I've not touched this for a while. My current thinking is that we need a different procedure for testing measures that have been created for a given goods nomenclature which works with a materialized view. When I have some downtime I'll come back to this

PR #777, opened 2022-09-15
HOTT-1968: Generate intercept message links

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1968

### What?

I have added/removed/altered:

- [x] Adds intercept message factories
- [x] Adds InterceptMessage#formatted_message method
- [x] Adds InterceptMessage#formatted_message specs
- [x] Update intercept message locales
- [x] Adjust key in intercept message locale to use space
- [x] Support optional string in goods_nomenclature type prefix
- [x] Adds formatted_message attribute to InterceptMessageSerializer

### Why?

I am doing this because:

- This is required so HMRC can arbitrarily specify goods nomenclature in intercept messages and we generate links for them

### DIFF
--- a/app/models/beta/search/intercept_message.rb
+++ b/app/models/beta/search/intercept_message.rb
@@ -1,10 +1,10 @@
 module Beta
   module Search
     class InterceptMessage
-      CHAPTER_REGEX = /(?<type>chapter)s? (?<optional>code)? (?<code>[0-9]{1,2})(?<terminator>[.,\s])/i
-      HEADING_REGEX = /(?<type>(?<!sub)heading)s? (?<optional>code)? (?<code>[0-9]{4})(?<terminator>[.,\s])/i
-      SUBHEADING_REGEX = /(?<type>subheading)s? (?<optional>code)? (?<code>[0-9]{6,8})(?<terminator>[.,\s])/i
-      COMMODITY_REGEX = /(?<type>commodity|commodities) (?<optional>code)? (?<code>[0-9]{10})(?<terminator>[.,\s])/i
+      CHAPTER_REGEX = /(?<type>chapter)s? (?<optional>code )?(?<code>[0-9]{1,2})(?<terminator>[.,\s])/i
+      HEADING_REGEX = /(?<type>(?<!sub)heading)s? (?<optional>code )?(?<code>[0-9]{4})(?<terminator>[.,\s])/i
+      SUBHEADING_REGEX = /(?<type>subheading)s? (?<optional>code )?(?<code>[0-9]{6,8})(?<terminator>[.,\s])/i
+      COMMODITY_REGEX = /(?<type>commodity|commodities) (?<optional>code )?(?<code>[0-9]{10})(?<terminator>[.,\s])/i
 
       GOODS_NOMENCLATURE_LINK_TRANSFORMERS = {
         CHAPTER_REGEX => lambda do |matched_text|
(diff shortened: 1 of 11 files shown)


app/models/beta/search/intercept_message.rb
```diff
@@ -10,14 +41,27 @@ class InterceptMessage
...
+        return if I18n.t("#{query}.message", default: nil).blank?
+
+        intercept_message = new
 
-        result = new
+        message = I18n.t("#{query}.message")
+        message = message.ends_with?('.') ? message : message.concat('.')
+
+        intercept_message.term = query
+        intercept_message.message = message
```
alexdesi, line 52: it'd be nice to be able to do
`intercept_message = new(query, message)`

maybe we need the `def initialize(query, message)`

app/models/beta/search/intercept_message.rb
```diff
@@ -10,14 +41,27 @@ class InterceptMessage
...
+      end
 
-        result.term = I18n.t("#{query}.title")
-        result.message = I18n.t("#{query}.message")
+      def formatted_message
+        return '' if message.blank?
 
-        result
+        GOODS_NOMENCLATURE_LINK_TRANSFORMERS.each_with_object(message.dup) do |(regex, transformer), transformed_message|
+          transformed_message.gsub!(regex) do |matched_text|
```
alexdesi, line 61: maybe you don't need the bang version 
`.gsub!`
since you are using a block

spec/models/beta/search/intercept_message_spec.rb
```diff
@@ -1,12 +1,80 @@
...
+      end
+
+      it { is_expected.to eq(expected_message) }
+    end
+
+    context 'when there are chapters to generate links for' do
+      subject(:formatted_message) { build(:intercept_message, :with_chapters_to_transform).formatted_message }
+
+      let(:expected_message) do
+        'This should point to (ChaPter 99)[/chapters/99] and (chapters 32)[/chapters/32] and (chapters 1)[/chapters/01] but not chapter 19812321 but for (chapter 9)[/chapters/09].'
```
alexdesi, line 36: is "ChaPter" a misspelling? 🤔

spec/models/beta/search/intercept_message_spec.rb
```diff
@@ -1,12 +1,80 @@
...
+      end
+
+      it { is_expected.to eq(expected_message) }
+    end
+
+    context 'when there are headings to generate links for' do
+      subject(:formatted_message) { build(:intercept_message, :with_headings_to_transform).formatted_message }
+
+      let(:expected_message) do
+        'This should point to (hEadIngs 0101)[/headings/0101] and (heading 0102)[/headings/0102] but not heading 2 or heading 012012 but for (heading 0105)[/headings/0105].'
```
alexdesi, line 46: I guess it was not a mistake, since I can see a similar one
`hEadIngs`

maybe you did it to the insensitiveness for Upcase/downcase ?

app/models/beta/search/intercept_message.rb
willfish: Using each with object so need to do this against the duplicate

spec/models/beta/search/intercept_message_spec.rb
willfish: Case testing

spec/models/beta/search/intercept_message_spec.rb
willfish: Exactly

alexdesi (approved):
nice one! ✅

PR #779, opened 2022-09-16
Fixes intermittent test failures

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Updates constraint for section position

### Why?

I am doing this because:

- Section ids go past 2 digits and the routes fail


alexdesi (approved):
✅

PR #781, opened 2022-09-20
HOTT-1906: Replace en.yml file

maverickvi:
### Jira link

[HOTT-<1906>](https://transformuk.atlassian.net/browse/HOTT-1906)

### What?

I have added/removed/altered:

- [ ] replaced the en.yml file

### Why?

I am doing this because:

- we need the new intercept messaging data


willfish (approved):

maverickvi:
This new version was attached to the ticket and I was asked to just replace the file. I did not work on the design of this implementation.

PR #782, opened 2022-09-21
HOTT-1990 Add support for section on intercept_message.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1990

### What?
To dynamically include hyperlinks to SECTIONS of the tariff in the search intercepts messages

### DIFF
--- a/app/lib/roman_numerals/converter.rb
+++ b/app/lib/roman_numerals/converter.rb
@@ -16,17 +16,6 @@ module RomanNumerals
       1000 => 'M'
     }
 
-    def self.to_roman(value)
-      result = ''
-      @base_digits.keys.reverse.each do |decimal|
-        while value >= decimal
-          value -= decimal
-          result += @base_digits[decimal]
-        end
-      end
-      result
-    end
-
     def self.to_decimal(value)
       value = value.upcase
       result = 0
--- a/app/lib/roman_numerals/converter.rb
+++ b/app/lib/roman_numerals/converter.rb
@@ -16,33 +16,35 @@ module RomanNumerals
       1000 => 'M',
     }
 
-    def self.to_roman(value)
-      result = ''
-      @base_digits.keys.reverse.each do |decimal|
-        while value >= decimal
-          value -= decimal
-          result += @base_digits[decimal]
+    class << self
+      def to_roman(value)
+        result = ''
+        @base_digits.keys.reverse.each do |decimal|
+          while value >= decimal
+            value -= decimal
+            result += @base_digits[decimal]
+          end
         end
+        result
       end
-      result
-    end
 
-    def self.to_decimal(value)
-      return value.to_i if already_decimal?(value)
+      def to_decimal(value)
+        return value.to_i if already_decimal?(value)
 
-      value = value.upcase
-      result = 0
-      @base_digits.values.reverse.each do |roman|
-        while value.start_with? roman
-          value = value.slice(roman.length, value.length)
-          result += @base_digits.key roman
+        value = value.upcase
+        result = 0
+        @base_digits.values.reverse.each do |roman|
+          while value.start_with? roman
+            value = value.slice(roman.length, value.length)
+            result += @base_digits.key roman
+          end
         end
+        result
       end
-      result
-    end
 
-    def self.already_decimal?(value)
-      (value =~ /^\d/) != nil
+      def already_decimal?(value)
+        (value =~ /^\d/) != nil
+      end
     end
   end
 end
--- a/app/lib/roman_numerals/converter.rb
+++ b/app/lib/roman_numerals/converter.rb
@@ -1,22 +1,35 @@
 module RomanNumerals
   class Converter
     @base_digits = {
-      1    => 'I',
-      4    => 'IV',
-      5    => 'V',
-      9    => 'IX',
-      10   => 'X',
-      40   => 'XL',
-      50   => 'L',
-      90   => 'XC',
-      100  => 'C',
-      400  => 'CD',
-      500  => 'D',
-      900  => 'CM',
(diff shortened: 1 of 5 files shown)


spec/lib/roman_numerals/converter_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    50   => 'L',
+    90   => 'XC',
+    100  => 'C',
+    400  => 'CD',
+    500  => 'D',
+    900  => 'CM',
+    1000 => 'M'
+  }
+
+  describe ".to_decimal" do
```
willfish, line 20: Do we want a to_roman spec or are we not going to use it?

spec/models/beta/search/intercept_message_spec.rb
```diff
@@ -76,5 +76,18 @@
 
       it { is_expected.to eq(expected_message) }
     end
+
+    context 'when there is a section followed by a roman number' do
+      subject(:formatted_message) do
+        build(:intercept_message,
```
willfish, line 82: We should chuck this in a factory to be consistent

app/lib/roman_numerals/converter.rb
```diff
@@ -0,0 +1,32 @@
...
+        while value.start_with? roman
+          value = value.slice(roman.length, value.length)
+          result += @base_digits.key roman
+        end
+      end
+      result
+    end
+  end
+
+end
```
willfish, line 32: Do you need to update your editor to handle file termination?

app/models/beta/search/intercept_message.rb
```diff
@@ -1,12 +1,21 @@
 module Beta
   module Search
     class InterceptMessage
+      SECTION_REGEX = /(?<type>section)s? (?<roman_number>[XVI]{0,10})[.,\s)]/i
```
willfish, line 4: I think we should support this as a number _and_ a roman number and probably also support optional position/id/code strings since sections are referred to in this way and the input is coming from HMRC/Matt/Sunny

spec/factories/intercept_message_factory.rb
```diff
@@ -33,6 +33,12 @@
       end
     end
 
+    trait :with_section_to_transform do
```
willfish, line 36: I'd hoist this above chapter to reflect the hierarchical position of the section but that's a small preference/semantic thing

spec/lib/roman_numerals/converter_spec.rb
alexdesi: Yes we do: we need it when we capture the section having decimal digits.
In the representation, we want to show the Roman form (I guess ... it looks nicer)

spec/models/beta/search/intercept_message_spec.rb
alexdesi: tbh I was considering the usefulness of the factories in this case,
I was always jumping back and forth in the factory/test, and it make the test couple with the factory 🤔

app/lib/roman_numerals/converter.rb
alexdesi: yep, this and the support for decimal was a WIP (both are done now)

app/models/beta/search/intercept_message.rb
alexdesi: done!

spec/models/beta/search/intercept_message_spec.rb
willfish: I'm happy to change to not using traits but I'd make the rest conform to this new standard rather than have a mixture, etc

willfish (approved):
No blocking comments :smile:

app/models/beta/search/intercept_message.rb
```diff
@@ -1,12 +1,23 @@
...
       SUBHEADING_REGEX = /(?<type>subheading)s? (?<optional>code )?(?<code>[0-9]{6,8})(?<terminator>[.,\s)])/i
       COMMODITY_REGEX = /(?<type>commodity|commodities) (?<optional>code )?(?<code>[0-9]{10})(?<terminator>[.,\s)])/i
 
       GOODS_NOMENCLATURE_LINK_TRANSFORMERS = {
+        SECTION_REGEX => lambda do |matched_text|
+          match = matched_text.match(SECTION_REGEX)
+
+          section_id = RomanNumerals::Converter.to_decimal(match[:roman_number])
+
+          roman_section_id = RomanNumerals::Converter.to_roman(section_id)
```
willfish, line 16: I guess you can pull this from the match[:roman_number] rather than converting again?

app/models/beta/search/intercept_message.rb
```diff
@@ -1,12 +1,23 @@
 module Beta
   module Search
     class InterceptMessage
+      SECTION_REGEX = /(?<type>section)s? (?<roman_number>[XVI\d]{0,10})(?<terminator>[.,\s)])/i
```
willfish, line 4:
```
  section code 1
  section position 1
  section id 1
  section 1
  section I
  section code I
  section postion I
  section id I
```
```ruby
(?<type>section)s? (?<optional>(code|position|id) )?(?<number>[XVI\d]{0,10})(?<terminator>[.,\s)])
```

app/models/beta/search/intercept_message.rb
alexdesi: (it is called just `code` now).
unfortunatelly, if the code is decimal, we cannot reuse `match[:code]`, because we want so show all the section links in Roman

willfish: I don't think that's true. I think I went with not transforming the way the text for the link appears in the other examples basically to preserve the what the user wrote in their spreadsheet.

PR #783, opened 2022-09-22
HOTT-1368: Separate UK and XI ETL runs

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1368

### What?

I have added/removed/altered:

- [x] Added two new schedules for an XI and a UK run with different cron configurations

### Why?

I am doing this because:

- This has been a long-standing requirement but was made important by the fact that we need to make sure that we can merge the UK/XI clusters which requires the two runs to not exhaust write limits by running at similar times.

### DIFF
--- a/app/lib/trade_tariff_backend.rb
+++ b/app/lib/trade_tariff_backend.rb
@@ -5,8 +5,6 @@ module TradeTariffBackend
       'xi' => 'EUR',
     }.freeze
 
-    STOP_WORDS_FILE = Rails.root.join('db/beta/search/stop_words.yml')
-
     def configure
       yield self
     end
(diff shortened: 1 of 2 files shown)


app/lib/trade_tariff_backend.rb
```diff
@@ -238,7 +236,11 @@ def lemmatizer
     end
 
     def stop_words
-      @stop_words ||= YAML.load_file(STOP_WORDS_FILE)[:stop_words]
+      @stop_words ||= YAML.load_file(stop_words_file)[:stop_words]
+    end
+
+    def stop_words_file
+      Rails.root.join('db/beta/search/stop_words.yml')
```
alexdesi, line 243: I guess this is Rubocop  :)

alexdesi (commented):
it looks good! 🍏

willfish:
Validated on dev and getting the correct configuration for each worker

PR #784, opened 2022-09-22
HOTT-2021: Avoid index name collisions between xi and uk

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2021

### What?

I have added/removed/altered:

- [x] Support pulling out search index via the plugin
- [x] Adds service (e.g. xi/uk) to namespace
- [x] Fixes specs after index namespacing

### Why?

I am doing this because:

- This is required to avoid collisions of index names when different service
  backends hit the same indexes

### More info

This is part of a wider merging of the two clusters on each space potentially saving thousands in infrastructure bills each year for HMRC.

The plan is to:

1. Separate out the UK/XI schedules to avoid spikes in writes between the two backend workers https://github.com/trade-tariff/trade-tariff-backend/pull/783
2. Separate out the UK/XI index names to avoid collisions in different data being written/read against the same index names
3. Create a new cluster that both backends will use and populate it with the newly named indexes - being careful to populate them sequentially
4. Run the e2e specs against the new cluster


alexdesi (approved):
it looks good! ✅

willfish:
Merging since this is behind a feature flag which I'll enable separately

PR #790, opened 2022-09-28
Hott-2045: Added condition to check if measure is of special nature c…

maverickvi:
…ertificate_type_code A

### Jira link

[HOTT-<2045>](https://transformuk.atlassian.net/browse/HOTT-2045)

### What?

I have added/removed/altered:

- [ ] Added condition to check if measure is of special nature certificate_type_code A

### Why?

I am doing this because:

- We need to perform this check in upcoming tickets

### DIFF
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -494,6 +494,10 @@ class Measure < Sequel::Model
     @meursing_measures ||= MeursingMeasureFinderService.new(self, meursing_additional_code_id).call
   end
 
+  def special_nature?
+    measure_conditions.map(&:certificate_type_code).include?(Certificate::SPECIAL_NATURE)
+  end
+
   private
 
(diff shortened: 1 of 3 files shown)


app/models/measure.rb
```diff
@@ -494,6 +494,10 @@ def meursing_measures
     @meursing_measures ||= MeursingMeasureFinderService.new(self, meursing_additional_code_id).call
   end
 
+  def special_nature?
+    measure_conditions.map(&:certificate_type_code).include?(Certificate::SPECIAL_NATURE)
```
willfish, line 498:
```suggestion
    measure_conditions.any?(&:special_nature?)
```

willfish:
I reckon you'll want to add this to the v2 measure serializer?

maverickvi:
This will be done as part of 2050 ticket.

PR #791, opened 2022-09-29
HOTT-1998: Enable synonyms

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2078

### What?

I have added/removed/altered:

- [x] Enable synonyms based on an environmentally controlled reference variable
- [x] Set development/testing synonyms

### Why?

I am doing this because:

- This is required to get recently configured opensearch packages working on every space


alexdesi (approved):
it looks good to me ✅

PR #793, opened 2022-10-03
HOTT-2087: Surface hjid for all geographical areas

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2087

### What?

I have added/removed/altered:

- [x] Adds globally-excluded geographical area factory trait
- [x] Enable no exclusions in cached_geographical_area_service
- [x] Pass exclude_none filter to service
- [x] Adds hjid to geo area serializer
- [x] Adds missing spec for serializer
- [x] Adds exclude_none behaviour checks

### Why?

I am doing this because:

We need this to automate the pipeline for providing XLSX files for HMRC (amongst other people).

- When the file includes a membership in the UK CDS file we need to be able to look up the description by the hjid.
- When the file includes memberships for excluded geographical areas we still want to show these in the file

### DIFF
--- a/app/controllers/api/v2/geographical_areas_controller.rb
+++ b/app/controllers/api/v2/geographical_areas_controller.rb
@@ -2,7 +2,7 @@ module Api
   module V2
     class GeographicalAreasController < ApiController
       def index
-        render json: CachedGeographicalAreaService.new(actual_date).call
+        render json: CachedGeographicalAreaService.new(actual_date, exclude_none:, countries: false).call
       end
 
       def show
@@ -10,7 +10,7 @@ module Api
       end
 
       def countries
-        render json: CachedGeographicalAreaService.new(actual_date, countries: true).call
+        render json: CachedGeographicalAreaService.new(actual_date, exclude_none:, countries: true).call
       end
 
       private
@@ -24,6 +24,10 @@ module Api
 
       def geographical_area
         GeographicalArea.actual.by_id(params[:id]).eager(:geographical_area_descriptions, :contained_geographical_areas).take
+      end
+
+      def exclude_none
+        params.fetch(:filter, {}).permit(:exclude_none).fetch(:exclude_none, false) == 'true'
       end
     end
   end
--- a/spec/controllers/api/v2/geographical_areas_controller_spec.rb
+++ b/spec/controllers/api/v2/geographical_areas_controller_spec.rb
@@ -2,61 +2,65 @@ RSpec.describe Api::V2::GeographicalAreasController do
   before do
     Rails.cache.clear
     allow(Rails.cache).to receive(:fetch).and_call_original
+
+    country_geographical_area
+    group_geographical_area
+    region_geographical_area
+    globally_excluded_area
   end
 
-  let!(:country_geographical_area) { create(:geographical_area, :with_description, :country) }
-  let!(:group_geographical_area) { create(:geographical_area, :with_description, :group) }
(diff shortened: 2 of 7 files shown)


app/controllers/api/v2/geographical_areas_controller.rb
```diff
@@ -25,6 +25,10 @@ def serialized_geographical_area
       def geographical_area
         GeographicalArea.actual.by_id(params[:id]).eager(:geographical_area_descriptions, :contained_geographical_areas).take
       end
+
+      def exclude_none
+        params.fetch(:filter, {}).permit(:exclude_none).fetch(:exclude_none, false) == 'true'
```
alexdesi, line 30: thinking about this:
if `exclude_none` is not present,
it returns `false` which is a `Boolean`,
and it is compared to `true` which is a `String`
... which sounds weird 🤔

spec/controllers/api/v2/geographical_areas_controller_spec.rb
```diff
@@ -2,61 +2,65 @@
...
 
     it 'caches the serialized countries' do
       do_response
 
-      expect(Rails.cache).to have_received(:fetch).with("_geographical-areas-countries-#{actual_date}", expires_in: 24.hours)
+      expect(Rails.cache).to have_received(:fetch).with("_geographical-areas-#{actual_date}-true-false", expires_in: 24.hours)
+    end
+
+    context 'when the exclude none filter is passed' do
+      subject(:do_response) { get :countries, params: { filter: { exclude_none: 'true' } } }
```
alexdesi, line 34: have you tried with 
`params: { filter: { exclude_none: 'false' } }`

and should we consider capitalized version FALSE/TRUE? (what happens in that cases?)

spec/controllers/api/v2/geographical_areas_controller_spec.rb
```diff
@@ -70,105 +74,42 @@
...
+          .dig('relationships', 'children_geographical_areas', 'data')
+          .pluck('id')
+          .sort
+
+        expected_children_geographical_areas = [
+          country_geographical_area.geographical_area_id,
+          region_geographical_area.geographical_area_id,
+        ].sort
+
+        expect(actual_children_geographical_areas).to eq(expected_children_geographical_areas)
```
alexdesi, line 90: I guess you don't need to sort the arrays with `match_array`:

```ruby
expect(actual_children_geographical_areas).to match_array [
                                                    country_geographical_area.geographical_area_id,
                                                      region_geographical_area.geographical_area_id,
                                                    ]
```

alexdesi (approved):
it looks good, but I'd double-check the `exclude_note` method.

spec/controllers/api/v2/geographical_areas_controller_spec.rb
willfish: I could do that but the reality is I'm never going to use the api in that way :thinking: and this is me driving it.

I'm also never going to document this feature as this is an endpoint used by the frontend, only

app/controllers/api/v2/geographical_areas_controller.rb
willfish: I think what's weird is we shouldn't have a separate endpoint for countries, it should also just be a filter

willfish: `/api/v2/geographical_areas?filter['countries']=false,filter['exclude_none']=true`

willfish:
This is being consumed https://github.com/trade-tariff/download-CDS-files/pull/2

PR #800, opened 2022-10-04
HOTT-2087: Surface measures for quota order number index

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2087

### What?

I have added/removed/altered:

- [x] Surface simple measures for quotas
- [x] Eager load definition measures
- [x] Update specs to reflect changes in serializer

### Why?

I am doing this because:

- This is required to avoid needing a database when we automate the generation of the XLSX files https://github.com/trade-tariff/download-CDS-files/pull/2


willfish:
This is an internal api, only. Merging

PR #801, opened 2022-10-05
Hott-2050: Added preference codes and descriptions to commodities API

maverickvi:
### Jira link

[HOTT-<2050>](https://transformuk.atlassian.net/browse/HOTT-2050)

### What?

I have added/removed/altered:

- [ ] Added preference codes and descriptions to commodities API

### Why?

I am doing this because:

- We need to access preference codes on front end

### Have you? (optional)

Screenshot 2022-10-05 at 12 09 34


app/models/preference_code.rb
```diff
@@ -5,4 +5,67 @@ def initialize(id:, description:)
   end
 
   attr_accessor :id, :description
+
+  def self.find(commodity, measure)
```
alexdesi, line 9: Wow, this looks really complicated :) 
we should aim for a simpler implementation, which I don't know yet 😄

alexdesi: maybe could we split logically this into different methods,
I noticed that some internal conditions (IF) is based on the `commodity` other on the `measure`, and some has not conditions, this seems interesting 🤔

spec/models/commodity_spec.rb
```diff
@@ -825,4 +825,51 @@
       it { is_expected.not_to be_fast_declarable }
     end
   end
+
+  describe '#special_nature?' do
+    context 'when commodity has measures with atleast one measure condition containing special nature certificate' do
```
alexdesi, line 830: space missing:
atleast => at least

spec/models/commodity_spec.rb
```diff
@@ -825,4 +825,51 @@
...
+          :with_measure_conditions,
+          :with_special_nature,
+          goods_nomenclature_sid: commodity.goods_nomenclature_sid,
+        )
+      end
+
+      it { expect(commodity.special_nature?).to eq(true) }
+    end
+
+    context 'when commodity does not have measures with atleast one measure condition containing special nature certificate' do
```
alexdesi, line 846: do you mean 
`when commodity has no measure conditions containing special nature certificate`
?

spec/models/measure_spec.rb
```diff
@@ -825,6 +825,20 @@
     end
   end
 
+  describe '#authorised_use' do
+    context 'when measure has atleast one measure condition with authorised use certificate' do
```
alexdesi, line 829: atleast -> at least (missing space)

app/models/commodity.rb
```diff
@@ -187,6 +187,14 @@ def goods_nomenclature_class
     declarable? ? 'Commodity' : 'Subheading'
   end
 
+  def special_nature?
+    measures.any? { |m| m.measure_conditions.map(&:certificate_type_code).include?(Certificate::SPECIAL_NATURE) }
```
willfish, line 191: I think it makes sense to add these to the commodity presenter which cached the result of calling the massively inefficient measures dataset. The measures dataset isn't actually eager loadable _and_ it does a whole bunch of unions to work out the start and end dates of a measure.

https://github.com/trade-tariff/trade-tariff-backend/blob/main/app/presenters/api/v2/commodities/commodity_presenter.rb#L14-L23

willfish: Another thing to know is that preference codes only apply to import measures as they're part of your declarations

PR #812, opened 2022-10-14
HOTT-2127 Added Solomon Islands to Pacific treaty

jebw:
### Jira link

HOTT-2127

### What?

I have added/removed/altered:

- [x] Added Solomon Islands (SB) to the Pacific treaty

### Why?

I am doing this because:

- Solomon islands is party to that treaty and rules of origin should reflect that

- Low risk

### Before without wizard

Screenshot from 2022-10-14 11-25-23

### After with wizard

Screenshot from 2022-10-14 11-26-25

### After without wizard

image


mattlavis-transform (approved):
Approved

PR #814, opened 2022-10-17
HOTT-2151: Stop classifying non-declarables

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2151

### What?

I have added/removed/altered:

- [x] Removed classifications for non-declarable results
- [x] Updated test coverage to reflect the change

### Why?

I am doing this because:

- We're not even interesting in chapters and headings from a search result perspective and instead target headings and commodities that can be traded


app/models/beta/search/goods_nomenclature_query.rb
```diff
@@ -89,19 +89,20 @@ def goods_nomenclature_item_id_term_query
...
-          bool: {
-            must: Api::Beta::GoodsNomenclatureFilterGeneratorService.new(@filters).call,
-          },
-        }
+        part = { bool: {} }
+
+        part[:bool][:must] = static_and_dynamic_filters
+        part[:bool][:must] << declarable_filter
+
+        part
```
alexdesi, line 105: Since this method set only `must` part of the filter, should this be called `must_part`? 🤔 
But actually there is another method called `must_part`, which apparently does not set `must` ... strange 🤔

alexdesi (approved):
Approved ✅ 
I just find a bit confusing the naming of the filter_part, and must_part

app/models/beta/search/goods_nomenclature_query.rb
willfish: Spoke offline and added some context for this

PR #816, opened 2022-10-17
HOTT-2154: Support simpler single classifications per facet

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2154

### What?

We used to accumulate multiple classifications for a given goods nomenclature for each level we go up the tree.

This isn't actually helpful as the only thing we're interested in classifying are things we can trade which have single (and therefore specific) classifications.

I have added/removed/altered:

- [x] Added support for simpler/singular classifications
- [x] Updated specs to reflect change

### Why?

I am doing this because:

- We only care about low-down and specific search results for specific declarables
- This actually matches Matt's implementation and is less clever by half.


alexdesi (approved):
it looks good! 🍏

PR #817, opened 2022-10-17
HOTT-2128 Namespace News Items

jebw:
### Jira link

HOTT-2128

### What?

I have added/removed/altered:

- [x] Namespaced the news items models and corresponding serializers

### Why?

I am doing this because:

- We'll be adding a News::Collections model so this change will allow grouping related these models together

- Low, internal refactoring with good specs


maverickvi (approved):

app/controllers/api/admin/news_items_controller.rb
```diff
@@ -8,13 +8,13 @@ def index
       end
 
       def show
-        news_item = NewsItem.with_pk!(params[:id])
+        news_item = ::News::Item.with_pk!(params[:id])
```
willfish, line 11: Should this technically mean a rename of the controller to follow standard patterns?

willfish:
```ruby
news/items_controller.rb
```

willfish (approved):

app/controllers/api/admin/news_items_controller.rb
jebw: decided to leave it since we're probably not adding more news controllers

PR #827, opened 2022-10-18
HOTT-2139: Fixes broken intercept message links

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2139

### What?

I have added/removed/altered:

- [x] Fixed the links generated for the formatted message of the intercept message model
- [x] Updated specs to reflect change

### Why?

I am doing this because:

- Otherwise we can't render these lives with GovSpeak


alexdesi (approved):
✅ ✅ ✅

PR #830, opened 2022-10-19
HOTT-2128 Add News Collections

jebw:
### Jira link

HOTT-2128

### What?

I have added/removed/altered:

- [x] Added a `news_collections` table + migration
- [x] Added a `news_collections_news_items` join table + migration
- [x] Added a News Collection model
- [x] Setup many_to_many relationships between `News::Collection` and `News::Item`
- [x] Added `collection_ids` accessor methods to `News::Item` to update the collection relationships
- [x] Removed some custom validation code from News::Item
- [x] Namespaced the existing Admin News Item controller
- [x] Added a News::Collection serializer
- [x] Added a `/api/admin/news/collection` end point for querying all the news items

### Why?

I am doing this because:

- We want to group news items into collections
- The `collection_ids` accessors were added because the Sequel gem doesn't provide an equivalent to active records functionality for updating relationships from the form via checkboxes on their ids
- Custom validation code is not needed if you change the config on `auto_validations` plugin
- Collections endpoint will be needed to list all the available collections a user can tick in the Admin area UI

- Includes database migrations but they are only additive and currently there are now dependencies from existing code onto the new database entities


app/controllers/api/admin/news/items_controller.rb
```diff
@@ -0,0 +1,94 @@
+module Api
+  module Admin
+    module News
+      class ItemsController < ApiController
+        before_action :authenticate_user!
+
+        def index
+          render json: serialize(news_items.to_a, pagination_meta)
```
willfish, line 8: I'd ask for the news items rather than tell, here.

```suggestion
          render json: serialized_news_items
```

willfish (approved):
LGTM

app/controllers/api/admin/news/items_controller.rb
jebw: :man_facepalming:

jebw: Looking at this in more detail - serialize is used with multiple sources - both a single instance (eg #create) and arrays, (#index) so its not possible to have a general serialization method without giving it context (ie passing in the data to be serialized)

I'll leave this as is otherwise I'm getting into adding separate methods for all scenarios rather then generalising the common bit of code (choosing the serializer to use)

PR #832, opened 2022-10-20
Cleanup some dead files

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [ ] Added foo in bar
- [ ] Removed baz

### Why?

I am doing this because:

-
-
-

### Have you? (optional)


lib/tasks/download.rake
```diff
@@ -1,80 +0,0 @@
-require 'zip'
-
-namespace :tariff do
-  namespace :download do
-    desc 'Download all Taric files from production AWS S3'
-    task taric: %w[environment] do
```
alexdesi, line 6: this rake tasks sounds useful,
sure we want to delete it?

lib/tasks/tariff.rake
```diff
@@ -110,228 +78,20 @@ namespace :tariff do
...
-        Dir[Rails.root.join('db/notes/chapters/*')].each do |file|
-          note = YAML.safe_load(File.read(file), [Symbol])
-          chapter_note = ChapterNote.find(section_id: note[:section],
-                                          chapter_id: note[:chapter].to_s) || ChapterNote.new(section_id: note[:section], chapter_id: note[:chapter].to_s)
-          chapter_note.content = note[:content]
-          chapter_note.save
-        end
-      end
-
-      desc 'Load measurment unit abbriviations'
```
alexdesi, line 172: nice cleaning ... it looks like this task would have more sense as data migration

PR #833, opened 2022-10-20
HOTT-2128 Extend news apis

jebw:
### Jira link

HOTT-2128

### What?

I have added/removed/altered:

- [x] Added migration to add `precis` and `slug` fields to database
- [x] Added precis attribute
- [x] Added slug attribute
- [x] Auto assign slug when not provided
- [x] Expose collections, slug and precis on v2 news item api
- [x] Expose collection_ids, slug and precis on admin news item api

### Why?

I am doing this because:

- We will want to use these attributes on the frontend
- We will want to edit these attributes in the admin area

- Significant: Changes `news_items` database table and news items api which is used on _all_ pages

### DIFF
--- a/app/models/news/item.rb
+++ b/app/models/news/item.rb
@@ -4,6 +4,7 @@ module News
     plugin :auto_validations, not_null: :presence
 
     DISPLAY_REGULAR = 0
+    MAX_SLUG_LENGTH = 254
 
     many_to_many :collections, join_table: :news_collections_news_items,
                                order: :name
@@ -26,12 +27,19 @@ module News
       super
 
       @collection_ids = normalise_ids(collection_ids)
+      generate_or_normalise_slug!
     end
 
     def after_save
       (collection_ids - collections.pluck(:id)).each(&method(:add_collection))
       (collections.pluck(:id) - (collection_ids & collections.pluck(:id)))
         .each(&method(:remove_collection))
+    end
+
+    def validate
+      super
+
+      validates_presence :slug
     end
 
     dataset_module do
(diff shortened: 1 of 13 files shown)


app/models/news/item.rb
```diff
@@ -67,5 +75,17 @@ def for_target(target_name)
     def normalise_ids(ids)
       Array.wrap(ids).map(&:to_i).compact.uniq
     end
+
+    def generate_or_normalise_slug!
+      current_slug = slug.presence || title.presence
+      return unless current_slug
+
+      normalised_slug = normalise_slug(slug.presence || title)
+      self.slug = normalised_slug if slug != normalised_slug
```
willfish, line 84: Isn't it better not to mutate existing state and just to ask for a slug?  It seems like the state mutation isn't actually  needed and generally prima-donna functions are a smell

```ruby
def slug
  @slug ||= begin
    current_slug = slug.presence || title
    
    if current_slug.present?
      slug.downcase.gsub(/\s+/, '-').gsub(/[^a-z0-9-]/, '').first(MAX_SLUG_LENGTH)
    end
  end
end
```

db/migrate/20221020121609_add_precis_and_slug_to_news_items.rb
```diff
@@ -0,0 +1,10 @@
+Sequel.migration do
+  change do
+    alter_table :news_items do
+      add_column :precis, String
```
willfish, line 4: I have no idea what a precis is. Is there something from common vernacular that we might use for this?

willfish (approved):
Nothing blocking but not a fan of prima-donna methods generally

db/migrate/20221020121609_add_precis_and_slug_to_news_items.rb
willfish: Okay, googled this. Could we use abstract/summary, instead?

PR #836, opened 2022-10-24
HOTT-2167: Adds admin quotas endpoint

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2167

image

### What?

I have added/removed/altered:

- [x] Adds route for quota order number
- [x] Adds controller for admin quota order numbers
- [x] Adds serializers for quota order numbers
- [x] Adds convenience scope for quota order number

### Why?

I am doing this because:

- This is required so we can find a specific quota and render its balance events in a graph in the admin UI

### DIFF
--- a/app/models/quota_order_number.rb
+++ b/app/models/quota_order_number.rb
@@ -9,6 +9,11 @@ class QuotaOrderNumber < Sequel::Model
   end
 
   dataset_module do
+    def by_order_number(order_number)
+      where(quota_order_number_id: order_number)
+        .actual
+    end
+
     def with_quota_definitions
       inner_join(:quota_definitions, quota_order_numbers__quota_order_number_sid: :quota_definitions__quota_order_number_sid)
         .actual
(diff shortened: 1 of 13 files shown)


app/models/quota_order_number.rb
```diff
@@ -9,6 +9,11 @@ class QuotaOrderNumber < Sequel::Model
   end
 
   dataset_module do
+    def by_order_number(order_number)
```
jebw, line 12: maybe `for_order_number` - **by** would imply sorting rather then filtering

or possibly `find_by_order_number` ? Though that might imply returning a single record rather than an array

jebw (approved):
I've added a naming suggestion to the scope. My only query is the retrieval of an _unlimited_ number of `quota_balance_events`? I'm assuming this would be a maximum of ~90 because of the definition changes every 3 months but I'm not sure if I've got that right?

app/models/quota_order_number.rb
willfish: I'm just going with a common idiom in the backend, here (by_code, by_producline_suffix, etc)

PR #838, opened 2022-10-25
HOTT-2167: Serialize definitions not order numbers

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2167

### What?

I have added/removed/altered:

- [x] Altered quota definitions endpoint to return quota definition NOT order number

### Why?

I am doing this because:

- This is the correct thing to do, but the real driver of this is that the admin app doesn't work with nested resources in the included section of the jsonapi response and changing this behaviour isn't a small piece of work


alexdesi (approved):
it looks good ✅

PR #839, opened 2022-10-25
HOTT-2167: Adds measurement unit to quota definition

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1531

image

### What?

I have added/removed/altered:

- [x] Serialize definitions not order numbers
- [x] Adds measurement unit to quota definition serializer

### Why?

I am doing this because:

- This is requirement for the presentation of the unit in the current quota definition page


alexdesi (approved):
lgtm ✅

PR #843, opened 2022-10-31
HOTT-2029: Adds measures show api

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2029

### What?

I have added/removed/altered:

- [x] Added an api to return information for a specific measure

### Why?

I am doing this because:

- This is required by DIT to validate specific measures

### DIFF
--- a/spec/presenters/api/v2/measures/measure_presenter_spec.rb
+++ b/spec/presenters/api/v2/measures/measure_presenter_spec.rb
@@ -10,6 +10,38 @@ RSpec.describe Api::V2::Measures::MeasurePresenter do
     end
   end
 
+  context 'when measure has a generating regulation' do
+    let(:measure) { create(:measure, :with_base_regulation) }
+
+    describe '#measure_generating_legal_act' do
+      it 'returns the MeasureLegalActPresenter' do
+        expect(presenter.measure_generating_legal_act).to be_instance_of(Api::V2::Measures::MeasureLegalActPresenter)
+      end
+    end
+
+    describe '#measure_generating_legal_act_id' do
+      it 'has correct ID' do
+        expect(presenter.measure_generating_legal_act_id).to eq(measure.measure_generating_regulation_id)
+      end
+    end
+  end
+
+  context 'when measure has a justification_legal_act' do
+    let(:measure) { create(:measure, :with_justification_regulation) }
+
+    describe '#measure_generating_legal_act' do
+      it 'returns the MeasureLegalActPresenter' do
+        expect(presenter.justification_legal_act).to be_instance_of(Api::V2::Measures::MeasureLegalActPresenter)
+      end
+    end
+
+    describe '#measure_generating_legal_act_id' do
+      it 'has correct ID' do
+        expect(presenter.justification_legal_act_id).to eq(measure.justification_regulation_id)
+      end
+    end
+  end
+
   describe 'exclusions' do
     let(:gb) { create(:geographical_area, geographical_area_id: 'GB') }
     let(:xi) { create(:geographical_area, geographical_area_id: 'XI') }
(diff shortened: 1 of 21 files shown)


spec/presenters/api/v2/measures/measure_presenter_spec.rb
```diff
@@ -10,6 +10,38 @@
     end
   end
 
+  context 'when measure has a generating regulation' do
+    let(:measure) { create(:measure, :with_base_regulation) }
+
+    describe '#measure_generating_legal_act' do
+      it 'returns the MeasureLegalActPresenter' do
+        expect(presenter.measure_generating_legal_act).to be_instance_of(Api::V2::Measures::MeasureLegalActPresenter)
```
jebw, line 18: This us a useful spec and should be kept but it also needs to be checking the object returned is the expected one? eg comparing the id as the `measure_generating_legal_act_id` spec does

spec/presenters/api/v2/measures/measure_presenter_spec.rb
```diff
@@ -10,6 +10,38 @@
...
+      it 'has correct ID' do
+        expect(presenter.measure_generating_legal_act_id).to eq(measure.measure_generating_regulation_id)
+      end
+    end
+  end
+
+  context 'when measure has a justification_legal_act' do
+    let(:measure) { create(:measure, :with_justification_regulation) }
+
+    describe '#measure_generating_legal_act' do
```
jebw, line 32: Incorrect method name

spec/presenters/api/v2/measures/measure_presenter_spec.rb
```diff
@@ -10,6 +10,38 @@
...
+      it 'returns the MeasureLegalActPresenter' do
+        expect(presenter.justification_legal_act).to be_instance_of(Api::V2::Measures::MeasureLegalActPresenter)
+      end
+    end
+
+    describe '#measure_generating_legal_act_id' do
```
jebw, line 38: Incorrect method name

spec/presenters/api/v2/measures/measure_presenter_spec.rb
```diff
@@ -10,6 +10,38 @@
...
+      end
+    end
+  end
+
+  context 'when measure has a justification_legal_act' do
+    let(:measure) { create(:measure, :with_justification_regulation) }
+
+    describe '#measure_generating_legal_act' do
+      it 'returns the MeasureLegalActPresenter' do
+        expect(presenter.justification_legal_act).to be_instance_of(Api::V2::Measures::MeasureLegalActPresenter)
```
jebw, line 34: Can this also have an additional spec for comparing the id as well as the type

jebw (changes requested):
Looks about ready to go, couple of cut n paste errors and missing checks for the content of the presented legal act

spec/presenters/api/v2/measures/measure_presenter_spec.rb
alexdesi: Hi @jebw ,
I think this is covered by the following spec: https://github.com/trade-tariff/trade-tariff-backend/pull/843/files#diff-131c0beeeee7f7d07f72f9baf49e3c3e1e46d6e74cc1dff07891103108ac602dR23

Do you agree?

spec/presenters/api/v2/measures/measure_presenter_spec.rb
alexdesi: I've added this:
`expect(presenter.justification_legal_act.base_regulation_id).to eq(measure.justification_regulation_id)`
to check the `justification_legal_act` is the correct one.

The following one checks the `measure_generating_legal_act_id` (which is a delegation).

jebw (approved):
LGTM :raised_hands:

PR #844, opened 2022-10-31
HOTT-2185: Adds updated synonyms

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2185

### What?

I have added/removed/altered:

- [x] Added updated synonyms

### Why?

I am doing this because:

- This is required to fix multi-word synonyms that are currently not supported and to introduce new synonyms


alexdesi (approved):
good one, I see only txt changes ✅

PR #846, opened 2022-10-31
Remove unused validity periods attribute

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removed goods_nomenclature_class from validity periods serializer

### Why?

I am doing this because:

- This creates a dependency on heading commodities and doesn't work with expired headings


jebw (approved):
Why weren't the specs failing before?

willfish:
The scenario that wasn't covered was fetching the goods nomenclature class when the heading was also expired.

PR #854, opened 2022-11-02
HOTT-2071: Exposed verbose duty field on backend API

maverickvi:
### Jira link

[HOTT-<2071>](https://transformuk.atlassian.net/browse/HOTT-2071)

### What?

I have added/removed/altered:

- [ ] Exposed verbose duty field on backend API

### Why?

I am doing this because:

- We need to access it on the frontend

### Have you? (optional)

Screenshot 2022-11-02 at 17 12 57


jebw (approved):
LGTM

app/models/measure.rb
```diff
@@ -393,7 +393,10 @@ def verbose_duty_expression
   end
 
   def prettify_generated_duty_expression(duty_expression)
-    duty_expression.sub(/\s%/, '%').sub(/\s\s/, ' ')
+    duty_expression.sub!(/\s\s/, ' ')
+    duty_expression.sub!(/\s%/, '%') if duty_expression.scan(/\d\s%/).present?
```
jebw, line 397: I believe this can be simplified with regex captures

```ruby
duty_expression.sub!(/(\d)\s%/, '\1%')
```

app/models/measure.rb
```diff
@@ -393,7 +393,10 @@ def verbose_duty_expression
...
+    duty_expression.sub!(/\/\s[a-zA-Z]/, &:downcase) if duty_expression.scan(/\/\s[a-zA-Z]/).present?
```
jebw, line 398: This `if` condition seems redundant? if there's no match then no substitution will occur anyway?

app/models/measure.rb
```diff
@@ -393,7 +393,10 @@ def verbose_duty_expression
...
+    duty_expression
```
jebw, line 399: In general, mutating a passed in variable can often lead to later bugs when another developer consumes a method without expecting it to mutate the data, particularly in this case where it both mutates the variable passed in, then returns the mutated form anyway.

Ruby in general is moving away from mutating strings with the push towards frozen strings literals (as seen in other languages)

I'd suggest either appending the method name with an exclamation mark to show its going to mutate data (eg difference between `'#sub` and `#sub!`. Or alternatively with the suggestions above, you can drop the `if` conditions and just chain the substitutions

jebw (approved):
LGTM - I've added some syntactic suggestions - take or leave as you feel appropriate :+1:

PR #859, opened 2022-11-08
HOTT-2212 Alert slack when missing CDS file

jebw:
### Jira link

HOTT-2212

### What?

I have added/removed/altered:

- [x] Added a notification to the slack channel when a CDS file has not shown up

### Why?

I am doing this because:

- We were missing a file from sunday even, and only realised sync didn't happen yesterday when todays file showed up but was rejected for being out of order

- Impacts overnight sync - should be seen to work DEV before merging, then should be seen to work on staging before release

### DIFF
--- a/app/workers/updates_synchronizer_worker.rb
+++ b/app/workers/updates_synchronizer_worker.rb
@@ -13,15 +13,10 @@ class UpdatesSynchronizerWorker
     if TradeTariffBackend.uk?
       TariffSynchronizer.download_cds
 
-      if check_for_todays_file && todays_file_has_not_yet_arrived?
-        if still_time_to_reschedule?
-          self.class.perform_in(TRY_AGAIN_IN, true)
-          logger.info "Daily file missing, retrying at #{TRY_AGAIN_IN.from_now}"
-          return
-        else
-          SlackNotifierService.call \
-            'Daily CDS file missing, max retry time passed - continuing without todays file'
-        end
+      if check_for_todays_file &&
+          todays_file_has_not_yet_arrived? &&
+          attempt_reschedule!
+        return
       end
 
       logger.info 'Applying...'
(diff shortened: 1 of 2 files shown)


app/workers/updates_synchronizer_worker.rb
```diff
@@ -13,13 +13,15 @@ def perform(check_for_todays_file = true, reapply_data_migrations = false)
...
-        return
+      if check_for_todays_file && todays_file_has_not_yet_arrived?
+        if still_time_to_reschedule?
+          self.class.perform_in(TRY_AGAIN_IN, true)
+          logger.info "Daily file missing, retrying at #{TRY_AGAIN_IN.from_now}"
+          return
+        else
+          SlackNotifierService.call \
+            'Daily CDS file missing, max retry time passed - continuing without todays file'
+        end
```
alexdesi, line 24: The logic looks perfect here. It would be nice to remove another if level.

A way could be moving the login in a method called "reschedule!", something like this:
 
`reschedule! and return` 

```
def reschedule!
        if still_time_to_reschedule?
          self.class.perform_in(TRY_AGAIN_IN, true)
          logger.info "Daily file missing, retrying at #{TRY_AGAIN_IN.from_now}"

          true # it was rescheduled
        else
          SlackNotifierService.call \
            'Daily CDS file missing, max retry time passed - continuing without todays file'

          false # no reschedule is possible
        end
end
```

alexdesi (approved):
It looks good to me! ✅ 
My only concern is nested IFs:
we have 4 levels now, which adds complexity (even if it is totally correct!).

maverickvi (approved):

app/workers/updates_synchronizer_worker.rb
jebw: Good call - now simplified :+1:

PR #862, opened 2022-11-14
Fix for undefined method role.

alexdesi:
### Jira link
HOTT-2029

### What?
This PR fixes the undefined `role` method.

### Why?
Because we are  🐛 🐛🐛 fighters!  🐛🔫

### DIFF
--- a/app/models/full_temporary_stop_regulation.rb
+++ b/app/models/full_temporary_stop_regulation.rb
@@ -19,4 +19,8 @@ class FullTemporaryStopRegulation < Sequel::Model
   def effective_start_date
     validity_start_date.to_date
   end
+
+  def role
+    full_temporary_stop_regulation_role
+  end
 end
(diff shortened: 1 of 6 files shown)


spec/models/measure_partial_temporary_stop.rb
```diff
@@ -0,0 +1,11 @@
+RSpec.describe MeasurePartialTemporaryStop do
+  let(:measure_partial_temporary_stop) { build :measure_partial_temporary_stop }
+
+  describe '#regulation_id' do
+    it { expect(measure_partial_temporary_stop.regulation_id).to be_a String }
```
jebw, line 5: Can this line check the value is the expected one, ie

```ruby
    it { expect(measure_partial_temporary_stop.regulation_id).to eql measure_partial_temporary_stop.partial_temporary_stop_regulation_id }
```

jebw (changes requested):
LGTM - just one minor comment on the spec for the existing method, nice to see backfilling of specs  :raised_hands:

jebw (approved):
LGTM :raised_hands:

PR #870, opened 2022-11-15
Fix delegation error.

alexdesi:
### Jira link
Bug fix.

### What?
Allow nil for the delegation on duty expression description.

### DIFF
--- a/app/models/measure_component.rb
+++ b/app/models/measure_component.rb
@@ -27,7 +27,7 @@ class MeasureComponent < Sequel::Model
   one_to_one :measure, key: :measure_sid,
                        primary_key: :measure_sid
 
-  delegate :description, :abbreviation, to: :duty_expression, prefix: true
+  delegate :description, :abbreviation, to: :duty_expression, prefix: true, allow_nil: true
   delegate :abbreviation, to: :monetary_unit, prefix: true, allow_nil: true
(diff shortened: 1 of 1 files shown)


alexdesi:
Screenshot 2022-11-15 at 14 55 46

alexdesi:
As discussed with @jebw , the root cause of this issue was the lack of Timemachine in the measure controller#show.
Fixed now!

spec/requests/api/v2/measures_controller_spec.rb
```diff
@@ -12,5 +12,13 @@
     end
 
     it_behaves_like 'a successful jsonapi response'
+
+    it 'the TimeMachine receives the correct Date' do
+      allow(TimeMachine).to receive(:at).and_call_original
+
+      make_request
+
+      expect(TimeMachine).to have_received(:at).with(Time.zone.today)
```
jebw, line 21: This is testing time machine rather then external behaviour and I'm pretty sure would pass even without the `.actual` change to the controller. 

I'd recommend a spec like this instead

```ruby
context 'for historical measure' do
  let(:measure) { create(:measure, validity_end_date: 2.days.ago) }

  it { is_expected.to have_http_status :not_found }
end
```

jebw (changes requested):

jebw (approved):
LGTM :raised_hands:

PR #871, opened 2022-11-16
HOTT-2130 expose news attributes to admin

jebw:
### Jira link

HOTT-2130

### What?

I have added/removed/altered:

- [x] Expose new News::Item attributes in the admin API

### Why?

I am doing this because:

- we wish to allow them to be updated from the admin area

- Low, affects an admin only api


app/models/news/item.rb
```diff
@@ -31,9 +31,11 @@ def before_validation
     end
 
     def after_save
-      (collection_ids - collections.pluck(:id)).each(&method(:add_collection))
-      (collections.pluck(:id) - (collection_ids & collections.pluck(:id)))
-        .each(&method(:remove_collection))
+      to_add = (collection_ids - collections.pluck(:id))
```
alexdesi, line 34: nitpicking: we can remove parenthesis here
`to_add = collection_ids - collections.pluck(:id)`

app/models/news/item.rb
```diff
@@ -31,9 +31,11 @@ def before_validation
...
+      to_add.each(&method(:add_collection))
+
+      to_remove = (collections.pluck(:id) - (collection_ids & collections.pluck(:id)))
```
alexdesi, line 37: This is cryptic:
it excludes from `collections.pluck(:id)` the  intersection between `collections_ids`  and the `collections.pluck(:id)`
If I am not wrong this should be the same as:

`collections.pluck(:id) - collection_ids`

jebw: You're right - I'm forgetting removing elements that aren't there is ignored - i was thinking it would raise an error, I'll change that

app/models/news/item.rb
jebw: lol - normally you're asking me to add the parentheses :smile: I'll get that changed

alexdesi (approved):
nice one! ✅

PR #873, opened 2022-11-17
Add more role to abrogation regulations

alexdesi:
### What?
Add method role for complete_abrogation_regulation and explicit_abrogation_regulation

### Why?
To fulfil the interface implemented in all kinds of regulation.


jebw (approved):
LGTM :raised_hands:

PR #883, opened 2022-11-22
HOTT-2248: Fixed bug on XI caused by monetary unit code not existing …

maverickvi:
…in currency hash

### Jira link

[HOTT-<2248>](https://transformuk.atlassian.net/browse/HOTT-2248)

### What?

I have added/removed/altered:

- [ ] Added EUC to currency hash and Else case

### Why?

I am doing this because:

- EUC monetary unit code was causing errors because it was not present in the currency hash

Screenshot 2022-11-22 at 13 59 27


jebw (approved):
LGTM

PR #885, opened 2022-11-28
HOTT-2136 update news item api

jebw:
### Jira link

HOTT-2136

### What?

I have added/removed/altered:

- [x] Added priority to News/Collection entity
- [x] Added description to News/Collection entity

### Why?

I am doing this because:

- We need to sort some collections above others
- We need to show additional content in the footer of the news stories relating to the collection

- Includes a database schema change

### DIFF
--- a/spec/models/news/item_spec.rb
+++ b/spec/models/news/item_spec.rb
@@ -66,7 +66,11 @@ RSpec.describe News::Item do
 
         let(:priority) { create(:news_collection, name: 'CCC', priority: 1) }
 
-        it { is_expected.to eq [priority.id] + collections.map(&:id).reverse }
+        let :high_then_low_priority_collection_ids do
+          [priority.id] + collections.map(&:id).reverse
+        end
+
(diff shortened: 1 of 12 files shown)


spec/models/news/item_spec.rb
```diff
@@ -60,6 +60,14 @@
       end
 
       it { is_expected.to eq collections.map(&:id).reverse }
+
+      context 'with priorities' do
+        before { item.add_collection priority }
+
+        let(:priority) { create(:news_collection, name: 'CCC', priority: 1) }
+
+        it { is_expected.to eq [priority.id] + collections.map(&:id).reverse }
```
alexdesi, line 69: This test is a bit cryptic,
what are the other priority ids?
should we check also the name attr here? (in case of duplicated id?)

maverickvi (approved):

spec/models/news/item_spec.rb
jebw: Moved to a `let` block to assign a name

PR #891, opened 2022-11-29
Clean up/Refactor spec using TimeMachine

alexdesi:
### What?
Removed specs related to the TimeMachine in the controlles: they are not effective, even removing `.actual` from the queries in the controller, specs continue to pass.

- Add specs that actually test when the validity_end_date is set to a past date.

### Jira link
No Jira

### DIFF
--- a/spec/controllers/api/v2/certificates_controller_index_spec.rb
+++ b/spec/controllers/api/v2/certificates_controller_index_spec.rb
@@ -34,12 +34,10 @@ RSpec.describe Api::V2::CertificatesController, type: :controller do
 
     it { is_expected.to match_json_expression(pattern) }
 
-    it 'the TimeMachine receives the correct Date' do
-      allow(TimeMachine).to receive(:at).and_call_original
+    context 'when the validity_end_date is set to a past date' do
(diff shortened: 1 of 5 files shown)


spec/controllers/api/v2/certificates_controller_index_spec.rb
```diff
@@ -34,12 +34,16 @@
 
     it { is_expected.to match_json_expression(pattern) }
 
-    it 'the TimeMachine receives the correct Date' do
-      allow(TimeMachine).to receive(:at).and_call_original
+    context 'when the validity_end_date is set to a past date' do
+      let(:certificate) { create(:certificate, validity_end_date: 1.day.ago) }
 
-      do_response
+      let(:empty_response_pattern) do
```
jebw, line 40: This could be a 1 liner instead of 5?

jebw (approved):
LGTM - one line could be written as a 1-liner but its certainly clear enough as is. Thanks for changing these :raised_hands:

spec/controllers/api/v2/certificates_controller_index_spec.rb
alexdesi: yeah, actually we can even fully get rid of the let.

PR #892, opened 2022-11-30
HOTT-2129 import historical news stories

jebw:
### Jira link

HOTT-2129

### What?

I have added/removed/altered:

- [x] Added an importer + dataset for historical trade stories from GovUK site

### Why?

I am doing this because:

- we are moving the source of information from the govuk site to our service

- Creates additional news stories


app/models/news/importer.rb
```diff
@@ -0,0 +1,138 @@
+module News
+  class Importer
+    DEFAULT_ITEMS_FILE = Rails.root.join('db/news/govuk_stories.json').freeze
```
alexdesi, line 3: nitpicking: maybe DEFAULT_ITEMS_FILE**NAME** ?

app/models/news/importer.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    def import!
+      raise NotAvailableOnXi if TradeTariffBackend.xi?
+
+      import_time = Time.zone.now
+
+      ::News::Item.db.transaction do
+        create_collections
+
+        stories.each { |story| import_story(story, import_time) }.length
```
alexdesi, line 36: Do you want to return the number of stories imported?

My understanding is that returns the object it was invoked upon.
therefore `stories.length` is independent of the success/failure of the import stories.

app/models/news/importer.rb
```diff
@@ -0,0 +1,138 @@
...
+        display_style: News::Item::DISPLAY_REGULAR,
+        show_on_uk: true,
+        show_on_xi: true,
+        show_on_updates_page: true,
+        show_on_banner: false,
+        show_on_home_page: false,
+        imported_at: import_time,
+      )
+
+      get_collections(item_data['themes']).each do |collection|
```
alexdesi, line 64: what is 'collection'?
maybe a story_collection (not sure) 🤔

app/models/news/importer.rb
```diff
@@ -0,0 +1,138 @@
...
+        item.add_collection collection
+      end
+    end
+
+    def data_has_all_keys?(story)
+      NEWS_ITEM_KEYS.all? { |k| story.key?(k) }
+    end
+
+    def get_collections(names)
+      names.split(',').map(&method(:get_collection))
```
alexdesi, line 74: I guess the simpler version does not work:
`.map(&:get_collection)`

doesn't it?

spec/models/news/importer_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+  let(:json_data) { file_fixture 'news/govuk_stories.json' }
+
+  describe '.new' do
+    it { is_expected.to be_instance_of described_class }
+  end
+
+  describe '#stories' do
+    subject { instance.stories }
+
+    it { is_expected.to have_attributes length: 3 }
```
alexdesi, line 15: It would be nice to check the names of the attributes.

spec/models/news/importer_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+  end
+
+  describe '#import!' do
+    it { expect { instance.import! }.to change(News::Item, :count).by(3) }
+    it { expect(instance.import!).to be 3 }
```
alexdesi, line 20: Instance seems to return the number of stories,
and not those that are imported.

spec/models/news/importer_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+    end
+
+    context 'with collection which does not exists' do
+      subject { News::Collection.all.pluck(:name) }
+
+      before { instance.import! }
+
+      let(:json_data) { StringIO.new incomplete_data.to_json }
+
+      let :incomplete_data do
```
alexdesi, line 69: Which piece of data is missing?
if only one field is missing,  we could  specify which one is missing.  

For example:
let(:news_with_missing_slug)

spec/models/news/importer_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+      context 'with collections' do
+        subject { first_item.collections }
+
+        it { is_expected.to have_attributes length: 1 }
+        it { is_expected.to all have_attributes name: 'Tariff notices' }
+      end
+    end
+
+    context 'with collection which does not exists' do
```
alexdesi, line 62: maybe just:
'when the collection does not exist'

maverickvi (approved):

alexdesi (commented):
Good implementation!
My only concern is the return value of the `import!` method,
the others are just minor comments.

app/models/news/importer.rb
jebw: any story which fails to import will bomb the transaction - so the number of stories imported _has_ to match the length of `.stories` - it _cannot_ be different?

app/models/news/importer.rb
jebw: thats a different method - long hand of mine is

`names.split(',').map { |name_string| self.get_collection(name_string) }`

long hand of your version would be

`names.split(',').map { |name_string| name_string.get_collection }`

So wouldn't work?

spec/models/news/importer_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+      context 'with collections' do
+        subject { first_item.collections }
+
+        it { is_expected.to have_attributes length: 1 }
+        it { is_expected.to all have_attributes name: 'Tariff notices' }
+      end
+    end
+
+    context 'when collection does not exists' do
```
alexdesi, line 62: exists -> exist

alexdesi (approved):
🚀 great, thanks for the changes!

PR #894, opened 2022-11-30
HOTT-2278 dont 404 for news items for unknown collection

jebw:
We should return an empty list instead of 404

### Jira link

HOTT-2278

### What?

I have added/removed/altered:

- [x] Filter by collection id instead of looking for the corresponding collection and finding its items

### Why?

I am doing this because:

- Previously an invalid collection id would return a 404 rather then no news items

- Changes behaviour of the key news item api but in a manner which should make its usage safer


willfish (approved):
Makes sense!

PR #895, opened 2022-12-01
HOTT-2275: Fixes copy/paste error in the admin commodities controller

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2275

### What?

I have added/removed/altered:

- [x] Altered incorrect reference to serializer in admin commodities controller
- [x] Added some missing specs that would have caught this

### Why?

I am doing this because:

- This is required so we can properly manage search references


jebw (approved):
LGTM

PR #901, opened 2022-12-02
Hott 2296 Add slug and published fields to collections

jebw:
### Jira link

HOTT-2296

### What?

I have added/removed/altered:

- [x] Added a slug field to news collections
- [x] Allow filtering collection slug as well as collection id
- [x] Added a published boolean field to news collections
- [x] Added a `published_collections` relationship which is a filtered list of collections
- [x] Changed the frontend news collections api to only expose published collections
- [x] Changed the frontend news items api to only include published collections
- [x] Changed the frontend news item filtering to only allow filtering on published collections

### Why?

I am doing this because:

- We need to hide all `tariff_notices` stories
- We want to use slugs for the urls when filtering stories

- Includes data migrations which change the news stories data in use in production

### DIFF
--- a/app/controllers/api/v2/news/items_controller.rb
+++ b/app/controllers/api/v2/news/items_controller.rb
@@ -22,7 +22,9 @@ module Api
                                                     .descending
                                                     .all
 
-          serializer = Api::V2::News::ItemSerializer.new(news_items_with_collections,
+          presented_news_items = Api::V2::News::ItemPresenter.wrap(news_items_with_collections)
+
+          serializer = Api::V2::News::ItemSerializer.new(presented_news_items,
                                                          include: %i[collections],
                                                          meta: pagination_meta(news_items_page))
 
@@ -40,7 +42,9 @@ module Api
                         scope.with_pk!(slug_or_id)
                       end
 
-          serializer = Api::V2::News::ItemSerializer.new(news_item, include: %i[collections])
+          presented_news_item = Api::V2::News::ItemPresenter.new(news_item)
+
+          serializer = Api::V2::News::ItemSerializer.new(presented_news_item, include: %i[collections])
 
           render json: serializer.serializable_hash
         end
--- a/app/controllers/api/v2/news/items_controller.rb
+++ b/app/controllers/api/v2/news/items_controller.rb
@@ -12,6 +12,7 @@ module Api
                                    .for_collection(params[:collection_id])
                                    .for_today
                                    .descending
+                                   .distinct
                                    .select { news_items[:id] }
                                    .paginate(current_page, per_page)
 
--- a/app/controllers/api/v2/news/items_controller.rb
+++ b/app/controllers/api/v2/news/items_controller.rb
@@ -13,11 +13,11 @@ module Api
                                    .for_today
                                    .descending
                                    .distinct
-                                   .select { news_items[:id] }
+                                   .select { [news_items[:id], news_items[:start_date]] }
                                    .paginate(current_page, per_page)
 
           # Why? you may ask - because #paginate ignores #eager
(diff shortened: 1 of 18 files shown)


app/models/news/item.rb
```diff
@@ -8,6 +8,11 @@ class Item < Sequel::Model(:news_items)
 
     many_to_many :collections, join_table: :news_collections_news_items
 
+    many_to_many :published_collections, join_table: :news_collections_news_items,
+                                         conditions: { published: true },
```
willfish, line 12: TIL :)

app/models/news/item.rb
```diff
@@ -45,7 +50,7 @@ def validate
 
     dataset_module do
       def descending
-        order(Sequel.desc(:start_date), Sequel.desc(:id))
+        order { [Sequel.desc(news_items[:start_date]), Sequel.desc(news_items[:id])] }
```
willfish, line 53: Is this not the same as before but just passing the configuration to an eval'd block?

db/data_migrations/20221202133634_assign_existing_news_to_trade_news.rb
```diff
@@ -0,0 +1,19 @@
+Sequel.migration do
+  # IMPORTANT! Data migrations should be Idempotent, they may get re-run as part
+  # of data rollbacks
+
+  up do
+    trade_news = News::Collection.find_or_create(name: 'Trade news') do |collection|
```
willfish, line 6: I'm wondering whether this should be a rake task if we don't want it to be rolled back ever?

willfish: And maybe add an explicit callback to force all items to have the trade_news collection?

db/migrate/20221202094157_add_slug_to_collections.rb
```diff
@@ -0,0 +1,8 @@
+Sequel.migration do
+  change do
+    alter_table :news_collections do
+      add_column :slug, String, size: 255, unique: true
+      add_column :published, :boolean, default: true
```
willfish, line 5: Should we defaultingly not publish new collections?

app/controllers/api/v2/news/items_controller.rb
```diff
@@ -12,30 +12,39 @@ def index
...
 
           # Why? you may ask - because #paginate ignores #eager
-          news_items_with_collections = ::News::Item.eager(:collections)
+          news_items_with_collections = ::News::Item.eager(:published_collections)
                                                     .where(id: news_items_page.pluck(:id))
                                                     .descending
                                                     .all
 
-          serializer = Api::V2::News::ItemSerializer.new(news_items_with_collections,
+          presented_news_items = Api::V2::News::ItemPresenter.wrap(news_items_with_collections)
```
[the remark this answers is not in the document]
jebw: :smile: Was trying to come up with an obvious name

app/models/news/item.rb
jebw: originally I was running into issues because the first syntax doesn't assign table names where as the second doesn't but looking at it again it does see unneeded so I'll test again and see if it is necessary - was solving this one right at the end of friday so was pretty fried :smile:

db/data_migrations/20221202133634_assign_existing_news_to_trade_news.rb
jebw: There's a separate PR for that - may roll out tomorrow but more likely on thursday once I know the data is straight

As regards rake task - would normally agree _but_ we'll end up with all news stories disappearing from the site including the home page block because they come to depend upon belonging to a published collection

PR #906, opened 2022-12-06
HOTT-2320: Downcase keys for intercept messages

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2320

### What?

I have added/removed/altered:

- [x] Make sure all intercept message keys are lower cased

### Why?

I am doing this because:

- This is required to work with the intercept message builder which only works iwth lower case search query strings

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -291,7 +291,7 @@ en:
     title: "aesthetic"
     message: "Based on your search term, we believe you are looking for anaesthetic, either anaesthetic medicaments of heading 3003 or heading 3004 or anaesthetic equipment of heading 9018."
 
-  Afghanistan:
+  afghanistan:
     title: "Afghanistan"
     message: "See more information about trading with [Afghanistan](https://www.gov.uk/world/organisations/department-for-international-trade-afghanistan)."
 
@@ -523,7 +523,7 @@ en:
     title: "algamating tape"
     message: "Based on your search term, we believe you are looking for subheading 391910, depending on the constituent material."
 
-  Algeria:
+  algeria:
     title: "Algeria"
     message: "See more information about trading with [Algeria](https://www.gov.uk/world/organisations/department-for-international-trade-algeria)."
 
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -6951,7 +6951,7 @@ en:
     title: "dome camera"
     message: "Based on your search term, we believe you are looking for heading 9006, then the full commodity code depends on functionality."
 
-  Dominican Republic:
+  dominican Republic:
```
jebw, line 6954: `Republic` should be `republic`

config/locales/en.yml
```diff
@@ -13223,27 +13223,27 @@ en:
...
+  netherlands:
     title: "Netherlands"
     message: "See more information about trading with [Netherlands](https://www.gov.uk/world/organisations/department-for-international-trade-netherlands)."
 
   new clothes:
     title: "new clothes"
     message: "New clothes are classified under chapter 61, chapter 62 or chapter 63. Classification depends on the type of garment, whether the item is knitted or woven, the sex of the garment and the type of textile, unisex garments are classified under women's clothing if there are no distinguishable left-over-right fastenings."
 
-  New Zealand:
+  new Zealand:
```
jebw, line 13238: `Zealand` => `zealand`

jebw (changes requested):
Only downcased first first in multiword keys

Related question - I'm assuming the normalisation is to aid lookup via key. If so, are multiword keys going to be problematic as well - would it make sense to normalise whitespace (drop?) as well?

willfish:
My derp, I didn't check the result/just assumed my regex was okay:

grep for key lines, apply a replace on the start of every words first char and change to lowercase recursively  (`:g/:$/s/\<./\l&/g`)

willfish:
Multiword keys are fine/handled properly by rails. I'll do a quick test when this deploys

jebw (approved):
LGTM

PR #907, opened 2022-12-06
HOTT-2322: Redirect single digit searches to chapter

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2322

### What?

I have added/removed/altered:

- [x] Added handling of single digit searches

### Why?

I am doing this because:

- Its simple enough to handle this here even if technically the short code is in the wrong format
- Discovered as part of performance testing here: https://sentry.io/organizations/engine-le/issues/3790300560/?project=5557002&referrer=slack

### DIFF
--- a/app/models/beta/search/open_search_result.rb
+++ b/app/models/beta/search/open_search_result.rb
@@ -174,6 +174,9 @@ module Beta
         id = short_code
 
         resource_path = case short_code.length
+                        when 1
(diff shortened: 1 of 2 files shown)


jebw (approved):
LGTM - 

this would actually have solved some bad data we had for RoO wizard - I'd fixed the data but this is more flexible

PR #910, opened 2022-12-06
HOTT-2308: Remove space before percentage signs on all commodity code…

maverickvi:
… formatted descriptions

The String contained a non standard space character which I had to replace with a standard space character because regex replace wasn't working. This can be checked by doing formatted_description.chars[-6].ord

Useful links:
https://stackoverflow.com/questions/48174592/change-all-nbsp-to-normal-blank-space-ruby
https://apidock.com/rails/String/ord

### Jira link

[HOTT-<2308>](https://transformuk.atlassian.net/browse/HOTT-2308)

### What?

I have added/removed/altered:

- [ ] Remove space before percentage signs on all commodity code formatted descriptions

### Why?

I am doing this because:

- We don't want a space before percentage signs

Screenshot 2022-12-06 at 14 25 41

Screenshot 2022-12-06 at 14 27 20

### DIFF
--- a/app/formatters/description_formatter.rb
+++ b/app/formatters/description_formatter.rb
@@ -6,6 +6,8 @@ class DescriptionFormatter
     return '' if str.blank?
 
     str.gsub!('|%', '%')
+    str.gsub!(160.chr("UTF-8")," ")
+    str.gsub!(/(\d)\s+%/, '\1%')
     str.gsub!('-|', "\n-")
     str.gsub!('|', '&nbsp;')
     str.gsub!('!1!', '<br />')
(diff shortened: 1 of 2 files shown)


app/formatters/description_formatter.rb
```diff
@@ -6,6 +6,8 @@ def self.format(opts = {})
     return '' if str.blank?
 
     str.gsub!('|%', '%')
+    str.gsub!(160.chr('UTF-8'), ' ')
```
willfish, line 9: Is this a non-breaking space?

willfish: TIL

willfish (approved):

PR #916, opened 2022-12-09
HOTT-2241 Fixed removing stories belonging to a collection

jebw:
### Jira link

HOTT-2241

### What?

I have added/removed/altered:

- [x] Fixed removing news stories which belong to 1 or more collections

### Why?

I am doing this because:

- All stories belong to collections, so no stories could be removed

- Low, bugfix

### DIFF
--- a/app/models/news/item.rb
+++ b/app/models/news/item.rb
@@ -42,6 +42,10 @@ module News
       to_remove.each(&method(:remove_collection))
     end
 
+    def before_destroy
+      collections.pluck(:id).each(&method(:remove_collection))
+    end
+
     def validate
       super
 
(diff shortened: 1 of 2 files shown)


app/models/news/item.rb
```diff
@@ -42,6 +42,10 @@ def after_save
       to_remove.each(&method(:remove_collection))
     end
 
+    def before_destroy
+      collections.pluck(:id).each(&method(:remove_collection))
```
willfish, line 46: Can we not use https://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/AssociationDependencies.html?

PR #921, opened 2022-12-12
HOTT-2297: Adds new transfer events to quota definition

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2297

### What?

I have added/removed/altered:

- [x] Added transfer events to commodity api responses

### DIFF
--- a/app/serializers/api/v2/quotas/quota_closed_and_transferred_event_serializer.rb
+++ b/app/serializers/api/v2/quotas/quota_closed_and_transferred_event_serializer.rb
@@ -12,7 +12,20 @@ module Api
           event.transferred_amount.try(:to_f)
         end
 
-        has_one :quota_definition, serializer: Api::V2::QuotaOrderNumbers::QuotaDefinitionSerializer
+        attribute :target_quota_definition_validity_start_date do |event|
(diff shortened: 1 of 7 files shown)


app/serializers/api/v2/quotas/quota_closed_and_transferred_event_serializer.rb
```diff
@@ -12,7 +12,20 @@ class QuotaClosedAndTransferredEventSerializer
           event.transferred_amount.try(:to_f)
         end
 
-        has_one :quota_definition, serializer: Api::V2::QuotaOrderNumbers::QuotaDefinitionSerializer
```
willfish, line 15: We have issues with recursing over measures and then definitions and then this event and back again in the parser on the frontend if we try and be good boys with jsonapi

willfish: I'd need to lazily load related resources from an accumulating graph of type/id includes with attributes if I wanted to fix this in the frontend which is quite a significant change

maverickvi (approved):

PR #932, opened 2022-12-15
HOTT-2374: Removes unmerged cluster option

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2374

### What?

I have added/removed/altered:

- [x] Move to always using separate index names for xi/uk backends

### Why?

I am doing this because:

- We're never going to go back to the 2k pm more costly implementation


jebw (approved):
LGTM

PR #935, opened 2022-12-16
HOTT-2361 Reduce sql queries for admin headings api

jebw:
Still has N+1 but should be ~1N rather then 3.5N

### Jira link

HOTT-2361

### What?

I have added/removed/altered:

- [x] Reduce number of queries for search references page

### Why?

I am doing this because:

- The page times out at the moment
- This isn't great tbh as there are still way to many queries but it does reduce the number and _may_ be enough to get things working in staging

- Limited - changes admin api only


app/serializers/api/admin/headings/commodity_serializer.rb
```diff
@@ -16,7 +16,7 @@ class CommoditySerializer
           if commodity.declarable?
             commodity.search_references.count
           else
-            Subheading.find(goods_nomenclature_sid: commodity.goods_nomenclature_sid).search_references.count
+            Subheading.call(commodity.values).search_references.count
```
willfish, line 19: TIL call :)

willfish (approved):
Awesome sauce :)

PR #936, opened 2022-12-19
HOTT-2373: Make intercept message search terms searchable 

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2373

### What?

I have added/removed/altered:

- [x] Added support for searching through intercept terms
- [x] Added support for generating references and injecting their corresponding search terms into the beta index
- [x] Fixed passing issue with heading link generation

### Why?

I am doing this because:

- This is required so that we can use the intercept messages as part of search at the subheading and commodity levels.
- Intercept messages have valuable hand-crafted meaning in them

### DIFF
--- a/app/models/beta/search/intercept_message.rb
+++ b/app/models/beta/search/intercept_message.rb
@@ -14,34 +14,46 @@ module Beta
           match = matched_text.match(SECTION_REGEX)
 
           section_id = RomanNumerals::Converter.to_decimal(match[:code])
+          reference_id = RomanNumerals::Converter.to_roman(section_id)
+          section_text = 'section '
+          section_text += "#{match[:optional]} " if match[:optional]
+          section_text += match[:code]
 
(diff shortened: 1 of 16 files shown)


app/models/beta/search/intercept_message.rb
```diff
@@ -62,30 +78,52 @@ def build(search_query)
...
+              merged.scan(/\w+/).join(' ').downcase
+            end
+          end
         end
 
         private
 
         def normalise_query(search_query)
-          search_query.downcase.scan(/\w+/).join(' ')
+          search_query.downcase.scan(/\w+/).reject { |term| TradeTariffBackend.stop_words.include?(term) }.join(' ')
```
willfish, line 101: Extend test example to include some stop words that get removed/matched on an intercept message

maverickvi (approved):

PR #938, opened 2022-12-19
HOTT-2378: Search reference duplication enhancements

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2377
https://transformuk.atlassian.net/browse/HOTT-2378

### What?

I have added/removed/altered:

- [x] Make search suggestions distinct
- [x] Stop redirecting when there are multiple search reference titles matching the search query

### Why?

I am doing this because:

- This is required to handle duplicate title search references better

### DIFF
--- a/spec/services/api/beta/search_service_spec.rb
+++ b/spec/services/api/beta/search_service_spec.rb
@@ -1,17 +1,17 @@
 RSpec.describe Api::Beta::SearchService do
   # rubocop:disable RSpec/MultipleMemoizedHelpers
   describe '#call' do
-    subject(:call) { described_class.new(search_query, filters: {}, spell: '1').call }
+    subject(:search_result) { described_class.new(search_query, filters: {}, spell: '1').call }
 
     context 'when the search query is an empty string' do
       let(:search_query) { '' }
 
-      it { expect(call.empty_query).to eq(true) }
+      it { expect(search_result.empty_query).to eq(true) }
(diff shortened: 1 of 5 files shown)


spec/services/api/beta/search_service_spec.rb
```diff
@@ -103,6 +103,31 @@
...
+          ),
+        )
+        allow(TradeTariffBackend.v2_search_client).to receive(:search).and_return(search_result)
+        allow(Api::Beta::SearchQueryParserService).to receive(:new).and_return(
+          instance_double('Api::Beta::SearchQueryParserService', call: build(:search_query_parser_result)),
+        )
+        allow(Beta::Search::GoodsNomenclatureQuery).to receive(:build).and_return(build(:goods_nomenclature_query))
+        allow(Beta::Search::OpenSearchResult::WithHits).to receive(:build).and_call_original
+
+        call
```
jebw, line 126: After some initial confusion this is not a `call` on the current object but a named subject. I feel `call` is sufficiently generic confuse other developers

jebw (approved):

spec/services/api/beta/search_service_spec.rb
willfish: I reflected on this quite a bit and looked at other services and decided I agree. This is also not the first time I've had this conversation so yeah

PR #942, opened 2022-12-20
HOTT-2390: Add support for multiple transfer events

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2390

### What?

I have added/removed/altered:

- [x] Added support for handling multiple transfer events per definition
- [x] Added ability to review skipped transfer events

### Why?

I am doing this because:

- The daily files can include transfer out and transfer in events. I didn't notice this until this morning since most files only have one event per quota definition primary node.
- I wanted the skipped feedback just for development purposes the ability to review skipped transfer nodes

### AC

- [x] Reviewed 100% of imported/skipped events from today's file and all looks good

### DIFF
--- a/app/lib/cds_importer.rb
+++ b/app/lib/cds_importer.rb
@@ -95,8 +95,8 @@ class CdsImporter
         oplog_inserts[:operations][operation][entity_class][:duration] += duration
         oplog_inserts[:operations][operation][entity_class][:mapping_path] = mapping_path
 
-        # We only accumulate missing destroy operations because we can work out from the file which record was inserted for non-missing operation types
-        if [CdsImporter::RecordInserter::DESTROY_MISSING_OPERATION, :skipped].include?(operation)
+        # We only accumulate missing destroy and skipped operations because we can work out from the file which record was inserted for non-missing operation types
+        if [CdsImporter::RecordInserter::DESTROY_MISSING_OPERATION, CdsImporter::RecordInserter::SKIPPED_OPERATION].include?(operation)
           oplog_inserts[:operations][operation][entity_class][:records] ||= []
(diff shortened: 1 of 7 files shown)


app/lib/cds_importer.rb
```diff
@@ -95,7 +96,7 @@ def subscribe_to_oplog_inserts
         oplog_inserts[:operations][operation][entity_class][:mapping_path] = mapping_path
 
         # We only accumulate missing destroy operations because we can work out from the file which record was inserted for non-missing operation types
-        if operation == CdsImporter::RecordInserter::DESTROY_MISSING_OPERATION
+        if [CdsImporter::RecordInserter::DESTROY_MISSING_OPERATION, :skipped].include?(operation)
```
jebw, line 99: Minor and certainly doesn't block this PR but is it worth putting `:skipped` as a constant on RecordInserter at some point?

app/lib/cds_importer/record_inserter.rb
```diff
@@ -60,6 +60,10 @@ def save_record(xml_key)
       nil
     end
 
+    def skip_record
```
jebw, line 63: `instrument_skip_record` or `log_record_skipped` ? The name implies an action (`skip`) but its actually logging an action which will happen anyway even without this method call

jebw (approved):
LGTM :raised_hands:

app/lib/cds_importer.rb
willfish: It is for sure. I'll do that

app/lib/cds_importer/record_inserter.rb
willfish: I thought this on my own reviewing of the code, yeah. I'll rename to instrument

PR #946, opened 2022-12-20
HOTT-2394: Adjustments for new guides files

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2394

### What?

I have added/removed/altered:

- [x] Added adjustments to import new guides files

### Why?

I am doing this because:

- This will unblock running the rake task


jebw (approved):
LGTM

PR #954, opened 2023-01-03
HOTT-2408: Remove vanuatu from VAT group

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2408

### What?

I have added/removed/altered:

- [x] Added data migration to remove Vanuatu from the VAT geographical area group

### Why?

I am doing this because:

- This is an urgent data fix until CDS produce a file with this adjustment


jebw (changes requested):
Has this been tested - there are 123 entries that day? and this only removes the first?

PR #960, opened 2023-01-04
HOTT-2415: Updated Chief CDS guidance file

maverickvi:
### Jira link

[HOTT-<TODO>](https://transformuk.atlassian.net/browse/HOTT-2415)

### What?

I have added/removed/altered:

- [ ] Updated Chief CDS guidance file

### Why?

I am doing this because:

- previous version was outdated

Screenshot 2023-01-04 at 14 00 29


willfish (approved):
So I know this has nothing to do with what you're doing but I'm wondering if its possible to pull these from an API somewhere?

PR #961, opened 2023-01-05
Fix tariff sync rake task

jebw:
### Jira link

HOTT-????

### What?

I have added/removed/altered:

- [x] Fixed `rake tariff:sync`
- [x] Amended `rake tariff:download_and_apply` to `rake tariff:download_apply_and_reindex`
- [x] Lint fixes
- [x] Resolve warning from Rubocop

### Why?

I am doing this because:

- `rake tariff:sync` was broken
- download and apply doesn't convey that unlike other tariff tasks this is going to perform a very long running action
- linter was complaining
- linter was complaining

- Low

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -64,7 +64,6 @@ Rails/UniqueValidationWithoutIndex:
 
 Style/Documentation:
   Enabled: false
-  EnforcedStyle: never
 
 Style/FrozenStringLiteralComment:
   Enabled: true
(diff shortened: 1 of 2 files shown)


willfish:
I think the warning is a false negative :thinking: . It actually has the effect I wanted at the time/I needed to dig into the implementation to work it out :eyes:

willfish (approved):

PR #963, opened 2023-01-06
HOTT-2228: Retry job or notify slack if ListDownloadFailedError is raised

maverickvi:
### Jira link

[HOTT-<2228>](https://transformuk.atlassian.net/browse/HOTT-2228)
https://transformuk.atlassian.net/browse/HOTT-2273

### What?

I have added/removed/altered:

- [ ] Raise error if status code is not 200
- [ ] Retry job or notify slack if ListDownloadFailedError is raised

### Why?

I am doing this because:

- We only want the code to execute if HTTP status code is 200
- We want to be notified on Slack or re-attempt failed job


jebw (commented):
Looks really good, I'm not sure we need to be checking the parameters of the requeue though - its enough just to know its been added back onto the queue?

willfish (approved):

PR #964, opened 2023-01-09
HOTT-2444: Use default date and time format for last_allocation_date

maverickvi:
### Jira link

[HOTT-<TODO>](https://transformuk.atlassian.net/browse/HOTT-2444)

### What?

I have added/removed/altered:

- [ ] Used default date and time format for last_allocation_date

### Why?

I am doing this because:

- We want a consistent Date/time format on our API so it doesn't cause issues for API consumers

Screenshot 2023-01-09 at 16 23 01


willfish:
Are there other places that use this field in the api and do we want to update them to be a non-date date?

willfish:
I'm thinking the quota search results for consistency for our frontend

willfish (approved):

PR #972, opened 2023-01-12
HOTT-1596: Adds implicit deletion to goods nomenclature primary

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1596

### What?

I have added/removed/altered:

- [x] Added implicit delete behaviour to goods nomenclature mapper
- [x] Added test coverage for the new behaviour

### Why?

I am doing this because:

- This is required to make sure we're clearing out indents that are no longer supplied at the end of the indent sequence (e.g. ironing boards being fixed by removal)


jebw (approved):
LGTM

PR #974, opened 2023-01-13
HOTT-2372: Do not load s3 bucket resource if no env

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2372

### What?

I have added/removed/altered:

- [x] Fixed bug in initialiser

### Why?

I am doing this because:

- This will let users load the s3 bucket without a configured environment


jebw (approved):
LGTM :raised_hands:

PR #1005, opened 2023-01-26
HOTT-1613: Quota order number missing deletes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1613

### What?

I have added/removed/altered:

- [x] Clear out excess unused content from CDS parser
- [x] Enable missing deletes of quota order numbers

### Why?

I am doing this because:

- The excess content has no impact - it just looks weird in a console
- This will mean that when CDS/DIT/TAP produce a file which is missing some order number secondaries in it we will clear them out


jebw (approved):
LGTM

PR #1006, opened 2023-01-27
HOTT-1612: Update primary keys of quota definition secondaries

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2590

### What?

I have added/removed/altered:

- [x] Altered the primary keys of various quota events
- [x] Fixed the incorrect primary key of the blocking period model

### Why?

I am doing this because:

- The current status of a quota definition can change multiple times (e.g. become exhausted, then critical and back again). This wouldn't be reflected if the exhausted could only exist once in the history of the quota definition. We may also want to plot these events on the graph in the admin UI at a later date.
- The blocking period model was not using its correct primary key


jebw (approved):
LGTM

PR #1008, opened 2023-01-30
HOTT-2593: Remove brittle check of transferred amount

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2593

### What?

I have added/removed/altered:

- Adds failing spec to capture transferred_amount issue
- Removes check on transferred amount to find equivalent node (making the test pass)

### Why?

I am doing this because:

- This is required to make sure we can import transfer events that have a different string representation for their transferred amounts than what we would pull out of the model

```ruby
a = event.transferred_amount.to_s => "1444634.0"
b = xml_node_event['transferredAmount'] => "1444634"

a == b => false
```


jebw (approved):
LGTM

PR #1010, opened 2023-01-30
HOTT-2352: Exposed additional codes on headings API

maverickvi:
### Jira link

[HOTT-<TODO>
](https://transformuk.atlassian.net/browse/HOTT-2352)

### What?

I have added/removed/altered:

- [ ] Exposed additional codes on headings API

### Why?

I am doing this because:

- We need to access these attributes on the frontend

Screenshot 2023-01-30 at 14 36 21


willfish:
Is it worth just fixing the 5 lint violations?

app/serializers/cache/heading_serializer.rb
```diff
@@ -143,7 +158,7 @@ def as_json
               measure_type_id: measure.measure_type.measure_type_id,
               description: measure.measure_type.description,
             },
-          }
+          }.merge(additional_code)
```
willfish, line 161: It'd be cleaner to just add this to the body of the Hash as far as I can make out

app/serializers/cache/heading_serializer.rb
```diff
@@ -119,12 +119,27 @@ def as_json
                                 [
                                   { measure_type: [:measure_type_description] },
                                   { measure_components: [{ duty_expression_description: :duty_expression_description }, :measurement_unit_qualifier] },
+                                  { additional_code: [:additional_code_description] },
```
willfish, line 122: This is such a slow query :(.

app/serializers/cache/heading_serializer.rb
jebw: @willfish This was actually my suggestion - its not possible to add it into the hash because the the contents of that `additional_code` hash are conditional? Sometimes they are an empty hash, sometimes there are keys -> if it was one level down you could just include it in the hash but unfortunately thats not the case here?

spec/serializers/cache/heading_serializer_spec.rb
```diff
@@ -23,13 +23,16 @@
       goods_nomenclature_sid: heading.goods_nomenclature_sid,
     )
   end
+  let!(:additional_code) { create :additional_code, validity_start_date: 3.years.ago.beginning_of_day }
```
willfish, line 26: AFAIK these are both violations of our linting rules?

I'd just add a trait for these (:with_additional_code)

app/serializers/api/v2/measures/overview_measure_serializer.rb
```diff
@@ -13,6 +13,7 @@ class OverviewMeasureSerializer
 
         has_one :duty_expression, serializer: Api::V2::Measures::DutyExpressionSerializer
         has_one :measure_type, serializer: Api::V2::Measures::MeasureTypeSerializer
+        has_one :additional_code, if: proc { |measure| measure.additional_code.present? }, serializer: Api::V2::AdditionalCodeSerializer
```
willfish, line 16: I don't think we need this if proc. Isn't this the default behaviour?

app/serializers/cache/heading_serializer.rb
```diff
@@ -119,12 +119,27 @@ def as_json
...
                                 ],
                               )
                             else
                               commodity.overview_measures
                             end
         commodity_attributes[:overview_measures] = overview_measures.map do |measure|
+          additional_code = if measure.additional_code
+                              {
+                                additional_code_id: measure.additional_code_sid.to_s,
```
willfish, line 131: JSONAPI ids automatically get to_s called on them. You can remove the to_s here and on line 133

app/serializers/api/v2/measures/overview_measure_serializer.rb
willfish: We should add a test for this serializer, too

willfish: So I think what's happening here is you've been trying to serialize a hashie mash which will ask for the id value to check the presence of the relation. Since you're conditionally serialize two different structures you're getting errors when using this serializer in the case where tehre isn't an additional code

app/serializers/api/v2/headings/commodity_serializer.rb
```diff
@@ -18,15 +18,10 @@ class CommoditySerializer
                    :description_plain,
                    :parent_sid,
                    :validity_start_date,
-                   :validity_end_date
+                   :validity_end_date,
+                   :declarable
```
willfish, line 22: Added this to reflect the fact that we're now using fast declarable and this never got updated

willfish (approved):

PR #1011, opened 2023-01-30
HOTT-2595: Fixes smelly design and reverse engineering attributes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2595

### What?

I have added/removed/altered:

- [x] Adds support for passing expanded attributes to callback functions
- [x] Removes unused code

### Why?

I am doing this because:

- This is required to simplify any reverse engineering of the attributes that are applicable for the current invocation of a before insert callback. Where before we tried to determine surrounding attributes of a given model instance, now we can just ask for them from the passed in expanded attributes

### DIFF
--- a/app/lib/cds_importer/entity_mapper/base_mapper.rb
+++ b/app/lib/cds_importer/entity_mapper/base_mapper.rb
@@ -182,22 +182,6 @@ class CdsImporter
         expanded.map(&method(:build_instance))
       end
 
-      def destroy_operation?
-        @xml_node.dig('metainfo', 'opType') == Sequel::Plugins::Oplog::DESTROY_OPERATION &&
-          primary?
-      end
-
-      # In the CDS file we treat the parent node differently from the
-      # secondary child nodes when it comes to support for the destroy operation.
-      #
-      # Each entity mapper represents either a child or a parent node in the xml file.
-      #
-      # Parent nodes have the same assigned entity class as their derived class. Our internal naming
-      # for this has been to name this parent node the primary.
-      def primary?
-        derived_entity_class == entity_class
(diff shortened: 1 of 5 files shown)


app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -164,7 +164,7 @@ def parse
               # iterating through all items in Array and creating @values copy
               value.each do |v|
                 # [1,2,3] => {1=>{2=>{3=>value}}
-                tmp = current_path.lazy.reverse_each.inject(v) do |memo, i|
+                tmp = current_path.reverse.inject(v) do |memo, i|
```
willfish, line 167: Lazy enumeration here makes 0 sense so to simplify I've got rid and not noticed any difference behaviourally

app/lib/cds_importer/entity_mapper/base_mapper.rb
```diff
@@ -54,7 +54,7 @@ def base_mapping
 
         # Register a callback to soft delete missing entities indicated by the passed in secondary mappers
         def delete_missing_entities(*secondary_mappers)
-          before_oplog_inserts do |xml_node, mapper_instance, primary_model_instance|
+          before_oplog_inserts do |xml_node, mapper_instance, primary_model_instance, _expanded_attributes|
```
willfish, line 57:
```suggestion
          before_oplog_inserts do |xml_node, mapper_instance, primary_model_instance|
```

The arguments we pass are dropped if the block arity is less than all of the arguments. No need to add this here

maverickvi (approved):

PR #1018, opened 2023-01-31
HOTT-2605: Migrate to opensearch client

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2605

### What?

You can see that the maintainers are the same for both gems - the api is also exactly the same:

2023-01-31-150155_3392x1370_scrot

I have added/removed/altered:

- [x] Upgrade to 2.x opensearch in docker-compose
- [x] Migrate from elasticsearch to opensearch client

### Why?

I am doing this because:

- We've been pinned to an older version of the elasticsearch gem (maintained by the same people and with the same api) for a long time now. I generally don't like not being able to upgrade gems and this solves that.
- In my local docker-compose I've been using 2.x for a while now - this reflects what we're using in aws, too

### AC

- [x] Can reindex beta search
- [x] Can view a heading retrieved from the opensearch cache on dev
- [x] Can perform legacy search on dev
- [x] Can perform beta search on dev

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,8 +26,7 @@ gem 'sidekiq'
 gem 'sidekiq-scheduler'
 
 # Elasticsearch
-gem 'elasticsearch', '~> 7.9.0' # Bumping this causes failures
-gem 'elasticsearch-extensions'
+gem 'opensearch-ruby'
 
 # Helpers
 gem 'bootsnap', require: false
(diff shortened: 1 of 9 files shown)


willfish:
Populating beta search indexes:

2023-01-31-151156_1311x1158_scrot

willfish:
Using beta search:

image

Using legacy search:

image
image

Fetching a heading out of the opensearch cache index:

image

maverickvi (approved):

PR #1030, opened 2023-02-03
HOTT-2623: Alter footnote type id field to support 3 chars

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2623

### What?

I have added/removed/altered:

- [x] Added support for a 3 character varying string on footnote type id

### Why?

I am doing this because:

- You can't alter the underlying table of a view so we needed to drop and recreate these


jebw (approved):
LGTM

PR #1033, opened 2023-02-06
HOTT-2631: Fix bug in Taric importer

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2631

### What?

I have added/removed/altered:

- [x] Adds characters callback specs in the XmlParser::Reader implementations of both CDS and Taric
- [x] Fixes issue with the characters callback which was not checking the start of the line

### Why?

I am doing this because:

- This is causing footnote descriptions to not be imported in the XI service.
- We would have been alerted to this not working on the UK service although I'm sure we will encounter it eventually


jebw (approved):
LGTM

PR #1041, opened 2023-02-07
HOTT-2381: Dereference referenced_id column

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2381

### What?

I have added/removed/altered:

- [x] Drop references to search reference referenced id
- [x] Consolidate reference_id/admin_id
- [x] Add slack notification for when we're clearing out old search references

### Why?

I am doing this because:

- This is just part of a general tidying up of search references and enables us to drop the referenced_id column


app/models/search_reference.rb
```diff
@@ -20,12 +19,10 @@ class SearchReference < Sequel::Model
     end
   end
 
-  referenced_dataset = proc do
+  many_to_one :referenced, key: :goods_nomenclature_sid, reciprocal: :search_references, reciprocal_type: :many_to_one, setter: referenced_setter, dataset: proc {
```
willfish, line 22: referenced_dataset is one of the methods the association dsl ends up attaching to the class singleton the DSL is called on so I'm removing it so that we don't confuse these.

willfish: I'm adding the key here because when you call `.referenced` you call `.referenced_id` under the hood which creates a stack overflow

app/models/search_reference.rb
```diff
@@ -43,19 +40,12 @@ def indexable
...
-             '/subheadings/:id'
-           else
-             '/commodities/:id'
-           end
+  def referenced_id
+    referenced.to_admin_param
+  end
 
-    path.sub(':id', referenced.to_param)
+  def resource_path
```
willfish, line 47: Passing cleanup

app/workers/clear_invalid_search_references.rb
```diff
@@ -5,10 +5,15 @@ class ClearInvalidSearchReferences
...
-      end
+      next if ref.referenced.current?
+
+      message = "Removed Search reference: id:#{ref.id}, title:'#{ref.title}'"
+
+      ref.delete
+
+      logger.info(message)
+
+      SlackNotifierService.call(message)
```
willfish, line 16: It's actually interesting for us to know when these have expired

jebw (approved):
LGTM

maverickvi (approved):

PR #1042, opened 2023-02-07
HOTT-2475 Added a maintenance mode

jebw:
### Jira link

HOTT-2475

### What?

I have added/removed/altered:

- [ ] Added Maintenance mode

### Why?

I am doing this because:

- if we do need it then it will be very useful

- Adds a before action on all api requests


app/controllers/healthcheck_controller.rb
```diff
@@ -1,4 +1,6 @@
 class HealthcheckController < ApplicationController
+  skip_before_action :maintenance_mode_if_active
```
willfish, line 2: :+1:

PR #1043, opened 2023-02-07
HOTT-2381: Drop referenced_id column

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2381

### What?

I have added/removed/altered:

- [x] Drop referenced_id column

### Why?

I am doing this because:

- This is part of us migrating to a more sensible relationship layout for search references - next step involves renaming the goods_nomenclature_sid column to be referenced_id to conform to polymorphic standards of associations

### DIFF
--- a/db/migrate/20230207114821_drop_referenced_id_on_search_references.rb
+++ b/db/migrate/20230207114821_drop_referenced_id_on_search_references.rb
@@ -1,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table :search_references do
+      drop_column :referenced_id
+    end
+  end
+
+  down do
+    alter_table :search_references do
+      add_column :referenced_class, String, size: 10
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


willfish:
Plan is to merge and release this after Thursday's release (which will include dereferencing the referenced_id)

db/migrate/20230207114821_drop_referenced_id_on_search_references.rb
```diff
@@ -0,0 +1,13 @@
+Sequel.migration do
+  up do
+    alter_table :search_references do
+      drop_column :referenced_id
+    end
+  end
+
+  down do
+    alter_table :search_references do
+      add_column :referenced_class, String, size: 10
```
jebw, line 10: This should be `referenced_id` rather then `referenced_class`

jebw (changes requested):
Looks good great, only query is the `down` migration will fail because it'll try and add an already existing column

db/migrate/20230207114821_drop_referenced_id_on_search_references.rb
willfish: :facepalm:

maverickvi (approved):

jebw (approved):
:+1:

PR #1067, opened 2023-02-17
HOTT-2393: Adds direct hit to search result

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2393

### What?

I have added/removed/altered:

- [x] Adds a short code classifier
- [x] Adds a direct hit model
- [x] Surface direct hit in the api
- [x] Adds test coverage

### Why?

I am doing this because:

- This is required to simplify single hit/search reference single hit and numeric single hit results into a single polymorphic form that the frontend can just redirect the user too

### DIFF
--- a/app/models/beta/search/direct_hit.rb
+++ b/app/models/beta/search/direct_hit.rb
@@ -5,13 +5,24 @@ module Beta
                     :goods_nomenclature_item_id,
                     :producline_suffix
 
-      attr_reader :description,
-                  :formatted_description,
-                  :validity_start_date,
-                  :validity_end_date
-
       def id
         "#{goods_nomenclature_item_id}-#{producline_suffix}"
+      end
+
+      def method_missing(method_name, *_arguments)
+        if method_name.to_s.start_with?('search_references')
+          []
+        elsif method_name.to_s.start_with?('guide')
+          []
+        elsif method_name.to_s.start_with?('ancestors')
+          []
+        end
+      end
+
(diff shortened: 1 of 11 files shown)


app/models/beta/search/direct_hit.rb
```diff
@@ -0,0 +1,56 @@
...
+    class DirectHit
+      attr_accessor :goods_nomenclature_class,
+                    :goods_nomenclature_item_id,
+                    :producline_suffix
+
+      def id
+        "#{goods_nomenclature_item_id}-#{producline_suffix}"
+      end
+
+      def method_missing(method_name, *_arguments)
```
willfish, line 12: There's a bunch of uninteresting (for redirecting) goods nomenclature stuff that I'm skipping over adding specific readers/default values for by using this callback

app/models/beta/search/direct_hit.rb
```diff
@@ -0,0 +1,56 @@
+module Beta
+  module Search
+    class DirectHit
+      attr_accessor :goods_nomenclature_class,
+                    :goods_nomenclature_item_id,
+                    :producline_suffix
+
+      def id
```
willfish, line 8: I needed an id and rather than doing a request to the db to fetch the sid it made sense just to make this what would otherwise be the unique (for today) identifier for the place in the goods nomenclature hierarchy

maverickvi (approved):

PR #1077, opened 2023-02-22
HOTT-2688 Refactored GoodsNomenclature factories

jebw:
### Jira link

HOTT-2688

### What?

I have added/removed/altered:

- [x] Removes use of rand in item ids
- [x] Removes hard coding of `.path` attribute
- [x] Removes hard coding of goods_nomenclature_sid
- [x] Creates indents by default since NestedSet will rely upon them

### Why?

I am doing this because:

The Nested Set relies heavily on the ability to create multiple hierarchies reliably 
- as part of that work I was running into both outright collisions on sids
- factories obscuring data they'd just created due to how the oplog handles duplicate sids
- fixing up some of this hardcoding in term broke a lot of existing specs relating to materialized path because the factories were hardcoding a hierarchy which was no longer correct. Hence I've refactored it to use a transient parent attribute and use that for dynamic generation of materialized paths `path` column
- NestedSet will rely on correct generation of `item_id` to support its hierarchy logic - ie creating 2 commodities should result in 2 siblings, not the second potentially occurring before the first once sorted

- Low - some minimal changes to specs but almost exclusively changes to just factories.
- No 'application code' is changed in this PR

### DIFF
--- a/spec/controllers/api/v1/headings_controller_spec.rb
+++ b/spec/controllers/api/v1/headings_controller_spec.rb
@@ -72,7 +72,7 @@ RSpec.describe Api::V1::HeadingsController, flaky: true do
 
       context 'when record is not present' do
         it 'returns not found if record was not found' do
-          id = heading.goods_nomenclature_item_id.first(4).to_i + 1
+          id = heading.goods_nomenclature_item_id.first(4).next
           get :show, params: { id: }, format: :json
 
           expect(response.status).to eq 404
@@ -117,7 +117,7 @@ RSpec.describe Api::V1::HeadingsController, flaky: true do
 
       context 'when record is not present' do
         it 'returns not found if record was not found' do
-          id = heading.goods_nomenclature_item_id.first(4).to_i + 1
+          id = heading.goods_nomenclature_item_id.first(4).next
           get :show, params: { id: }, format: :json
 
           expect(response.status).to eq 404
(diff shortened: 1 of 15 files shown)


spec/controllers/api/v1/headings_controller_spec.rb
```diff
@@ -72,7 +72,7 @@
 
       context 'when record is not present' do
         it 'returns not found if record was not found' do
-          id = heading.goods_nomenclature_item_id.first(4).to_i + 1
+          id = heading.goods_nomenclature_item_id.first(4).next
```
jebw, line 75: This increments 0109 to 0110 instead of 110 which is an invalid heading code, causing this and similar specs to fail when using dynamic item ids - this pattern used elsewhere in this PR

spec/factories/chapter_factory.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :chapter, parent: :goods_nomenclature, class: 'Chapter' do
-    goods_nomenclature_item_id { "#{2.times.map { Random.rand(9) }.join}00000000" }
+    goods_nomenclature_item_id { "#{generate(:chapter_item_id)}00000000" }
```
willfish, line 3:
```suggestion
    goods_nomenclature_item_id { "#{generate(:chapter_short_code)}00000000" }
```

PR #1092, opened 2023-02-28
HOTT-2640 Add tree node relationships

jebw:
### Jira link

HOTT-2640

### What?

I have added/removed/altered:

- [x] Added a more generic mechanism for adding TimeMachine constraints to queries
- [x] Added a query generated for previous sibling
- [x] Added a query generated for next sibling
- [x] Added nested set parent relationship to Goods Nomenclature
- [x] Added nested set ancestors relationship to Goods Nomenclature
- [x] Added nested set children relationship to Goods Nomenclature
- [x] Added nested set descendants relationship to Goods Nomenclature

### Why?

I am doing this because:

- I want fast time machine compatible mechanisms to query the hierarchy

- Purely addition of unused relationships so should be safe

### DIFF
--- a/app/lib/sequel/plugins/time_machine.rb
+++ b/app/lib/sequel/plugins/time_machine.rb
@@ -43,6 +43,24 @@ module Sequel
         def relevant_query?
           Thread.current[::TimeMachine::THREAD_RELEVANT_KEY]
         end
+
+        def validity_dates_filter(table = self,
+                                  start_column: :validity_start_date,
+                                  end_column: :validity_end_date)
+          return self if point_in_time.blank?
+
+          table_name = if table.is_a?(Class) && table < Sequel::Model
+                         table.table_name
+                       else
+                         table
+                       end
+
+          qualified_start_column = Sequel.qualify(table_name, start_column)
+          qualified_end_column   = Sequel.qualify(table_name, end_column)
+
+          (qualified_start_column <= point_in_time) &
+            ((qualified_end_column >= point_in_time) | (qualified_end_column =~ nil))
+        end
       end
 
       module InstanceMethods
@@ -97,6 +115,14 @@ module Sequel
             filter { |o| o.<=(klass.period_start_date_column, klass.point_in_time) & (o.>=(klass.period_end_date_column, klass.point_in_time) | ({ klass.period_end_date_column => nil })) }
           else
             self
+          end
+        end
+
+        def with_validity_dates(table = model.table_name,
+                                start_column: :validity_start_date,
+                                end_column: :validity_end_date)
+          where do |_query|
+            model.validity_dates_filter(table, start_column:, end_column:)
           end
         end
       end
--- a/app/lib/sequel/plugins/time_machine.rb
+++ b/app/lib/sequel/plugins/time_machine.rb
@@ -47,7 +47,7 @@ module Sequel
         def validity_dates_filter(table = self,
                                   start_column: :validity_start_date,
                                   end_column: :validity_end_date)
-          return self if point_in_time.blank?
+          return Sequel.expr(true) if point_in_time.blank?
 
           table_name = if table.is_a?(Class) && table < Sequel::Model
                          table.table_name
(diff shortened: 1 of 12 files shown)


app/models/goods_nomenclatures/nested_set.rb
```diff
@@ -0,0 +1,129 @@
...
+                   right_primary_key: :goods_nomenclature_sid,
+                   right_key: :goods_nomenclature_sid,
+                   class_name: '::GoodsNomenclature',
+                   join_table: Sequel.as(:goods_nomenclature_tree_nodes, :ancestor_nodes),
+                   read_only: true do |ds|
+        ds.order(:ancestor_nodes__depth)
+          .with_validity_dates(:ancestor_nodes)
+          .select_append(:ancestor_nodes__depth)
+          .join(Sequel.as(:goods_nomenclature_tree_nodes, :origin_nodes)) do |origin_table, ancestors_table, _join_clauses|
+            ancestors_position = Sequel.qualify(ancestors_table, :position)
```
willfish, line 27: There's quite a bit of duplication between the association methods with these variables? Is it worth extracting them out?

app/models/goods_nomenclatures/tree_node.rb
```diff
@@ -8,6 +17,39 @@ class << self
...
+          .where do |_query|
+            (siblings_depth =~ origin_depth) &
+              (siblings_position < origin_position) &
+              (siblings_position >= TreeNode.start_of_chapter(origin_position)) &
+              validity_dates_filter(:siblings)
+          end
+      end
+
+      def start_of_chapter(position_column)
+        (position_column / 10_000_000_000) * 10_000_000_000
```
willfish, line 37: Would be good to give this number a name

willfish (approved):
Some of the associations are quite complex. I'm not sure if they can be tidied with subqueries/shared sequel expressions, etc

app/models/goods_nomenclatures/nested_set.rb
jebw: Fair point - I'll have a think, the complication is there's no object to attach the state to so you just end up pass a lot of params around - should be able to do something though

app/models/goods_nomenclatures/tree_node.rb
jebw: Yeh, good should mean to circle back to this and forgot

app/lib/sequel/plugins/time_machine.rb
```diff
@@ -43,6 +43,24 @@ def point_in_time
         def relevant_query?
           Thread.current[::TimeMachine::THREAD_RELEVANT_KEY]
         end
+
+        def validity_dates_filter(table = self,
+                                  start_column: :validity_start_date,
+                                  end_column: :validity_end_date)
+          return Sequel.expr(true) if point_in_time.blank?
+
+          table_name = if table.is_a?(Class) && table < Sequel::Model
```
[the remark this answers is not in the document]
jebw: Yeh, I had to look this up - its a bit quirky because I'm testing the model class rather then an a model instance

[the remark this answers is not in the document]
willfish: Thinking about this, to implement the plugin you'd need to inherit from the Sequel::Model ancestor. Is there any point in checking this?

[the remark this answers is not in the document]
willfish: I mean it doesn't hurt

willfish (approved):

app/lib/sequel/plugins/time_machine.rb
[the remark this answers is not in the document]
jebw: Yeh - but table is a value passed in - its intended to handle different scenarios - if its given a model class, it'll work with that and fetch out `table_name` but if its given a symbol for the table then it needs to treat it as a table name and not be clever about it

I ended up adding that stuff because the existing methods aren't particularly flexible and I needed something I could use in a few different scenarios in the complex joins

app/models/goods_nomenclatures/nested_set.rb
```diff
@@ -1,9 +1,9 @@
+# See docs/goods-nomenclature-nested-set.md for an explanation of this
```
willfish, line 1: Good idea

PR #1093, opened 2023-02-28
HOTT-2690 custom loader for descendants

jebw:
### Jira link

HOTT-2690

### What?

I have added/removed/altered:

- [x] Added an RSpec matcher to compare models just by primary keys rather then all values
- [x] Added a custom loader for descendants to recursively populate children relationship, together with their descendants, children and parents
- [x] Added a custom loader for ancestors to recursively populate, parents and ancestors of all ancestors
- [x] Added a nested set version of `declarable?`

### Why?

I am doing this because:

- Objects returned via relationships have an additional `depth` value - and just fetching out SIDs to compare objects can be DRYed out
- This avoids unneeded relationship queries for objects we can already derive from the descendants we have already loaded
- This avoids unneeded relationship queries for objects we can already derive from the ancestors we have already loaded
- This version can operate without querying the database if children are already loaded (eg via an eager load of descendants)

- Low, shouldn't change any in use code

### DIFF
--- a/app/models/goods_nomenclatures/nested_set.rb
+++ b/app/models/goods_nomenclatures/nested_set.rb
@@ -168,5 +168,9 @@ module GoodsNomenclatures
     def depth
       values.key?(:depth) ? values[:depth] : tree_node.depth
     end
+
+    def ns_declarable?
+      producline_suffix == GoodsNomenclatureIndent::NON_GROUPING_PRODUCTLINE_SUFFIX && ns_children.empty?
+    end
   end
 end
--- a/app/models/goods_nomenclatures/nested_set.rb
+++ b/app/models/goods_nomenclatures/nested_set.rb
@@ -19,6 +19,7 @@ module GoodsNomenclatures
                    right_key: :goods_nomenclature_sid,
                    class_name: '::GoodsNomenclature',
                    join_table: Sequel.as(:goods_nomenclature_tree_nodes, :ancestor_nodes),
+                   after_load: :recursive_ancestor_populator,
                    read_only: true do |ds|
         ds.order(:ancestor_nodes__depth)
           .with_validity_dates(:ancestor_nodes)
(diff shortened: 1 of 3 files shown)


app/models/goods_nomenclature.rb
```diff
@@ -24,6 +24,8 @@ class GoodsNomenclature < Sequel::Model
     end
   }
 
+  include GoodsNomenclatures::NestedSet
```
willfish, line 27: I think I've seen this in a previous PR. Do we need to rebase?

app/models/goods_nomenclatures/nested_set.rb
```diff
@@ -0,0 +1,176 @@
...
+      parents_ancestors = ancestors.dup
+      parent = parents_ancestors.pop
+      @associations[:ns_parent] ||= parent
+      return if ancestors.empty?
+
+      parent.recursive_ancestor_populator(parents_ancestors)
+    end
+
+    def recursive_descendant_populator(descendants, parent = nil)
+      @associations ||= { ns_descendants: descendants }
```
willfish, line 139: Is this populating the same association cache that sequel uses internally?

app/models/goods_nomenclatures/nested_set.rb
```diff
@@ -0,0 +1,176 @@
...
+                  :coalesce,
+                  TreeNode.next_sibling(origin_position, origin_depth),
+                  END_OF_TREE,
+                )
+              )
+          end
+      end
+    end
+
+    def recursive_ancestor_populator(ancestors)
```
willfish, line 127: I'm wondering if there's anything from the rcte_tree plugin we could reuse here? Feels like a lot of overlap

app/models/goods_nomenclatures/nested_set.rb
jebw: Yes

app/models/goods_nomenclatures/nested_set.rb
jebw: Its using the `after_load` approach from it - that rcte plugin is quite a homogeneous blob, there's not really individual parts we can pull in to use to handle the loading 

I'd used that as the inspiration for this though

willfish (approved):
Awesome and clean

PR #1099, opened 2023-03-02
HOTT-2641 Override existing tree methods

jebw:
### Jira link

HOTT-2641

### What?

I have added/removed/altered:

- [x] Added a feature flag to enable use of NestedSet for existing code paths
- [x] Added overrides for different goods nomenclature types to use nested set for existing hierarchy methods

### Why?

I am doing this because:

- So we can turn this feature on and off
- It provides a starting point for integrating the nested set work

- With the flag off, low, 
- With the flag on, high since it changes our queries used for the hierarchy

Note: the time machine fix is cherry-picked back into the relationships branch where it was introduced and will go away when the intermediate branches get rebased.

### DIFF
--- a/app/lib/trade_tariff_backend.rb
+++ b/app/lib/trade_tariff_backend.rb
@@ -290,5 +290,9 @@ module TradeTariffBackend
     def opensearch_debug
       ENV.fetch('OPENSEARCH_DEBUG', 'false') == 'true'
     end
+
+    def use_nested_set?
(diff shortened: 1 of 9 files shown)


willfish (approved):
Excited :scream_cat:

jebw:
Concluded this was too likely to introduce N+1s and since most code areas have been manually converted to nested set

PR #1117, opened 2023-03-13
HOTT-2832: Handle failing healthchecks with a proper body and status

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2823

### What?

I have added/removed/altered:

- [x] Added 500 status response with proper body when the healthcheck fails
- [x] Updated tests to reflect changes

### Why?

I am doing this because:

- This is to work seamlessly with the current healthcheck implementation and actually makes checking a bit easier


app/models/healthcheck.rb
```diff
@@ -32,6 +33,16 @@ def read_revision_file
     nil
   end
 
+  def all_healthy?
+    [
```
willfish, line 37: Make short-circuiting or

app/models/healthcheck.rb
```diff
@@ -32,6 +33,16 @@ def read_revision_file
     nil
   end
 
+  def all_healthy?
```
jebw, line 36: This will result in all the health checks getting run twice?

how about something like

```
checks = {
  sidekiq: sidekiq_healthy?
}

checks.merge(healthy: check.values.all?)
```

jebw (approved):
Looks good but does end up running all the checks twice, including the HTTP ones?

Should be enough to just check the values in the `check` hash are all true?

app/models/healthcheck.rb
jebw: Also is failing the specs? Looks like it might not be returning service unavailable?

willfish: I had the same though, yeah. Will fix

PR #1124, opened 2023-03-14
HOTT-2829 Use depth for number_indents if available

jebw:
### Jira link

HOTT-2829

### What?

I have added/removed/altered:

- [x] Avoid fetch indents records if we already know `depth` 

### Why?

I am doing this because:

- It will potentially avoid N+1s
- If we've got `depth`, we can derive indent level from that without fetching data

- Low, behind feature flag


willfish (approved):

willfish:
Think this might need a rebase/has become a bit more than the tweak to indent fetching?

jebw:
There's better approaches to this so closing and will create alternative PR

PR #1126, opened 2023-03-14
HOTT-2840: Clear expired goods nomenclature from suggestions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2840

### What?

I have added/removed/altered:

- [x] Move to generating only current goods nomenclature as suggestions
- [x] Add test coverage for new behaviour

### Why?

I am doing this because:

- We have a bunch of expired search suggestions that are cluttering up the typeahead

### DIFF
--- a/app/services/api/v2/suggestions_service.rb
+++ b/app/services/api/v2/suggestions_service.rb
@@ -18,10 +18,6 @@ module Api
       def handle_search_reference_record(search_reference)
         Api::V2::SuggestionPresenter.new(search_reference.id, search_reference.title)
       end
-
-      def handle_chemical_record(chemical)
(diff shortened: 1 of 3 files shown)


app/services/api/v2/suggestions_service.rb
```diff
@@ -18,10 +18,6 @@ def handle_commodity_record(commodity)
       def handle_search_reference_record(search_reference)
         Api::V2::SuggestionPresenter.new(search_reference.id, search_reference.title)
       end
-
-      def handle_chemical_record(chemical)
```
willfish, line 22: We've stopped populating the typeahead with chemicals - this was just leftover cruft

maverickvi (approved):

PR #1127, opened 2023-03-14
HOTT-2844 validity periods for schemes

jebw:
### Jira link

HOTT-2844

### What?

I have added/removed/altered:

- [x] Added support for validity dates in the Rules of Origin schemes file

### Why?

I am doing this because:

- We have some schemes that are only applicable up to or from certain dates

- Low

### DIFF
--- a/app/models/rules_of_origin/scheme.rb
+++ b/app/models/rules_of_origin/scheme.rb
@@ -6,7 +6,7 @@ module RulesOfOrigin
                   :fta_intro_file, :countries, :footnote, :adopted_by_uk, :country_code, :notes,
                   :unilateral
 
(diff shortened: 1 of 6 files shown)


app/models/rules_of_origin/scheme.rb
```diff
@@ -105,5 +117,18 @@ def read_rule_sets
     rescue Errno::ENOENT
       { 'rule_sets' => [] }
     end
+
+    def parse_date(value, date_cast)
```
willfish, line 121: Do we need all of this for what is probably Matt just picking a format?

willfish (approved):
LGTM

PR #1132, opened 2023-03-16
HOTT-2879: Cache chapter responses

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2879

### What?

I have added/removed/altered:

- [x] Adds caching to v2 chapter responses
- [x] Adds specs for the caching

### Why?

I am doing this because:

- This is a temporary fix just to diminish the impacts of users scripting concurrent requests to chapter endpoints

### DIFF
--- a/app/controllers/api/v2/chapters_controller.rb
+++ b/app/controllers/api/v2/chapters_controller.rb
@@ -3,57 +3,71 @@ require 'goods_nomenclature_mapper'
 module Api
   module V2
     class ChaptersController < ApiController
-      before_action :find_chapter, only: %i[show changes headings]
+      CACHE_VERSION = 1
 
       def index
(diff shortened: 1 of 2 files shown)


app/controllers/api/v2/chapters_controller.rb
```diff
@@ -3,57 +3,71 @@
 module Api
   module V2
     class ChaptersController < ApiController
-      before_action :find_chapter, only: %i[show changes headings]
```
willfish, line 6: Removing this out of a callback because we might not need to fetch this from the database if we have something cached for the chapter id and actual date on the current service

willfish:
See https://github.com/trade-tariff/trade-tariff-testing/pull/1110 for smoke test failures

jebw (approved):
LGTM

PR #1133, opened 2023-03-16
HOTT-2879: Caches v1 chapter responses

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2879

### What?

I have added/removed/altered:

- [x] Added caching to the v1 chapter responses
- [x] Added coverage for this new caching behaviour

### Why?

I am doing this because:

- This is required to make sure these endpoints don't keep flooding the database with queries

### DIFF
--- a/app/controllers/api/v1/chapters_controller.rb
+++ b/app/controllers/api/v1/chapters_controller.rb
@@ -3,46 +3,77 @@ require 'goods_nomenclature_mapper'
 module Api
   module V1
     class ChaptersController < ApiController
-      before_action :find_chapter, only: [:show, :changes]
-
       def index
-        @chapters = Chapter.eager(:chapter_note).all
+        cache_key = "_v1_chapters-#{actual_date}-view"
 
-        respond_with @chapters
+        serialized_result = Rails.cache.fetch(cache_key, expires_at: actual_date.end_of_day) do
+          render_to_string(
+            template: 'api/v1/chapters/index',
+            formats: [:json],
+            locals: { chapters: },
+          )
+        end
+
+        render json: serialized_result
       end
 
       def show
-        @headings = GoodsNomenclatureMapper.new(@chapter.headings_dataset
-                                                        .eager(:goods_nomenclature_descriptions,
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/chapters_controller.rb
```diff
@@ -3,46 +3,77 @@
...
   module V1
     class ChaptersController < ApiController
-      before_action :find_chapter, only: [:show, :changes]
-
       def index
-        @chapters = Chapter.eager(:chapter_note).all
+        cache_key = "_v1_chapters-#{actual_date}-view"
+
+        serialized_result = Rails.cache.fetch(cache_key, expires_at: actual_date.end_of_day) do
+          render_to_string(
```
willfish, line 10: This block basically means we can bypass any database queries when something exists for this rendered view

spec/controllers/api/v1/chapters_controller_spec.rb
```diff
@@ -1,193 +1,229 @@
-RSpec.describe Api::V1::ChaptersController, 'GET #show' do
-  render_views
+RSpec.describe Api::V1::ChaptersController do
```
willfish, line 1: Promoting this to a request spec was one step too far

app/controllers/api/v1/chapters_controller.rb
```diff
@@ -3,46 +3,77 @@
...
 
-        render 'api/v1/changes/changes'
+        render json: serialized_result
       end
 
       private
 
-      def find_chapter
+      def chapter
         @chapter = Chapter.actual
```
jebw, line 54: `||=` - otherwise assignment is unnecesary

spec/controllers/api/v1/chapters_controller_spec.rb
jebw: Nah, better to minimise changes anyway :+1:

jebw (commented):
Looks great, only query is on variable assignment - api response is very quick!

jebw (approved):
LGTM :+1:

PR #1135, opened 2023-03-20
HOTT-2769 Allow overriding generating regulation for Measure factory

jebw:
### Jira link

HOTT-2769

### What?

I have added/removed/altered:

- [x] Extraction of factory changes from main 2769 branch

### Why?

I am doing this because:

- This will simplify the specs for HOTT-2874 

- Low - spec changes only


willfish (approved):
Makes sense. LGTM

PR #1139, opened 2023-03-21
HOTT-2911 Hide goods nomenclature from nested set

jebw:
### Jira link

HOTT-2911

### What?

I have added/removed/altered:

- [x] Exclude hidden goods_nomenclature from nested set descendant and children results

### Why?

I am doing this because:

- We don't want them to show up in the relationships
- It makes eager loading the relationships easier

- Removes the default ordering from the HiddenGoodsNomenclature model - I can't see a purpose since its never exposed to the end user to this and forces a subselect into the relationship join so just removing instead

### DIFF
--- a/app/models/goods_nomenclature.rb
+++ b/app/models/goods_nomenclature.rb
@@ -146,7 +146,7 @@ class GoodsNomenclature < Sequel::Model
     end
 
     def non_hidden
-      filter(Sequel.~(goods_nomenclature_item_id: HiddenGoodsNomenclature.codes))
+      filter(Sequel.~(goods_nomenclatures__goods_nomenclature_item_id: HiddenGoodsNomenclature.codes))
     end
 
     def indexable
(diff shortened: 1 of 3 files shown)


spec/models/goods_nomenclature_spec.rb
```diff
@@ -337,6 +350,28 @@
     it { is_expected.to all(have_attributes(producline_suffix: '80')) }
   end
 
+  describe '.exclude_hidden' do
+    subject { described_class.exclude_hidden.all }
+
+    before { hidden && comm2 }
+
+    let(:comm1) { create :commodity }
```
willfish, line 358: Won't you miss this if it isn't created beforehand?

```ruby
before do
  hidden_commodity
  non_hidden_commodity
end

let(:hidden_commodity) { create :commodity, goods_nomenclature_item_id: hidden.goods_nomenclature_item_id }
let(:non_hidden_commodity) { create :commodity }

let :hidden_goods_nomenclature { create :hidden_goods_nomenclature }
```

willfish (approved):

PR #1140, opened 2023-03-22
HOTT-2874: Display measures only if approved_flag is set to true in b…

maverickvi:
…ase_regulation/modification_regulation

### Jira link

[HOTT-<TODO>](https://transformuk.atlassian.net/browse/HOTT-2874)

### What?

I have added/removed/altered:

- [ ] Added conditions to Display measures only if approved_flag is set to true in base_regulation/modification_regulation

### Why?

I am doing this because:

- We don't want to display measures where approved_flag is set to false

### DIFF
--- a/spec/factories/measure_factory.rb
+++ b/spec/factories/measure_factory.rb
@@ -420,7 +420,7 @@ FactoryBot.define do
       measure_generating_regulation_role { Measure::MODIFICATION_REGULATION_ROLE }
 
       after(:build) do |measure, _evaluator|
-        create(:modification_regulation, :unapproved,  modification_regulation_id: measure.measure_generating_regulation_id)
+        create(:modification_regulation, :unapproved, modification_regulation_id: measure.measure_generating_regulation_id)
       end
     end
 
--- a/spec/factories/measure_factory.rb
+++ b/spec/factories/measure_factory.rb
@@ -121,6 +121,18 @@ FactoryBot.define do
       end
     end
 
+    trait :with_unapproved_base_regulation do
+      after(:create) do |measure, evaluator|
+        create(
+          :base_regulation,
+          :unapproved,
+          base_regulation_id: measure.measure_generating_regulation_id,
+          base_regulation_role: measure.measure_generating_regulation_role,
+          effective_end_date: evaluator.base_regulation_effective_end_date || Time.zone.today.in(10.years),
+        )
+      end
+    end
+
     trait :with_justification_regulation do
       after(:create) do |measure, _evaluator|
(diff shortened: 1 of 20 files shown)


spec/factories/measure_factory.rb
```diff
@@ -121,6 +121,18 @@
...
     end
 
+    trait :with_unapproved_base_regulation do
+      after(:create) do |measure, evaluator|
+        create(
+          :base_regulation,
+          :unapproved,
+          base_regulation_id: measure.measure_generating_regulation_id,
+          base_regulation_role: measure.measure_generating_regulation_role,
+          effective_end_date: evaluator.base_regulation_effective_end_date || Time.zone.today.in(10.years),
```
willfish, line 131:
```suggestion
          effective_end_date: evaluator.base_regulation_effective_end_date
```

willfish: 10 years in the future is akin to having no end date AFAIK

willfish:
Have you validated that measures/overview measures no long returns unapproved regulation measures using a real database?

spec/factories/measure_factory.rb
```diff
@@ -116,7 +116,19 @@
           :base_regulation,
           base_regulation_id: measure.measure_generating_regulation_id,
           base_regulation_role: measure.measure_generating_regulation_role,
-          effective_end_date: evaluator.base_regulation_effective_end_date || Time.zone.today.in(10.years),
+          effective_end_date: evaluator.base_regulation_effective_end_date,
+        )
+      end
+    end
+
+    trait :with_unapproved_base_regulation do
```
jebw, line 124: I'd push to replace the `after(:create)` with the already existing `generating_regulation` transient property because its a lot simpler

```ruby
generating_regulation { create :base_regulation, :unapproved }
```

Advantages are 
* the regulation gets created before the measure which depends upon it
* keeps logic creating regulations in the regulation factory
* the regulation is responsible for determining its own ids
* the measure just reuses the ids from the regulation
* you can ignore dates

spec/factories/measure_factory.rb
```diff
@@ -404,6 +416,14 @@
       end
     end
 
+    trait :with_unapproved_modification_regulation do
```
jebw, line 419: same as base regulation change - this block can get replaced with

```
generating_regulation { create :modification_regulation, :unapproved }
```

jebw (dismissed):
Looks really good - small suggested change to the factories to keep them maintainable

spec/factories/measure_factory.rb
willfish: Amazing. Thanks Jeremy

PR #1141, opened 2023-03-23
HOTT-2690 Fix for after_load step

jebw:
### Jira link

HOTT-2690

### What?

I have added/removed/altered:

- [x] Fix after loader when there is a complex eager load tree

### Why?

I am doing this because:

- If the internal `@associations` variable was already populated, then `ns_ancestors` was not being populated, but `ns_parent` was
- Theoretically the same may have happened with descendants but I didn't observer that

- Low, not currently in use - fixes existing bug


willfish (approved):
Makes sense

PR #1142, opened 2023-03-24
HOTT-2915 Fix sorting for NestedSet#applicable_measures

jebw:
### Jira link

HOTT-2915

### What?

I have added/removed/altered:

- [x] Added `Measure#sort_key` to allow sorting measures in ruby the same way we do in SQL
- [x] Change `NestedSet#applicable_measures` to sort the measures using the new `#sort_key`
- [x] Change `NestedSet#applicable_overview_measures` to sort the measures using the new `#sort_key`

### Why?

I am doing this because:

- We want to maintain the same ordering for the new `#applicable_measures` as the existing `#measures` it replaces
- We want to maintain the same ordering for the new `#applicable_overview_measures` as the existing `#overview_measures` it replaces

- Low, impacts nested set which currently is not in use

### DIFF
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -18,6 +18,8 @@ class Measure < Sequel::Model
     DEFINITIVE_ANTIDUMPING_ROLE,
   ].freeze
 
+  FAR_FUTURE_END_DATE = Date.parse('3023-12-31').end_of_day.freeze
+
   set_primary_key [:measure_sid]
 
   plugin :time_machine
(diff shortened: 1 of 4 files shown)


app/models/measure.rb
```diff
@@ -553,6 +555,19 @@ def meursing_measures
...
+  def sort_key
+    end_date_key = values.key?(:effective_end_date) ? :effective_end_date : :validity_end_date
+
+    [
+      geographical_area_id,
+      measure_type_id,
+      additional_code_type_id,
+      additional_code_id,
+      ordernumber,
+      values[end_date_key] || FAR_FUTURE_END_DATE,
```
willfish, line 567: Lol. I'm waiting for 3023 like :skull_and_crossbones:

willfish (approved):
Y2K

app/models/measure.rb
jebw: Haha - yeh, it was a bit of a hacky solution to ruby unable to compare a date to a nil - SQL just orders it at the end so I picked a (very) far off date :)

PR #1144, opened 2023-03-24
HOTT-2919 Use nested set for subheadings

jebw:
### Jira link

HOTT-2919

### What?

I have added/removed/altered:

- [x] Use nested set for subheadings endpoint

### Why?

I am doing this because:

- it is faster and generates less queries

- Mid - feature flagged, but does change the implementation of an important endpoint


jebw:
Already run a comparison script on ~60% of json responses and they are the same so pretty happy from that PoV - obviously its feature flagged so merging doesn't mean its necessarily getting turned on just yet

PR #1145, opened 2023-03-27
Converted nightly CSV to nested set

jebw:
### Jira link

HOTT-2927

### What?

I have added/removed/altered:

- [x] Converted the 'all goodsnomenclatures' csv generation to use nested set

### Why?

I am doing this because:

- It is faster and generates less queries
- It allows for good comparison of output between existing and nested set approaches
- It excludes hidden goods nomenclatures from the output
- It includes Heading 7019 in the ancestors list for commodities under Heading 7019

- Low - easy to compare/contrast new versus old, actual generation task runs in a background job

### DIFF
--- a/app/serializers/api/admin/csv/goods_nomenclature_serializer.rb
+++ b/app/serializers/api/admin/csv/goods_nomenclature_serializer.rb
@@ -10,17 +10,19 @@ module Api
         column :description, column_name: 'Description'
         column :validity_start_date, column_name: 'Start date'
         column :validity_end_date, column_name: 'End date'
-        column :number_indents, column_name: 'Indentation'
-        column :end_line, column_name: 'End line', &:path_declarable?
+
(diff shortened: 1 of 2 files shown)


app/serializers/api/admin/csv/goods_nomenclature_serializer.rb
```diff
@@ -10,17 +10,19 @@ class GoodsNomenclatureSerializer
         column :description, column_name: 'Description'
         column :validity_start_date, column_name: 'Start date'
         column :validity_end_date, column_name: 'End date'
-        column :number_indents, column_name: 'Indentation'
-        column :end_line, column_name: 'End line', &:path_declarable?
+
+        column(:number_indents, column_name: 'Indentation') do |goods_nomenclature|
```
willfish, line 14: I think you could add this to the goods_nomenclature model as ns_number_indents/equivalent?

willfish (approved):

PR #1157, opened 2023-04-03
HOTT-2977: Disable database workflows until we're ready

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2977

### What?

I have added/removed/altered:

- [x] Disable database workflows
- [x] Remove superfluous workflow

### Why?

I am doing this because:

- We need to consolidate around a single approach to managing the database in all spaces otherwise we're fighting against ourselves
- The removed workflow was to try and make sure that staging had the same search references as production. We have since streamlined this so this is redundant


jebw (approved):
LGTM

PR #1162, opened 2023-04-04
HOTT-2985: Use leaf to determine declarability of commodity

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2985

image

### What?

I have added/removed/altered:

- [x] Added new declarable implementation to the heading commodity serializer
- [x] Added new specs for the commodity serializer

### Why?

I am doing this because:

- This is required because materialized path requires time machine is applied on the headings cache in order to determine declarability. We've opted to just use leaf over fast_declarable? to avoid this requirement


jebw (approved):
LGTM

PR #1174, opened 2023-04-12
HOTT-2992 Fix footnote ids

jebw:
### Jira link

HOTT-2992

### What?

I have added/removed/altered:

- [x] Use full footnote id in Headings endpoints
- [x] Use full footnote id in Subheadings endpoint
- [x] Use full footnote id in Commodities

### Why?

I am doing this because:

- When using only the `Footnote#footnote_id` discreet footnotes which share a `footnote_id` but with a different `footnote_type_id` get deduplicated during serialization to jsonapi, eg so **PN001** and **TN001** will get deduplicated to a single foot note which is incorrect.
- The commodities endpoint actually used a mix of `Footnote#id` and `Footnote#footnote_id` for different footnotes further confusing things

- Changes the ids used for footnotes in the jsonapi response on major endpoints.

### DIFF
--- a/app/presenters/api/v2/commodities/commodity_presenter.rb
+++ b/app/presenters/api/v2/commodities/commodity_presenter.rb
@@ -29,7 +29,7 @@ module Api
         end
 
         def footnote_ids
-          footnotes.map(&:footnote_id)
+          footnotes.map(&:id)
         end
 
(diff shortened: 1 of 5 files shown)


willfish (approved):
Wayyyy cleaner

PR #1176, opened 2023-04-13
HOTT-3031: Migrations: Adds full chemicals table and search suggestion columns

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3031

### What?

I have added/removed/altered:

- [x] Adds full-chemicals csv file
- [x] Adds full chemicals and adjustments to search suggestions

### Why?

I am doing this because:

- These are needed to augment search


db/structure.sql
```diff
@@ -12,79 +12,261 @@ SET client_encoding = 'UTF8';
...
+$$;
 
 
 --
--- Name: EXTENSION "uuid-ossp"; Type: COMMENT; Schema: -; Owner: -
+-- Name: reassign_owned(); Type: FUNCTION; Schema: public; Owner: -
 --
 
-COMMENT ON EXTENSION "uuid-ossp" IS 'generate universally unique identifiers (UUIDs)';
+CREATE FUNCTION public.reassign_owned() RETURNS event_trigger
```
willfish, line 137: These functions and their corresponding event triggers are actually managed by rds and can be safely ignored

willfish: I'm okay adding them here to just avoid noisy schema changes in future

db/structure.sql
```diff
@@ -2528,17 +2716,37 @@ CREATE SEQUENCE public.fts_regulation_actions_oid_seq
...
+ALTER SEQUENCE uk.fts_regulation_actions_oid_seq OWNED BY uk.fts_regulation_actions_oplog.oid;
 
 
 --
--- Name: full_temporary_stop_regulations_oplog; Type: TABLE; Schema: public; Owner: -
+-- Name: full_chemicals; Type: TABLE; Schema: uk; Owner: -
 --
 
-CREATE TABLE public.full_temporary_stop_regulations_oplog (
+CREATE TABLE uk.full_chemicals (
```
jebw, line 2729: Looks like this shouldn't be in here? Isn't this from the chemicals PR?

jebw (changes requested):
Looks good but appears to have been generated from an existing DB since its got the 'chemicals' table in there

jebw (approved):
Awesome work - really like the the 'it just works' approach for `db:structure:dump`

PR #1178, opened 2023-04-13
HOTT-2941 Use nested set for GoodsNomenclature CSV endpoints

jebw:
### Jira link

HOTT-2941

### What?

I have added/removed/altered:

- [x] Use nested set to retrieve goods nomenclatures for serialization

### Why?

I am doing this because:

- It will allow effeciently determining things like `declarable?` values which is needed for later tickets

- Changes headings output to not include the 'grouping' parent heading - this appears to be a bug in the existing implementation though


willfish (approved):
LGTM

PR #1180, opened 2023-04-14
HOTT-3033: Adds full chemical populator and basic validations

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3033

### What?

I have added/removed/altered:

- [x] Added full chemical model, populator service and specs

### Why?

I am doing this because:

- This enables us to import chemicals and use them in search later

### Out of scope

I don't need a temporary rake task to run this the once. I've optimised it a bunch in anticipation of moving to scraping these from the SOAP api later.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -76,6 +76,7 @@ group :test do
   gem 'rspec-json_expectations'
   gem 'rspec_junit_formatter'
   gem 'rspec-rails'
+  gem 'sequel_spec'
   gem 'simplecov', require: false
   gem 'webmock'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -76,7 +76,6 @@ group :test do
   gem 'rspec-json_expectations'
   gem 'rspec_junit_formatter'
   gem 'rspec-rails'
-  gem 'sequel_spec'
   gem 'simplecov', require: false
   gem 'webmock'
 end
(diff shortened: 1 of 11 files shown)


app/services/full_chemical_populator_service.rb
```diff
@@ -0,0 +1,77 @@
+class FullChemicalPopulatorService
+  CSV_FILE = Rails.root.join('db/full_chemicals.csv')
+
+  def call
+    FullChemical.unrestrict_primary_key
+
+    full_chemicals.each_slice(5000) do |values|
```
jebw, line 7: No idea how much memory balloons (from memory this 65k csv rows?? is that right) up to with this but you could read CSVs iteratively (ie `CSV.foreach`) - build those into an array and flush (INSERT) to db every 5k records the once after the iteration finishes?

Gemfile
```diff
@@ -76,6 +76,7 @@ group :test do
   gem 'rspec-json_expectations'
   gem 'rspec_junit_formatter'
   gem 'rspec-rails'
+  gem 'sequel_spec'
```
jebw, line 79: Mildly cautious of including a gem which hasn't been changed in 5 years and its last significant change was 6 years ago.

It does allow writing much nicer validation specs though so :man_shrugging: 

An alternative for the validations - not as nice or thorough but still fairly compact with discrete tests for different attributes
* https://github.com/trade-tariff/trade-tariff-backend/blob/main/spec/models/news/item_spec.rb#L18

jebw (approved):
LGTM - couple of comments around gem and pulling the entire CSV into RAM in a single go?

app/services/full_chemical_populator_service.rb
willfish: It significant enough to justify this change I think

PR #1184, opened 2023-04-17
HOTT-3032: Populate and filter search suggestions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3032

### What?

I have added/removed/altered:

- [x] Added full chemicals when populating search suggestions
- [x] Make use of types (behind a feature flag) when filtering suggestions for the numeric and text type searches
- [x] Updated test coverage 
- [x] Populated environment variables

### Why?

I am doing this because:

- This enables us to use the recently populated types to filter the
specific kinds of search suggestions.
- It also improves performance by removing the expensive similarity checks where 99% of the value comes from the ilike


app/services/api/v2/suggestions_service.rb
```diff
@@ -1,22 +1,120 @@
 module Api
   module V2
```
willfish, line 2: Referencing the api version for this global service makes 0 sense since no V1 api exists and this is global

maverickvi (approved):

willfish:
Okay, just validated on dev:

1. Using the search suggestions without the type populated
2. Populated the new suggestions and with feature flag off
3. Populated the new suggestions and with feature flag on

All working as expected

PR #1189, opened 2023-04-18
HOTT-3040 Include negative measure conditions

jebw:
### Jira link

HOTT-3040

### What?

I have added/removed/altered:

- [x] Stop filtering negative measure conditions from permutation groups if they have a document code

### Why?

I am doing this because:

- Whilst the data is invalid DBT would prefer that we expose it to our users

- Low, exposes addition measure conditions in permutation groups which is an internal api extension

### Before

image

### After

image


willfish (approved):
Would be good to screenshot the before/after UI experience

PR #1190, opened 2023-04-18
HOTT-2987 Precache headings and subheadings

jebw:
### Jira link

HOTT-2987

### What?

I have added/removed/altered:

- [x] Added a service and worker to precache the jsonapi output of all non-declarable headings
- [x] Extended to also precache subheadings
- [x] Schedule worker to also cache all of tomorrows headings and subheadings at 10pm UTC
- [x] Sets default order for commodities footnotes relationship

### Why?

I am doing this because:

- It should reduce load on postgres at the start of the day
- Footnotes order ensures consistency of presentation across pages, particularly when eager loading

- Low, if nested set headings/subheadings are used it will speed them up, if not it will just iterate through the chapters without caching anything

### DIFF
--- a/app/models/goods_nomenclature.rb
+++ b/app/models/goods_nomenclature.rb
@@ -104,7 +104,8 @@ class GoodsNomenclature < Sequel::Model
                            left_primary_key: :goods_nomenclature_sid,
                            left_key: :goods_nomenclature_sid,
                            right_key: %i[footnote_type footnote_id],
-                           right_primary_key: %i[footnote_type_id footnote_id] do |ds|
+                           right_primary_key: %i[footnote_type_id footnote_id],
+                           order: %i[footnote_type_id footnote_id] do |ds|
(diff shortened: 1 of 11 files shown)


app/models/goods_nomenclature.rb
```diff
@@ -104,7 +104,8 @@ class GoodsNomenclature < Sequel::Model
                            left_primary_key: :goods_nomenclature_sid,
                            left_key: :goods_nomenclature_sid,
                            right_key: %i[footnote_type footnote_id],
-                           right_primary_key: %i[footnote_type_id footnote_id] do |ds|
+                           right_primary_key: %i[footnote_type_id footnote_id],
```
willfish, line 107: Has this change snuck into this PR? I mean it looks correct to me to order based on the composite primary keys

willfish (approved):
LGTM. Quite a lot of conditional behaviour that we can gut once we're on nested set wholesale :tada:

PR #1195, opened 2023-04-24
HOTT-3075: Adds handling for non-zip content when downloading CDS files

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3075

image

image

### What?

I have added/removed/altered:

- [x] Prevent non-zip files in the CDS download process from being persisted to S3

### Why?

I am doing this because:

- We have a repeating issue with CDS files which are not actually zip files. This reduces to burden to rollback/rollforward which can now just be done in the admin UI

### DIFF
--- a/app/lib/tariff_synchronizer/tariff_downloader.rb
+++ b/app/lib/tariff_synchronizer/tariff_downloader.rb
@@ -85,7 +85,7 @@ module TariffSynchronizer
     end
 
     def write_update_file(response_body)
-      handle_zip_verification(response_body) do
+      if should_write_file?(response_body)
         FileService.write_file(file_path, response_body)
 
         instrument('downloaded_tariff_update.tariff_synchronizer',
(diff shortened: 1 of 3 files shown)


willfish:
For the curious, a zip file is made up of the following:

```
      local file header signature     4 bytes  (0x04034b50)
      version needed to extract       2 bytes
      general purpose bit flag        2 bytes
      compression method              2 bytes
      last mod file time              2 bytes
      last mod file date              2 bytes
      crc-32                          4 bytes
      compressed size                 4 bytes
      uncompressed size               4 bytes
      file name length                2 bytes
      extra field length              2 bytes

      file name (variable size)
      extra field (variable size)
```

The local file header signature is present even in empty zips and hasn't changed over time in the definition of a zip file so this should cover all scenarios where we actually have a zip file and not garbage HTML content

jebw (approved):
LGTM

PR #1196, opened 2023-04-24
HOTT-2210 Sidekiq 7 upgrade

jebw:
### Jira link

HOTT-2210

### What?

I have added/removed/altered:

- [x] Upgrade to Redis 5 and Redis-client
- [x] Lock versions to redis >= 5.0.6 and and redis-client >= 0.11.2
- [x] Drop redis-elasticache
- [x] Upgraded sidekiq-scheduler
- [x] Upgraded sidekiq to v7
- [x] Upgraded redlock
- [x] Updated sidekiq.yml for new scheduler config key

### Why?

I am doing this because:

- Redis-Elasticache is incompatible with the Redis 5 rubygem
- Its purpose is to reconnect on READONLY errors caused by a failover in the elasticache cluster
- That functionality is now provided by the redis-client (>= 0.11.2) and redis (>= 5.0.6) gems
- Upgraded sidekiq-scheduler to be compatible with Sidekiq 7 and Redis 5 gem
- Upgraded redlock to be compatible with Redis 5 gem

- Replaces a key component
- Removes the `redis-elasticache` gem which had solved our errors from the frequent failures from redis at weekends - my reading of the docs implies this should be a 'solved issue' now even without that gem

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -416,7 +416,7 @@ jobs:
           POSTGRES_USER: postgres
           POSTGRES_DB: tariff_test
           POSTGRES_PASSWORD: postgres
-      - image: circleci/redis:4.0.9
+      - image: cimg/redis:6.2
         environment:
           REDIS_URL: "redis://localhost:6379/"
       - image: opensearchproject/opensearch:1.2.4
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -416,7 +416,7 @@ jobs:
           POSTGRES_USER: postgres
           POSTGRES_DB: tariff_test
           POSTGRES_PASSWORD: postgres
-      - image: circleci/redis:4.0.9
+      - image: cimg/redis:6.2
```
willfish, line 419: Yay \0/

willfish (approved):
LGTM. Thanks for sorting this :tada:

PR #1203, opened 2023-04-25
HOTT-3064 Remove old headings caching into elastic search

jebw:
### Jira link

HOTT-3064

### What?

I have added/removed/altered:

- [x] Removed the feature flags for nested set headings and nested set subheadings
- [x] Removed non-nested set code paths from Subheading serializer
- [x] Removed non-nested set code paths from `CachedSubheadingService`
- [x] Removed the `PrewarmSubheadingsWorker`
- [x] Removed non-nested set serialization for Headings
- [x] Removed the CachedHeadingService
- [x] Removed the Cache::HeadingIndex
- [x] Removed the Cache::HeadingSerializer
- [x] Removed `AnnotatedCommodityService`
- [x] Converted the `/headings/1234/commodities.csv`  to nested set

### Why?

I am doing this because:

- This is now all dead code so removal reduces our maintenance requirements
- Heading Commodities CSV was still using CachedHeadingService and blocked removal of the rest, new implementation is approx 3x faster on my machine
- Should remove 90 minutes of overnight recaching jobs

- Medium risk, AFAICT this should all be dead code


app/services/heading_service/heading_serialization_service.rb
```diff
@@ -51,12 +48,9 @@ def serialization_service
       if heading.declarable?
         HeadingService::Serialization::DeclarableService
```
willfish, line 49:
```suggestion
        HeadingService::Serialization::NsDeclarableService
```

willfish (approved):
giphy

app/services/heading_service/heading_serialization_service.rb
```diff
@@ -51,12 +48,9 @@ def serialization_service
       if heading.declarable?
```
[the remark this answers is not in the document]
jebw: Yeh, good point - I didn't depend upon it originally because it was all feature flagged but I can now

In general the loading logic is a bit odd because of how the caching is handled, we load outside the cache, then reload with a bunch of eager loading inside the cache block. I need to have a rethink about this, possibly using `yield` but in the mean time, yes using ns_declarable would definitely be beneficial

app/services/heading_service/heading_serialization_service.rb
jebw: Oops - this should be erroring like mad :thinking:

app/services/cached_subheading_service.rb
```diff
@@ -30,54 +30,13 @@ def call
...
-
-      AnnotatedCommodityService.new(presented).call
-
-      presented
-    end
-  end
-
-  def cache_version
-    "-v#{CACHE_VERSION}" if TradeTariffBackend.nested_set_subheadings?
+    Api::V2::Subheadings::SubheadingPresenter.new(ns_eager_loaded_subheading)
```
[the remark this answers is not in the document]
jebw: Probably not just yet - this is reloaded to eager load all the associate data. I'll do a follow up PR to rejig the data fetching not I'm not needing to handle branching

app/services/heading_service/heading_serialization_service.rb
jebw: Ok, realised what this is - this one is because there was originally `DeclarableService` and `NondeclarableService`, then `NsNondeclarableService` got added, and now `NondeclarableService` has been removed

I'm going to leave this as is for now I think because the `DeclarableService` is more about a 'commodities' end point for headings and I suspect all of this will get further rejigged in a future PR. If I leave it as is for now it at least makes this PR clear whats being removed rather then introducing renames alongside everything else.

But I'll do a follow up to standardise all this once the dust has settled

willfish (approved):

PR #1204, opened 2023-04-26
HOTT-3026 overriding tree nodes depth

jebw:
### Jira link

HOTT-3026

### What?

I have added/removed/altered:

- [x] Added a TreeNodeOverrides model + table
- [x] Changed the `goods_nomenclatures_tree_nodes` to favour the newest of the indent from the overrides table and the one from the indents view

### Why?

I am doing this because:

- This provides an 'escape hatch' to allow us to correct errors in the indents provided by CDS data without actually changing the indents data
- Which in turn means we can continue to present the CDS indents value to the user (even if its incorrect) whilst also ensuring the elements are at the correct location in the tree

- Drops and recreates the tree nodes view which may cause a few errors temporarily - the retry handling should cover these anyway but the tree nodes view is used by many endpoints just yet


db/structure.sql
```diff
@@ -9986,6 +10022,27 @@ CREATE INDEX gono_succ_pk ON uk.goods_nomenclature_successors_oplog USING btree
...
+--
+
+CREATE INDEX goods_nomenclature_tree_node_overrides_created_at_index ON uk.goods_nomenclature_tree_node_overrides USING btree (created_at);
+
+
+--
+-- Name: goods_nomenclature_tree_node_overrides_goods_nomenclature_inden; Type: INDEX; Schema: uk; Owner: -
+--
+
+CREATE UNIQUE INDEX goods_nomenclature_tree_node_overrides_goods_nomenclature_inden ON uk.goods_nomenclature_tree_node_overrides USING btree (goods_nomenclature_indent_sid);
```
willfish, line 10036: I'm guessing we hit a character limit, here (re inden)

jebw:
Challenge is even if the indents data is wrong - if thats what we're getting from upstream its what we should be publishing 

_but_

we still need an alternative source if we're to correct the impact of broken indents data on the tree hierarchy

This means in the scenario where the indent jumps from 3 to 5 say between levels, we can fix our representation of the tree but still show the indents as being 3 and 5 respectively to the api consumer

PR #1205, opened 2023-04-26
HOTT-3025 Grouping headings

jebw:
### Jira link

HOTT-3025

### What?

I have added/removed/altered:

- [x] Replaced the tree_nodes view with one which takes into account grouping headings

### Why?

I am doing this because:

- Whilst there is a lot of inconsistency in the code base, general conclusion is that grouping headings should be above their non-grouping equivalents rather then adjacent to them
- This is how GoodsNomenclatureMapper handles it _if_ given a full tree of nodes, which it only is for the chapters endpoint
- Tarics UI nests non-grouped headings under grouped
- It is easier to drop and replace the view now whilst it has limited usage rather then later when it has heavy usage

### Technical notes

Technically this is incorrect! 

Prior to now the position in the tree of every node could be decided from the information provided by that indent record itself - this is necessary for time machine compatibility.

That is no longer possible, every node needs to know how many grouping headings are above it in order to know its position. The approach I take is to look for available grouping headings at the goods_nomenclature_indent validity_start_date - this does mean if the number of grouping headings in the hierarchy above changes over the lifetime of that indent then the depth value becomes incorrect. 

I've concluded a SQL implementation is irreconcilable with 'correct behaviour' and since the grouping indents have _never_ changed over the lifetime of the tariff (ie 1970's onwards) I don't see this as a significant risk. If a new grouping heading is published, and new indents are published for the goods_nomenclatures below it then it would also not be a problem.

### Deployment risks

- Drops and replaces the tree nodes view - this isn't heavily used yet so shouldn't cause any issues
- This is built to modify the migration from an earlier PR because its better to do this as a single migration

Screenshot from 2023-04-21 10-22-15

### DIFF
--- a/app/presenters/api/v2/subheadings/subheading_presenter.rb
+++ b/app/presenters/api/v2/subheadings/subheading_presenter.rb
@@ -42,7 +42,7 @@ module Api
         end
 
         def heading
-          @heading ||= ns_ancestors.find { |a| a.is_a? Heading }
+          @heading ||= ns_ancestors.reverse.find { |a| a.is_a? Heading }
         end
       end
     end
--- a/db/migrate/20230419084212_add_tree_nodes_overrides.rb
+++ b/db/migrate/20230419084212_add_tree_nodes_overrides.rb
@@ -25,7 +25,10 @@ Sequel.migration do
         indents.validity_start_date,
         COALESCE(indents.validity_end_date, MIN(replacement_indents.validity_start_date) - INTERVAL '1 second', nomenclatures.validity_end_date) as validity_end_date,
         indents.oid,
-        COALESCE(overrides.depth, indents.number_indents + 2 - (indents.goods_nomenclature_item_id LIKE '%00000000' AND indents.number_indents = 0)::integer) AS "depth"
+        COALESCE(
+          overrides.depth,
+          indents.number_indents + 2 - (indents.goods_nomenclature_item_id LIKE '%00000000' AND indents.number_indents = 0)::integer + COUNT(grouping_headings.goods_nomenclature_sid)
+        ) AS "depth"
       FROM goods_nomenclature_indents indents
       INNER JOIN goods_nomenclatures nomenclatures ON
         indents.goods_nomenclature_sid = nomenclatures.goods_nomenclature_sid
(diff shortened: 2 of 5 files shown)


db/migrate/20230419084212_add_tree_nodes_overrides.rb
```diff
@@ -25,7 +25,10 @@
         indents.validity_start_date,
         COALESCE(indents.validity_end_date, MIN(replacement_indents.validity_start_date) - INTERVAL '1 second', nomenclatures.validity_end_date) as validity_end_date,
         indents.oid,
-        COALESCE(overrides.depth, indents.number_indents + 2 - (indents.goods_nomenclature_item_id LIKE '%00000000' AND indents.number_indents = 0)::integer) AS "depth"
+        COALESCE(
```
willfish, line 28: Won't this migration have already ran in development, staging, production spaces?

app/presenters/api/v2/subheadings/subheading_presenter.rb
```diff
@@ -42,7 +42,7 @@ def chapter
         end
 
         def heading
-          @heading ||= ns_ancestors.find { |a| a.is_a? Heading }
+          @heading ||= ns_ancestors.reverse.find { |a| a.is_a? Heading }
```
willfish, line 45:
```suggestion
          @heading ||= ns_ancestors_dataset.declarable.find { |a| a.is_a? Heading }
```

willfish: Its a micro optimisation for slightly large numbers of ancestors I think. No strong feeling re this and I actually feel like yours is more readable

db/migrate/20230419084212_add_tree_nodes_overrides.rb
```diff
@@ -36,6 +39,14 @@
       LEFT JOIN goods_nomenclature_tree_node_overrides overrides ON
         indents.goods_nomenclature_indent_sid = overrides.goods_nomenclature_indent_sid
         AND indents.operation_date < coalesce(overrides.updated_at, overrides.created_at)
+      LEFT JOIN goods_nomenclatures grouping_headings ON
+        indents.goods_nomenclature_item_id NOT LIKE '%00000000'
+        AND grouping_headings.producline_suffix < '80'
```
willfish, line 44: So from memory 100% of grouping headings are pls of 10. I guess we want to handle the future where there can be multiple grouping headings

willfish: And maybe < is more performant on a presorted list of goods nomenclatures

willfish (approved):
LGTM :tada:. I guess its best to make this change now/reflect what the EU site does rather than do it later.

db/migrate/20230419084212_add_tree_nodes_overrides.rb
jebw: Normally yes, but since they'll go out into the same deployment - then no - I'm actually going to cancel CI on the other PR to stop it deploying to staging

Its just avoiding having the two changes as two transactions - I'd split the PRs just for ease of code review but originally they were a single PR

app/presenters/api/v2/subheadings/subheading_presenter.rb
jebw: `ns_ancestors_dataset` isn't already loaded where as `ns_ancestors` is . In the precache scenario we load the chapter once, then feed the different headings and subheadings under it into the serializer - so this would introduce a N+1 query. I'll leave it as is I think because it'll be quicker to reverse the list then do a db query

db/migrate/20230419084212_add_tree_nodes_overrides.rb
jebw: You're right they are at present and I had a debate around this one. The GoodsNomenclatureMapper has logic for multiple levels of grouping subheading - I tested the query both with on PLS=10 and the 'count grouping headings' approach and there was no performance difference so I'd opted to follow the existing code.

If they are never anything other then 10 - it'll have no impact, if they are then it'll work (I did some manual testing to make sure it behaves as expected)

jebw:
Yeh - I'm mildly skeptical of its benefit now but do think if its going to get done - makes sense to do before its massively in use

PR #1209, opened 2023-04-27
Fix flaky additional code specs

jebw:
### Jira link

HOTT-???

### What?

I have added/removed/altered:

- [x] Fixed the AdditionalCode factory output to not choose an `additional_code_type_id` excluded by the Cache::AdditionalCodeIndex
- [x] Changed Measure factory to not set `additional_code_type_id` to a value potentially excluded by Cache::AdditionalCodeIndex
- [x] Reenabled specs marked as flaky

### Why?

I am doing this because:

- It makes any spec using the AdditionalCode factory and relying on the behaviour of CacheAdditionalCodeIndex#dataset flaky

- None, spec only changes

### Notes

Addresses these failures on CI Job 23062


willfish (approved):
:superhero:

PR #1210, opened 2023-04-28
HOTT-1622: Exposed origins and origin exclusions on quota search API

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1622

### What?

I have added/removed/altered:

- [ ] Exposed origins and origin exclusions on quotas search API

### Why?

I am doing this because:

-  We need access to this data on the API

### Have you? (optional)

Screenshot 2023-04-28 at 11 20 17

Screenshot 2023-04-28 at 11 17 52


app/models/quota_order_number_origin_exclusion.rb
```diff
@@ -3,4 +3,16 @@ class QuotaOrderNumberOriginExclusion < Sequel::Model
                                  excluded_geographical_area_sid]
 
   set_primary_key %i[quota_order_number_origin_sid excluded_geographical_area_sid]
+
+  many_to_one :geographical_area, key: :excluded_geographical_area_sid,
+                                  primary_key: :geographical_area_sid,
+                                  class_name: 'GeographicalArea' do |ds|
```
willfish, line 9: You shouldn't need the class name, here

app/models/quota_order_number_origin_exclusion.rb
```diff
@@ -3,4 +3,16 @@ class QuotaOrderNumberOriginExclusion < Sequel::Model
...
+    ds.with_actual(GeographicalArea)
+  end
+
+  delegate :geographical_area_id, to: :geographical_area
+
+  def id
```
willfish, line 15: You don't need this method

app/serializers/api/v2/quota_order_numbers/quota_order_number_origin_serializer.rb
```diff
@@ -0,0 +1,19 @@
+module Api
+  module V2
+    module QuotaOrderNumbers
+      class QuotaOrderNumberOriginSerializer
+        include JSONAPI::Serializer
+
+        set_type :quota_order_number_origin
+
+        set_id :quota_order_number_origin_sid
```
willfish, line 9:
```suggestion
        set_id :id
```

spec/models/quota_order_number_origin_exclusion_spec.rb
```diff
@@ -0,0 +1,7 @@
+RSpec.describe QuotaOrderNumberOriginExclusion do
+  describe '#id' do
+    subject(:exclusion) { build(:quota_order_number_origin_exclusion, quota_order_number_origin_sid: 111, excluded_geographical_area_sid: 222) }
+
+    it { expect(exclusion.id).to eq('111-222') }
```
willfish, line 5:
```suggestion
    it { expect(exclusion.id).to eq("#{exclusion.oid}-111-222") }
```

willfish (changes requested):
We shouldn't duplicate the id implementation

willfish:
We'll want to extend the eager load list in the quota service to make sure we're not introducing n+1s around these relationshiips.

willfish (approved):
Having reviewed the n+1 issues we've actually fixed a whole bunch of quite bad ones and this is a massive speed up despite despite actually adding more queries.

PR #1218, opened 2023-05-04
HOTT-3118 Add leaf virtual column

jebw:
### Jira link

HOTT-3118

### What?

I have added/removed/altered:

- [x] Added a dataset scope to add a virtual `leaf` column to GoodsNomenclature

### Why?

I am doing this because:

- It allows `ns_declarable?` to determine whether a commodity is declarable without requiring a database lookup or retrieving all its children

- Low, not currently in use - will be used in following PRs

### Performance benefit

This is checking `ns_declarable?` for every GoodsNomenclature in our database

Screenshot from 2023-05-04 11-08-09

### DIFF
--- a/app/models/goods_nomenclatures/nested_set.rb
+++ b/app/models/goods_nomenclatures/nested_set.rb
@@ -124,6 +124,15 @@ module GoodsNomenclatures
           .without_excluded_types
           .overview
       end
+
+      def_column_accessor :leaf
+
+      dataset_module do
+        def with_leaf_column
+          association_join(tree_node: proc { |ds| ds.with_leaf_column })
+            .select_append(:number_indents, :depth, :leaf)
+        end
+      end
     end
 
     def recursive_ancestor_populator(ancestors)
--- a/app/models/goods_nomenclatures/nested_set.rb
+++ b/app/models/goods_nomenclatures/nested_set.rb
@@ -8,6 +8,7 @@ module GoodsNomenclatures
       one_to_one :tree_node, key: :goods_nomenclature_sid,
                              class_name: 'GoodsNomenclatures::TreeNode',
                              reciprocal: :goods_nomenclature,
+                             graph_use_association_block: true,
                              read_only: true do |ds|
         ds.with_actual(GoodsNomenclatures::TreeNode)
       end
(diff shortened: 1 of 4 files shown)


app/models/goods_nomenclatures/nested_set.rb
```diff
@@ -124,6 +124,15 @@ module NestedSet
           .without_excluded_types
           .overview
       end
+
+      def_column_accessor :leaf
```
willfish, line 128: TIL

app/models/goods_nomenclatures/tree_node.rb
```diff
@@ -64,10 +70,34 @@ def ancestor_node_constraints(origin, ancestors)
...
+          .select_append(
+            (
+              Sequel.function(:count, :child_nodes__goods_nomenclature_indent_sid) =~ 0
+            ).as(:leaf),
+          )
+          .group(columns.map { |col| Sequel.qualify(model.table_name, col) })
+          .left_join(
+            model.table_name,
+            (
+              (children.depth =~ (origin.depth + 1)) &
```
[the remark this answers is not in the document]
willfish: I'm wondering if we could consolidate :thinking:

willfish (approved):

app/models/goods_nomenclatures/nested_set.rb
```diff
@@ -8,6 +8,7 @@ module NestedSet
       one_to_one :tree_node, key: :goods_nomenclature_sid,
                              class_name: 'GoodsNomenclatures::TreeNode',
                              reciprocal: :goods_nomenclature,
+                             graph_use_association_block: true,
```
willfish, line 11: Que pasa?

jebw: Normally the 'block' is ignored when a relationship is `JOIN`ed - hence why all the time machine stuff gets ignored with `#eager_graph`

This is an option toggle that behaviour which `Sequel` brought in in the January 2023 release. The default behaviour is a bit counter productive for our use case but would likely cause subtle issues if we turned it on everywhere. 

The relationship is new and I'm confident it wont cause problems and solves me needing to set time machine dates when using `association_join`.

willfish: Thanks Jeremy. Very useful :)

app/models/goods_nomenclatures/nested_set.rb
```diff
@@ -124,6 +125,24 @@ module NestedSet
...
+          association_join(tree_node: proc { |ds| ds.join_child_sids })
+            .select_all(:goods_nomenclatures)
+            .select_append(:tree_node__number_indents, :tree_node__depth)
+            .select_append((Sequel.function(:count, :tree_node__child_sid) =~ 0).as(:leaf))
+            .group(qualified_columns + %i[tree_node__number_indents tree_node__depth])
+        end
+
+      private
+
+        def qualified_columns(qualifier = model.table_name)
```
willfish, line 142: Feels like something that should be builtin

willfish (approved):

app/models/goods_nomenclatures/nested_set.rb
jebw: Agreed, couldn't find it in the docs though so :man_shrugging:

PR #1219, opened 2023-05-04
HOTT-3048: Removes redundant legacy search behaviour

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3048

### What?

I have added/removed/altered:

- [x] Supports short codes for subheadings instead of automatically assuming these are partial commodities
- [x] Kills redundant behaviour around single digit chapter behaviour (which Matt agrees is stupid)
- [x] Moves to relying on search suggestions exclusively when doing exact match
- [x] Updates the specs to reflect the reliance on suggestions

### Why?

I am doing this because:

- This is required to support subheadings search suggestions and significantly reduces the complexity of the exact search behaviour


spec/factories/goods_nomenclature_factory.rb
```diff
@@ -236,6 +236,10 @@
       validity_end_date   { 1.year.ago.beginning_of_day  }
     end
 
+    trait :with_heading do
```
jebw, line 239: Am I missing something - this feels like it would confuse a future developer?

If I call `factory :goods_nomenclature, :with_heading` - I'd anticipate it creating me a goods nomenclature with a heading? But this doesn't look like it will?

jebw (approved):
Looks great overall but I'm a bit wary of the `with_heading` trait which doesn't create a heading? I wasn't following why it was needed?

spec/factories/goods_nomenclature_factory.rb
willfish: Given only specific types of goods nomenclature have headings I think this okay. Updated the comment to reflect this

PR #1220, opened 2023-05-04
HOTT-2838-2839: Uploaded New Zealand and Australia documents to backend

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2838
https://transformuk.atlassian.net/browse/HOTT-2839

### What?

I have added/removed/altered:

- [ ] Uploaded New Zealand and Australia documents to backend

### Why?

I am doing this because:

- We need to display these documents in the New Zealand and Australia rules of origin journeys


jebw (approved):
LGTM

PR #1223, opened 2023-05-05
HOTT-3025 Fix chapters api for chapter 99

jebw:
### Jira link

HOTT-3025

### What?

I have added/removed/altered:

- [x] Fix chapters api for chapter 99

### Why?

I am doing this because:

- The chapter `9905000000-10`'s children were being excluded. This was because chapter 99 has headings at both indent 0 and indent 1

- Low, fixes minor bug


app/controllers/api/v2/chapters_controller.rb
```diff
@@ -74,16 +74,15 @@ def headings
...
         Chapter
-          .by_code(chapter_id)
+          .actual
           .non_hidden
-          .eager(ns_children: %i[ns_children goods_nomenclature_descriptions])
-          .limit(1)
-          .all
-          .first || (raise Sequel::RecordNotFound)
+          .by_code(chapter_id)
+          .take
```
willfish, line 80: Won't this kill eagerloading?

willfish, line 80: Should we get the take PR merged?

willfish: I guess not since we're doing no eager loading here

app/presenters/api/v2/chapters/chapter_presenter.rb
```diff
@@ -2,12 +2,17 @@ module Api
   module V2
     module Chapters
       class ChapterPresenter < SimpleDelegator
+        def initialize(chapter, flat_headings)
+          @flat_headings = flat_headings
```
willfish, line 6: What are flat headings?

PR #1228, opened 2023-05-09
HOTT-3140: Plumb together generation of basic reports

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3140

### What?

I have added/removed/altered:

- [x] Added basic reporting class and specs

### Why?

I am doing this because:

- This is a requirement from Kingfisher


jebw (approved):
This looks great.

One thing I am aware of is this pulls _a lot_ of data in a single go, ie every gn with PLS=80 and every measure for every commodity and every ancestor for every commodity and every child for every commodity.

An alternative would be to iterate chapter by chapter - you can then pull data like the Precache worker does - ie, you fetch a chapter, eager loading its ancestors and its descendants then - the descendants get their ancestors automatically and the in memory objects are reused from their parents. Thats only a suggestion though - this all runs out of  a worker so mem usage is less of a concern.

PR #1229, opened 2023-05-09
HOTT-3150 use nested set for gn class

jebw:
### Jira link

HOTT-3150

### What?

I have added/removed/altered:

- [x] Use nested set for `goods_nomenclature_class`
- [x] Improved `cast` functionality
- [x] Removed some duplication in serializer logic
- [x] Dropped `ns_goods_nomenclature_class` method

### Why?

I am doing this because:

- The new implementation of `goods_nomenclature_class` is faster then the existing implementation
- It is eager loadable
- It can benefit from `with_leaf_column` if loading goods nomenclatures in a non-hierarchical manner
- It is superceded by goods_nomenclature_class utilising nested set

- Changes heavily used method


app/serializers/api/beta/goods_nomenclature_serializer.rb
```diff
@@ -3,6 +3,18 @@ module Beta
     class GoodsNomenclatureSerializer
       include JSONAPI::Serializer
 
+      class << self
+        def serializer_proc
```
willfish, line 7:
```suggestion
        def ancestor_serializer_proc
```

willfish (approved):
\0/

PR #1236, opened 2023-05-15
HOTT-3139: Generate declarable duties report

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3139

### What?

I have added/removed/altered:

- [x] Fixes eager loading of quota definitions
- [x] Adds Excel/xlsx gem
- [x] Update factories
- [x] Generate a declarable duties report

### Why?

I am doing this because:

- This is required to prevent these from being generated in python for clients and from Matt's machine


spec/factories/measure_factory.rb
```diff
@@ -38,7 +38,7 @@
     geographical_area_id { generate(:geographical_area_id) }
     validity_start_date { default_start_date }
     validity_end_date   { nil }
-    reduction_indicator { [nil, 1, 2, 3].sample }
```
jebw, line 41: :pray:

app/lib/reporting/declarable_duties.rb
```diff
@@ -0,0 +1,227 @@
...
+          else
+            measure.public_send(header)
+          end
+        end
+      end
+
+      def each_declarable_and_measure
+        index = 1
+        TimeMachine.now do
+          goods_nomenclatures.each do |goods_nomenclature|
```
jebw, line 163: `each_goods_nomenclature`

app/lib/reporting/declarable_duties.rb
```diff
@@ -0,0 +1,227 @@
...
+
+              yield goods_nomenclature, measure
+
+              index += 1
+            end
+          end
+        end
+      end
+
+      def goods_nomenclatures
```
jebw, line 186:
```ruby
def each_goods_nomenclature
  Chapter.actual.non_hidden.all do |chapter|
    eager_chapter = Chapter.actual
                           .where(goods_nomenclature_sid: chapter.goods_nomenclature_sid)
                           .declarable
                           .eager(GOODS_NOMENCLATURE_EAGER)
                           .non_hidden
                           .non_classifieds
                           .take
      
    yield eager_chapter
    eager_chapter.ns_descendants.each do |goods_nomenclature|
      yield goods_nomenclature
    end
  end
end
```

app/lib/reporting/declarable_duties.rb
```diff
@@ -0,0 +1,227 @@
...
+      { geographical_area: :geographical_area_descriptions },
+      { measure_excluded_geographical_areas: [{ geographical_area: :geographical_area_descriptions }] },
+      :footnotes,
+      :measure_conditions,
+    ].freeze
+
+    GOODS_NOMENCLATURE_EAGER = [
+      { ns_ancestors: [{ ns_measures: MEASURE_EAGER }] },
+      { ns_measures: MEASURE_EAGER },
+      :ns_children,
```
jebw, line 26: { ns_descendants: MEASURE_EAGER },

jebw (approved):
Looks good _but_ this is effective pulling our entire data set for gns and measures in a single load - just gn's alone are pulled 3 times - once for every gn, once as part of ns_children and once as part of ns_ancestors

I've added a suggested approach to fetching goods nomenclatures chapter by chapter which I would think is more efficient but its untested so :man_shrugging:

app/lib/reporting/declarable_duties/presented_measure.rb
```diff
@@ -56,7 +56,7 @@ def measure_reduction_indicator
       end
 
       def measure__footnotes
-        footnotes.map(&:code).join('|')
+        footnotes.map(&:code).join('|').presence || nil
```
jebw, line 59: `|| nil` is redundant ?

app/lib/reporting/declarable_duties.rb
```diff
@@ -0,0 +1,241 @@
...
+      { geographical_area: :geographical_area_descriptions },
+      { measure_excluded_geographical_areas: [{ geographical_area: :geographical_area_descriptions }] },
+      :footnotes,
+      :measure_conditions,
+    ].freeze
+
+    GOODS_NOMENCLATURE_EAGER = [
+      {
+        ns_ancestors: [{ ns_measures: MEASURE_EAGER }],
+        ns_descendants: [{ ns_measures: MEASURE_EAGER }],
```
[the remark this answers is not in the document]
jebw: Should take out one query for every descendant getting serialized, ie ~16k

app/lib/reporting/declarable_duties.rb
```diff
@@ -0,0 +1,241 @@
...
+        ns_measures: MEASURE_EAGER,
+      },
+      :goods_nomenclature_descriptions,
```
jebw, line 29: Guessing this one isn't needed if the chapter records themselves aren't getting serialized (since they're not declarable)

jebw (dismissed):
LGTM

I think there's an N+1 on goods nomenclature descriptions for the rows getting serialized

PR #1237, opened 2023-05-15
Fix elasticsearch connection issues

jebw:
### Jira link

HOTT-????

### What?

I have added/removed/altered:

- [x] Retry connecting to ElasticSearch if its not yet up

### Why?

I am doing this because:

- Personal frustation - ElasticSearch is slow to open its socket after being started - this change allows rspec to retry the connection for 30 seconds rather then me swapping to another screen, coming back 5 minutes later only to find yet again ES wasn't ready and RSpec crashed out right at the start :wink: 

- Safe, only changes how test suite starts

### DIFF
--- a/spec/support/elasticsearch.rb
+++ b/spec/support/elasticsearch.rb
@@ -1,6 +1,20 @@
 RSpec.configure do |config|
   config.before(:suite) do
+    connection_attempts ||= 0
+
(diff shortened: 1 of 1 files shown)


jebw:
:laughing: ended up running into it one to many times - finally motivated me

PR #1264, opened 2023-05-31
HOTT-2638 add NestedSet#ns_declarable scope

jebw:
### Jira link

HOTT-2638

### What?

I have added/removed/altered:

- [x] Added a NestedSet#ns_declarable scope to filter goods nomenclature queries by only declarable GNs
- [x] Changed the `#by_code` methods on Chapter and Heading to use `=` instead of `LIKE`
- [x] Changed the `#by_code` methods on Subheading and Commodity to include the table in the where clause
- [x] Change the existing codebase to use `#ns_declarable` scope instead of `#declarable`

### Why?

I am doing this because:

- It is able to return only declarable GNs, the old method only filtered on PLS - so subsequent filtering in Ruby was required to find if a GN was really declarable - this often involved other expensive db queries and was not eager loadable
- Drive by fix but this should potentially speed up `#by_code` for Chapter and Heading - it certainly shouldn't be slower then using `LIKE`
- This is necessary if we are joining on tree nodes because they also have a `goods_nomenclature_item_id` column
- Replace use of the old scope with the new one because it is faster

- Replaces existing code with nested set variants with differing behaviour

### DIFF
--- a/app/controllers/api/v1/commodities_controller.rb
+++ b/app/controllers/api/v1/commodities_controller.rb
@@ -31,7 +31,7 @@ module Api
             :additional_code,
             :full_temporary_stop_regulations,
             :measure_partial_temporary_stops,
-          ).all
+          ).all,
         ).filter
 
         @commodity_cache_key = "commodity-#{@commodity.goods_nomenclature_sid}-#{actual_date}-#{TradeTariffBackend.currency}"
@@ -53,12 +53,10 @@ module Api
 
       def find_commodity
         @commodity = Commodity.actual
-                              .declarable
+                              .non_hidden
+                              .ns_declarable
                               .by_code(params[:id])
                               .take
-
-        raise Sequel::RecordNotFound if @commodity.children.any?
-        raise Sequel::RecordNotFound if @commodity.goods_nomenclature_item_id.in? HiddenGoodsNomenclature.codes
       end
     end
   end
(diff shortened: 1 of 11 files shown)


app/controllers/api/v1/commodities_controller.rb
```diff
@@ -53,12 +53,10 @@ def changes
 
       def find_commodity
         @commodity = Commodity.actual
-                              .declarable
+                              .non_hidden
+                              .ns_declarable
```
willfish, line 57: This is awesome!

app/elastic_search_indexes/search/goods_nomenclature_index.rb
```diff
@@ -4,7 +4,7 @@ def dataset
       TimeMachine.now do
         Commodity # Filter out headings and chapters
           .actual
-          .declarable # Avoid the majority of subheadings
+          .ns_declarable # Avoid all subheadings
```
willfish, line 7: <3

app/models/concerns/ten_digit_goods_nomenclature.rb
```diff
@@ -30,7 +30,7 @@ module TenDigitGoodsNomenclature
 
     dataset_module do
       def by_code(code = '')
-        filter(goods_nomenclature_item_id: code.to_s.first(10))
+        filter(goods_nomenclatures__goods_nomenclature_item_id: code.to_s.first(10))
```
willfish, line 33: I guess you're fully qualifying to enable this to be used with association/other kinds of join?

willfish (approved):

app/models/concerns/ten_digit_goods_nomenclature.rb
jebw: We're now joining on tree nodes and theres a `goods_nomenclature_item_id` column on that table as wel so the old code results in ambiguity (and PG raises an exception)

PR #1306, opened 2023-06-12
HOTT-3372: Refactor beta search indexer

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3372

### What?

I have added/removed/altered:

- [x] Move to indexing subheadings
- [x] Update eager load definition to make sure we're reducing n+1s
- [x] Update serializer implementation and surrounding services to make sure we can eager load
- [x] Refactor serializer implementation to remove unused/duplicate method

### Why?

I am doing this because:

- This speeds up the implementation significantly (c 280 seconds to c 40 seconds) and reduces the impact of a reindex on the postgres service


jebw (approved):
LGTM

PR #1317, opened 2023-06-14
HOTT-3372: Bulk search api first cut

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3365

### What?

This is a bulk-search api proof of concept.

The api receives a jsonapi-compatible structure in a POST request to /bulk_search. This generates a unique id, sets the searches against that id in redis and proceeds to queue a job to the bulk search worker.

The api user then uses the uuid returned by the api to subsequently poll the bulk search api with GET /bulk_search/:id.

The status of the search transitions from :queued, to :processing and finally to one of :completed or :error.

When the search is completed the user will receive searches with the scores for 6 digit subheadings accumulated. There are typically many of these but where there are none or if the commodity hit is a 6 digit itself, we'll return the commodity or the corresponding heading.

The api relies heavily on the goods nomenclature index created as part of beta search and does a full document search against this for now. In future, we might want to add:

- spelling correction
- parts of speech deconstruction
- client-specific handling of input terms
- trader terms fallbacks
- prioritised trader terms
- csv outputs of results for comparison with training data
- an ADR/api documentation

I have added/removed/altered:

- [x] Added bulk search enqueue api
- [x] Added bulk search long poll api
- [x] Added async functionality to process the result collection and attach the search results

### Why?

I am doing this because:

- There are no corresponding tests for this yet because this is just a POC


app/services/ancestor_search_result_service.rb
```diff
@@ -0,0 +1,82 @@
...
+      reason = :no_matching_digit_ancestor
+      # Try the 6 digit ancestor
+      candidate_ancestor = hit._source.ancestors.find do |ancestor|
+        ancestor.short_code.length == ancestor_digits
+      end
+
+      reason = :matching_digit_ancestor if candidate_ancestor.present?
+
+      # Otherwise check if the hit is a 6 digit code or a declarable heading
+      candidate_ancestor ||= begin
```
willfish, line 19: We could hoist this first and avoid repeat ancestor loops

config/application.rb
```diff
@@ -42,6 +42,7 @@ class Application < Rails::Application
     # Tells Rails to serve error pages from the app itself, rather than using static error pages in public/
     config.exceptions_app = routes
 
+    config.middleware.use Rack::Deflater
```
willfish, line 45: This has no impact on existing requests. It just provides additional functionality for those requests that include compression headers (specifically only gzip)

app/models/concerns/ten_digit_goods_nomenclature.rb
```diff
@@ -113,8 +113,8 @@ def code
       goods_nomenclature_item_id
     end
 
-    def short_code
-      if path_declarable?
+    def short_code(fetch_commodity_short_code: false)
```
jebw, line 116: This feels like its repurposing an existing method when you actually need to be able to call just part of the existing method? How about breaking it up instead? Then avoids the `NullObject` overrides as well, eg

```
class GoodsNomenclature
    def some_other_short_code
      short_code
    end
  end
end

module TenDigitGoodsNomenclature
  def short_code
    ns_declarable? ? goods_nomenclature_item_id : some_other_short_code
  end

  def some_other_short_code
    case goods_nomenclature_item_id
        when /^\d+0000$/ then harmonised_system_code
        when /^\d+00$/ then combined_nomenclature_code
        else taric_code
        end
  end
end
```

test.sh
```diff
@@ -0,0 +1,26 @@
...
+job_id=$(curl --compressed -s -X POST "http://localhost:3000/bulk_search" -d @body.json -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Accept-Encoding: gzip' | jq -r '.data.id')
+
+echo "Job ID: $job_id"
+# Continually poll the status of the job until it's completed
+while true; do
+    # GET request to check the status of the job
+    response=$(curl --compressed -s "http://localhost:3000/bulk_search/${job_id}" -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Accept-Encoding: gzip')
+
+    # Extract the status from the response
+    status=$(echo "$response" | jq -r '.data.attributes.status')
```
jebw, line 13: I'd anticipate clients to drive their flow control from the HTTP status codes - more REST than SOAP?

jebw (commented):
A few code comments - I appreciate this is still very much PoC stage - but looks great overall

I think if we've got clients polling every half second we may have concerns (I'd expect more like once a minute or every 30 seconds)

The part I'm fuzziest on is the AncestorService - started looking at it but was struggling to follow the code so I'll take another look once its had a bit of a refactor

app/controllers/api/v2/bulk_searches_controller.rb
```diff
@@ -0,0 +1,27 @@
+module Api
+  module V2
+    class BulkSearchesController < ApiController
+      def create
+        @result = ::BulkSearch.enqueue(bulk_search_params)
+        @serialized_result = Api::V2::BulkSearch::ResultCollectionSerializer.new(@result, include: [:searches]).serializable_hash
+
+        render json: @serialized_result, status: @result.http_code
```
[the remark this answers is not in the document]
willfish: I could just set this to 202, yeah

app/workers/bulk_search_worker.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  sidekiq_options queue: :bulk_search, retry: false
+
+  delegate :redis, :v2_search_client, to: TradeTariffBackend
+
+  def perform(id)
+    Rails.logger.info("BulkSearchWorker: #{id} - #{BulkSearch.find(id)}")
+    BulkSearchService.new(id).call
+    Rails.logger.info("BulkSearchWorker: #{id} - #{BulkSearch.find(id)}")
+  rescue StandardError => e
```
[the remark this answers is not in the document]
willfish: I'm actually going to be handling errors separately in a follow on PR and give it some thought

config/application.rb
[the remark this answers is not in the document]
willfish: They don't add these headers no. But we'll put this through its paces on the e2e suite before this goes anywhere near production

test.sh
willfish: Yeah, I'm not actually going to commit this file :)

app/serializers/api/v2/bulk_search/search_ancestor_serializer.rb
```diff
@@ -0,0 +1,21 @@
...
+        include JSONAPI::Serializer
+
+        set_type :search_result_ancestor
+
+        attributes :short_code,
+                   :goods_nomenclature_item_id,
+                   :description,
+                   :producline_suffix,
+                   :goods_nomenclature_class,
+                   :declarable,
```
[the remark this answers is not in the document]
willfish: This just isn't a thing. The job that produces these values runs asynchronously and these are just being read out of opensearch

app/controllers/api/v2/bulk_searches_controller.rb
```diff
@@ -5,7 +5,7 @@ def create
         @result = ::BulkSearch::ResultCollection.enqueue(bulk_search_params)
         @serialized_result = Api::V2::BulkSearch::ResultCollectionSerializer.new(@result).serializable_hash
 
-        render json: @serialized_result, status: @result.http_code
+        render json: @serialized_result, status: :accepted
```
jebw, line 8: Can we include a location with this - the logic being a lot of http clients will be able to follow a location header out of the box (eg Faraday can). This is standard practice for REST

Without it clients will require custom parsing logic to extract the url from the JSON body where they should be polling for the result. This isn't a lot of code to write but if we can follow standard practice, I think we should?

app/serializers/search/goods_nomenclature_serializer.rb
```diff
@@ -7,7 +7,7 @@ def serializable_hash(_opts = {})
         serializable = {
           id:,
           goods_nomenclature_item_id:,
-          short_code: short_code(fetch_commodity_short_code: true),
+          short_code: respond_to?(:specific_system_short_code) ? specific_system_short_code : short_code,
```
jebw, line 10: This teaches the generic serializer about varients of GoodsNomenclature

Whats the scenario where system specific short code isn't available? Can you just add an implementation of this to GoodsNomenclature, and have that call `#short_code`. Then it would be overridden by TenDigit ?

Then all goods_nomenclatures behave the same and the serializer does need logic to determine how to serializer the GN?

jebw (commented):
Looks great - only point I'd really pick up on is not telling the client in the HTTP headers what the GET url is - the big win with REST is avoiding duplicating the bits HTTP has already standardised

Other nit is introspecting the goods nomenclature from the serializer - feels like goods nomenclature should be presenting a consisten interface to the serializer rather then putting the logic into its serializer?

app/serializers/search/goods_nomenclature_serializer.rb
willfish: That's a good shout, actually.

app/models/null_chapter.rb
```diff
@@ -1,3 +1,3 @@
 class NullChapter < NullObject
-  def short_code(_ = {}); end
+  def short_code; end
```
jebw: Not needed?

https://github.com/trade-tariff/trade-tariff-backend/blob/main/app/lib/null_object.rb#L10-L12

config/initializers/sentry.rb
```diff
@@ -5,6 +5,7 @@
     Sequel::Plugins::RailsExtensions::ModelNotFound
     Sequel::NoMatchingRow
     Sequel::RecordNotFound
+    BulkSearch::RecordNotFound
```
jebw, line 8: This should be `BulkSearch::ResultCollection::RecordNotFound` I think

jebw (commented):
Looks good - specs are failing though and looks like the wrong exception is listed in the sentry config

PR #1322, opened 2023-06-16
HOTT-2642 Drop custom STI logic

jebw:
### Jira link

HOTT-2642

### What?

I have added/removed/altered:

- [x] Moved the generic casting logic up to GoodsNomenclature
- [x] Renamed it the `#cast_according_to_declarable` to `#sti_cast`
- [x] Dropped the `custom_sti_goods_nomenclature` code from `SearchReference`
- [x] Dropped the `custom_sti_goods_nomenclature` code from `SearchSuggestion`

### Why?

I am doing this because:

- DRYing out code is good

- Low


willfish (approved):
LGTM. I've written a whole bunch of e2e coverage for legacy/beta search so hopefully we'd pick up any issues instantly (not assuming there will be any)

PR #1323, opened 2023-06-16
HOTT-2642 Drop materialized path

jebw:
### Jira link

HOTT-2642

### What?

I have added/removed/altered:

- [x] Removes materialized path since its no longer in use

### Why?

I am doing this because:

- We have it in version control if required in the future but for now it is redundant, removal simplifies our code base

- Low


willfish (approved):
Yay \0/

willfish (approved):

PR #1335, opened 2023-06-20
HOTT-3429 Use nested set for commodities

jebw:
### Jira link

HOTT-3429

### What?

I have added/removed/altered:

- [x] Use nested set for commodities API

### Why?

I am doing this because:

- the nested set mechanism is faster

### Have you? (optional)

- Some, changes mechanisms used for loading measures

### DIFF
--- a/app/presenters/api/v2/commodities/commodity_presenter.rb
+++ b/app/presenters/api/v2/commodities/commodity_presenter.rb
@@ -48,8 +48,12 @@ module Api
           commodity.chapter.goods_nomenclature_sid
         end
 
+        def ancestors
+          @ancestors ||= ns_ancestors.select { |ancestor| ancestor.is_a?(Commodity) }
+        end
+
         def ancestor_ids
-          commodity.ancestors.pluck(:goods_nomenclature_sid)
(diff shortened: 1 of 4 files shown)


app/presenters/api/v2/commodities/commodity_presenter.rb
```diff
@@ -48,8 +48,12 @@ def chapter_id
           commodity.chapter.goods_nomenclature_sid
         end
 
+        def ancestors
+          @ancestors ||= ns_ancestors.select { |ancestor| ancestor.is_a?(Commodity) }
```
willfish, line 52: This is a little confusing. So these are basically all subheadings?

willfish (approved):

willfish (approved):

app/presenters/api/v2/commodities/commodity_presenter.rb
jebw: Yeh, agreed - though at present STI will return them as Commodities 

I was planning to hoist this up to goods nomenclature in the next week or so and I'll make it check for either at that point (more in the scope of the STI ticket though?)

PR #1336, opened 2023-06-20
HOTT-3453 - Update the UK RoO schemes definition

jebw:
### Jira link

HOTT-3452
HOTT-3453

### What?

I have added/removed/altered:

- [x] Disambiguate the two DCTS schemes
- [x] Removed Vietnam from the dcts-ldcs scheme

### Why?

I am doing this because:

- On the proofs screen on the frontend which lists proofs for all schemes, you cannot tell which scheme is which
- Vietnam was included in dcts-ldcs by mistake

- Low


willfish (approved):
Definitely better way of displaying/avoiding confusion

PR #1342, opened 2023-06-22
HOTT-3381-3384: Load exchange rates data

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3381
https://transformuk.atlassian.net/browse/HOTT-3382
https://transformuk.atlassian.net/browse/HOTT-3383
https://transformuk.atlassian.net/browse/HOTT-3384
https://transformuk.atlassian.net/browse/HOTT-3483
https://transformuk.atlassian.net/browse/HOTT-3482

### What?

I have added/removed/altered:

- [ ] Added exchange rates data loaders

### Why?

I am doing this because:

- we need to display this data on the exchange rates frontend

### DIFF
--- a/app/models/exchange_rate_currency_rate.rb
+++ b/app/models/exchange_rate_currency_rate.rb
@@ -4,21 +4,6 @@ class ExchangeRateCurrencyRate < Sequel::Model
   RATES_FILE = 'data/exchange_rates/all_rates.csv'.freeze
   SPOT_RATES_FILE = 'data/exchange_rates/all_spot_rates.csv'.freeze
 
-  def before_save
-    self.rate_type = determine_rate_type if validity_start_date
-    super
-  end
-
-  private
-
-  def determine_rate_type
-    if scheduled_rate?
-      'scheduled'
-    elsif spot_rate?
-      'spot'
-    end
-  end
-
   def scheduled_rate?
     validity_end_date.present? && validity_start_date.day == 1 && validity_end_date == validity_start_date.end_of_month
   end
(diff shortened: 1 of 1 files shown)


app/models/exchange_rate_currency.rb
```diff
@@ -0,0 +1,21 @@
+require 'csv'
+
+class ExchangeRateCurrency < Sequel::Model
+  CURRENCY_FILE = 'data/exchange_rates/currency.csv'.freeze
+
+  class << self
+    def populate
```
jebw, line 7: `def populate(src = CURRENCY_FILE)`

app/models/exchange_rate_currency.rb
```diff
@@ -0,0 +1,21 @@
...
+      unrestrict_primary_key
+
+      CSV.foreach(CURRENCY_FILE, headers: true) do |row|
```
jebw, line 10: `CSV.foreach(src, headers: true) do |row|`

spec/models/exchange_rate_country_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+RSpec.describe ExchangeRateCountry do
+  let(:csv_file) { 'spec/fixtures/exchange_rates/territory.csv' }
+  let(:dh) { described_class.where(country_code: 'DH').take }
+  let(:kn) { described_class.where(country_code: 'KN').take }
+  let(:lc) { described_class.where(country_code: 'LC').take }
+
+  before do
+    stub_const('ExchangeRateCountry::COUNTRY_FILE', Rails.root.join(csv_file))
+    described_class.populate
```
jebw, line 11:
```ruby
described_class.populate Rails.root.join(csv_file)
```

jebw (changes requested):
Looks really good - nice a clean

One change I would suggest is changing the `.populate` methord to take the csv file as a parameter. It can still default to the one in the constant but that will avoid the need to `stub_const` in the specs which is poking into the class internals and isn't really needed in this case

A second observation is generally I try to group the migrations into one larger one - the rationale being that I want it to succeed or fail as a single unit and each migration runs in its own transaction. That isn't enough of a concern to be worth changing this though and its table additions if its half complete then we end up with orphan tables but that wont impact running code.

jebw (dismissed):
LGTM

willfish (approved):

PR #1343, opened 2023-06-26
HOTT-3472: Added exchange rates rake tasks

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3472

### What?

I have added/removed/altered:

- [ ] [Added exchange rates rake tasks

### Why?

I am doing this because:

- This will make it easier to run the exchange rates tasks

### DIFF
--- a/lib/tasks/exchange_rates.rake
+++ b/lib/tasks/exchange_rates.rake
@@ -1,0 +1,25 @@
+namespace :exchange_rates do
+  desc 'Load CSV data into ExchangeRateCurrencyRate table'
+  task recreate_currency_rates: %w[environment] do
+    ExchangeRateCurrencyRate.db.transaction do
+      ExchangeRateCurrencyRate.dataset.delete
+      ExchangeRateCurrencyRate.populate
+    end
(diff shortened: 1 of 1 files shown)


jebw (approved):
LGTM but would be nice to add a task to cover all the others - this can be done with the dependencies feature of rake, eg 

https://github.com/trade-tariff/trade-tariff-backend/blob/main/lib/tasks/tariff.rake#L41

lib/tasks/exchange_rates.rake
```diff
@@ -0,0 +1,25 @@
+namespace :exchange_rates do
+  desc 'Load CSV data into ExchangeRateCurrencyRate table'
+  task recreate_currency_rates: %w[environment] do
+    ExchangeRateCurrencyRate.db.transaction do
+      ExchangeRateCurrencyRate.dataset.delete
```
willfish, line 5:
```suggestion
      ExchangeRateCurrencyRate.truncate
```

willfish (dismissed):

PR #1344, opened 2023-06-26
HOTT-3507 Let indexer control reindexing environment

jebw:
### Jira link

HOTT-3507

### What?

I have added/removed/altered:

- [x] Ensure index applies relevant contraints when reindexing

### Why?

I am doing this because:

- Previously if data was excluded from an eager load (eg because of bad references, or because of a missing eager load) then the subsequent attempt to fetch the data at serialization time would either do a non-timemachine query or in the case of nested sets would actually raise an error because it was outside of time machine

- Changes reindexing but should improve consistency so lower risk of different behaviour under differing circumstances

### DIFF
--- a/app/elastic_search_indexes/concerns/point_in_time_index.rb
+++ b/app/elastic_search_indexes/concerns/point_in_time_index.rb
@@ -3,10 +3,14 @@
 
 module PointInTimeIndex
   def dataset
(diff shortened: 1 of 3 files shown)


willfish (approved):
LGTM

PR #1345, opened 2023-06-27
HOTT-3368 Set cache headers

jebw:
### Jira link

HOTT-3368

### What?

I have added/removed/altered:

- [x] Set default cache headers across entire API - derived from latest REVISION file and last changed to TariffUpdates
- [x] Enforce no caching on Admin endpoints
- [x] Enforce no caching on various other endpoints
- [x] Add SimpleCaching module for just caching for a fixed length of time without using ETags
- [x] Use the `updated_at` date for the news items to determine whether they are current
- [x] Add db indexes where necessary

### Why?

I am doing this because:

- These headers can be used by both our frontends and our CDN to reduce load on the backend and speed up the site

- Limited, the `Cache-Control` headers are currently hidden by the api proxying in the frontend

### Notes

- There's is a corresponding frontend PR which will start using this content to enable caching on the frontend
- There will be a future frontend PR to expose these headers in the caching proxy, and disable caching in the rest of the frontend
- There will be a future PR to turn on caching in the CDN

### Screenshots

Screenshot from 2023-06-27 16-24-48
Screenshot from 2023-06-27 16-19-34

### DIFF
--- a/app/lib/tariff_synchronizer/base_update.rb
+++ b/app/lib/tariff_synchronizer/base_update.rb
@@ -155,6 +155,16 @@ module TariffSynchronizer
       raise NotImplementedError
     end
 
+    def cache_key_with_version
+      [
+        'tariff-update',
+        filename,
+        update_type,
+        state,
+        applied_at.iso8601,
+      ].map(&:to_s).join('-')
+    end
+
     class << self
       delegate :instrument, to: ActiveSupport::Notifications
 
--- a/app/lib/trade_tariff_backend.rb
+++ b/app/lib/trade_tariff_backend.rb
@@ -1,5 +1,6 @@
 module TradeTariffBackend
   MAX_LOCK_LIFETIME = 600_000
+  REVISION_FILE = Rails.root.join('REVISION').to_s.freeze
 
   class << self
     SERVICE_CURRENCIES = {
@@ -245,6 +246,14 @@ module TradeTariffBackend
 
     def opensearch_debug
       ENV.fetch('OPENSEARCH_DEBUG', 'false') == 'true'
+    end
+
+    def revision
+      @revision ||= begin
+        File.read(REVISION_FILE).chomp if File.file?(REVISION_FILE)
+      rescue Errno::EACCES
+        nil
+      end
     end
   end
 end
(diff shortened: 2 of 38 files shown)


app/lib/tariff_synchronizer/base_update.rb
```diff
@@ -155,6 +155,16 @@ def import!
       raise NotImplementedError
     end
 
+    def cache_key_with_version
+      [
+        'tariff-update',
+        filename,
```
willfish, line 161: Probably the filename is enough, here but no harm including the other fields

app/lib/trade_tariff_backend.rb
```diff
@@ -246,5 +247,13 @@ def opensearch_host
     def opensearch_debug
       ENV.fetch('OPENSEARCH_DEBUG', 'false') == 'true'
     end
+
+    def revision
+      @revision ||= begin
+        File.read(REVISION_FILE).chomp if File.file?(REVISION_FILE)
+      rescue Errno::EACCES
```
willfish, line 254: Is this a no permission error from the operating system? Do we want to also check if the error is missing file, etc? `ENOENT`

willfish: Or just rescue the superclass of these?

```ruby
 Errno::EACCES.superclass
=> SystemCallError
```

willfish (dismissed):
Looks great to me \0/.

app/lib/trade_tariff_backend.rb
jebw: This is the same code as we were previously using in the healthcheck endpoint and it seemed to work fine there, there are specs for verifying it works with both a missing file or an unreadable (permissions) file?

I'm inclined to go with what we've already been using rather then introducing change to this part.

app/lib/tariff_synchronizer/base_update.rb
jebw: Yeh, it was like that originally but then changed it to follow what ActiveRecord was doing man_shrugging

docs/adr/2023-06-28_add_cache_headers.md
```diff
@@ -0,0 +1,43 @@
+# Adding cache headers
+
+Date: 28 June 2023
+
+## Context
+
+We currently have a CDN which has its cache support disable and always just passes requests straight through to first the frontend end and then onward to the backend.
```
willfish, line 7:
```suggestion
We currently have a CDN which has its cache support disabled and always just passes requests straight through to first the frontend end and then onward to the backend.
```

willfish (approved):

PR #1346, opened 2023-06-28
HOTT-3447: Adds support for csv to bulk search api

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added CSV support to bulk search api

### Why?

I am doing this because:

- This is required for simple reporting from the results of this api


jebw (approved):
LGTM

PR #1347, opened 2023-06-28
HOTT-3447: Refactor bulk search

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Extract picking the relevant ancestor

### Why?

I am doing this because:

- To support other consumers working out their 6 digit ancestor separately


jebw (dismissed):
LGTM

PR #1348, opened 2023-06-28
HOTT-3447: Bulk search reporting

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Adds reporting implementation for bulk search

### Why?

I am doing this because:

- This is required to handover to HMRC a worksheet for review.


jebw (dismissed):
LGTM - only minor concern is lots of iteration in long methods breaking it up might make it easier for future developers revisiting the code?

willfish:
So my plan is to actually delete all of this soon but just give me something I can reuse over the next month or so

PR #1349, opened 2023-06-28
HOTT-3474: Create periods API

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3474

### What?

I have added/removed/altered:

- [x] Added api to surface exchange rate period list information
- [x] Added serializers, controllers, models and routes

### Why?

I am doing this because:

- This is going to be used by a UI in the frontend and reduces the number
of api requests the client needs to make

### DIFF
--- a/app/models/exchange_rates/period.rb
+++ b/app/models/exchange_rates/period.rb
@@ -7,18 +7,17 @@ module ExchangeRates
     end
 
     class << self
-      def build(months, year)
+      def wrap(months, year)
         months.map do |month|
-          period = new
-          period.month = month
-          period.year = year
-          period.files = files_for(month, year)
-          period
+          build(month, year)
         end
       end
 
-      def files_for(_month, _year)
-        []
+      def build(month, year)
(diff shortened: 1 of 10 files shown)


app/models/exchange_rate_currency_rate.rb
```diff
@@ -12,6 +12,27 @@ def spot_rate?
...
+      where(rate_type: 'spot')
+    end
+
+    def by_year(year)
+      return if year.blank?
+
+      start_of_year = Time.zone.parse("#{year}-01-01").beginning_of_year
+      end_of_year = start_of_year.end_of_year
+
+      where { (validity_start_date >= start_of_year) & (validity_start_date <= end_of_year) }
```
shjohnson, line 30: Realise this is still a work in progress! Just a thought but should this be:

```
where { (validity_start_date >= start_of_year) & (validity_end_date <= end_of_year) }
```

willfish:
:sheep: :it:

PR #1351, opened 2023-06-29
BAU: Debug rails constant

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Qualify rails constant in backend module

### Why?

I am doing this because:

- Try to fix worker deployment issues


jebw (dismissed):
LGTM

jebw (changes requested):
Still not right - logic of that is

if ENV['SERVICE'] is set; then use 'uk'; else use ENV['SERVICE']

jebw (approved):

PR #1366, opened 2023-07-04
HOTT-3571: Handle expired nomenclature in chemical suggestions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3571

### What?

I have added/removed/altered:

- [x] Added handling for chemical suggestions without current goods nomenclature

### Why?

I am doing this because:

- This is required to unlock the population of suggestions

### DIFF
--- a/app/services/suggestions_service.rb
+++ b/app/services/suggestions_service.rb
@@ -11,6 +11,7 @@ class SuggestionsService
       ].flatten.compact
     end
     SearchSuggestion.restrict_primary_key
+    Rails.logger.debug("Query count: #{::SequelRails::Railties::LogSubscriber.count}")
     suggestions
   end
 
@@ -27,9 +28,10 @@ class SuggestionsService
   def build_search_references_search_suggestions
     SearchReference
       .select(:id, :title, :goods_nomenclature_sid, :referenced_class)
-      .eager(:referenced)
+      .eager(referenced: :ns_children)
       .distinct(:title)
       .order(Sequel.desc(:title))
+      .all
       .map { |search_reference| build_search_reference_record(search_reference) }
   end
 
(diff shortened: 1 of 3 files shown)


app/services/suggestions_service.rb
```diff
@@ -11,6 +11,7 @@ def call
       ].flatten.compact
     end
     SearchSuggestion.restrict_primary_key
+    Rails.logger.debug("Query count: #{::SequelRails::Railties::LogSubscriber.count}")
```
jebw, line 14: FYI this doesn't really work - because nothing is resetting the counter?

In the context of a call from a controller the counter gets reset in a before_action, and written out to the log in an after action - but here it will

1) When called from a controller: Show the count of all queries from the start of the controller action up to the point this line is executed

2) When it is called outside of a controller, it will probably list all queries from app boot until this line is called

_and_ you don't want to be resetting it here because it will break the query counts on countroller actions?

app/services/suggestions_service.rb
```diff
@@ -1,5 +1,6 @@
 class SuggestionsService
   def call
+    ::SequelRails::Railties::LogSubscriber.reset_count
```
jebw, line 3: Since this is normally reset before the worker runs, it would be better to avoid this getting called here and the log line was potentially inaccurate - rather then resetting outside of a known lifecycle potentially leaving a point of confusion for other developers if this service ever ends up part of other jobs/actions in the future

jebw (changes requested):
Only query is around the resetting of the global counter of db queries

Whilst the reporting line may be inaccurate, the resetting line is actually a potential source of confusion for future developers?

jebw (approved):
The logic changes are great so approving

jebw (approved):
LGTM

PR #1367, opened 2023-07-04
HOTT-3571 Check for Commodity vs Subheading using appropriate date

jebw:
### Jira link

HOTT-3571

### What?

I have added/removed/altered:

- [x] Changed how we determine whether an exact result for goods nomenclature is a Subheading or Commodity - we now check at
    - The last date valid if it is historical
    - The first date valid if it is in the future
    - The as_of search date if it is current

### Why?

I am doing this because:

- Previously this would look for any children at all points in time, this meant if a GoodsNomenclature had ever been a Subheading the it would always be treated as a Subheading. This was true even if a the point of search it was now a Commodity

### Have you? (optional)

- Changes behaviour of the ExactSearch class

### DIFF
--- a/app/services/search_service/exact_search.rb
+++ b/app/services/search_service/exact_search.rb
@@ -58,15 +58,19 @@ class SearchService
         .limit(1)
         .first
 
-      # TimeMachine is implicitly used when fetching declarability with the GoodsNomenclatureMapper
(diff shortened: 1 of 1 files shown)


willfish (approved):
Makes loads more sense

PR #1368, opened 2023-07-04
HOTT-3572: Handle hierarchy movements in search suggestions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3571

### What?

I have added/removed/altered:

- [x] Handle hierarchy movements in search suggestions

### Why?

I am doing this because:

- This is required because the search suggestions use values which change as the hierarchy reorganises but these changes were not being reflected by clearing out the no-longer-current values.


jebw (dismissed):
LGTM

PR #1377, opened 2023-07-06
HOTT-3119 Use leaf where available for Single Table Inheritance

jebw:
### Jira link

HOTT-3119

### What?

I have added/removed/altered:

- [x] Reintroduce the Single Table Inheritance change
- [x] Fixed a bug discovered where a check for Commodity was introduced in a separate branch concurrently with this change

### Why?

I am doing this because:

- The functionality is desirable but was reverted in a hotfix after discover of the bug

- Medium - its been deployed once already, and we know it doesn't trigger any outright problems. The origin of how the bug crept in has been established - multiple concurrent branches, and a new check for instances of `is_a?(Commodity)` has been done. Specs have also been introduced checking the ancestors behaviour from the problematic presenter.

### DIFF
--- a/app/models/goods_nomenclature.rb
+++ b/app/models/goods_nomenclature.rb
@@ -19,7 +19,11 @@ class GoodsNomenclature < Sequel::Model
     elsif gono_id.ends_with?('000000') && gono_id.slice(2, 2) != '00'
       'Heading'
     elsif !gono_id.ends_with?('000000')
-      'Commodity'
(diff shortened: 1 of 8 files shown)


jebw:
Sorry didn't think to take one but it was the commodity tree was missing the subheadings - only occurred on the commodities page

Most of 'todays' pages had cached data so suspect it only really showed up for users looking at different dates

PR #1379, opened 2023-07-07
Hott 3399

shjohnson:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [ ] Added foo in bar
- [ ] Removed baz

### Why?

I am doing this because:

-
-
-

### Have you? (optional)


app/services/exchange_rates/upload_monthly_csv.rb
```diff
@@ -0,0 +1,47 @@
+module ExchangeRates
+  class UploadMonthlyCsv
+    attr_reader :current_time, :month, :year, :date_string
+
+    delegate :instrument, :subscribe, to: ActiveSupport::Notifications
```
willfish, line 5:
```suggestion
    delegate :instrument, to: ActiveSupport::Notifications
```

Gemfile
```diff
@@ -25,6 +25,7 @@ gem 'redis-client', '>= 0.11.2'
 gem 'redlock'
 gem 'sidekiq', '< 8'
 gem 'sidekiq-scheduler'
+gem 'whenever', require: false
```
willfish, line 28: I'm not convinced by us adding a new way of scheduling jobs currently. Why would we use this rather than sidekiq-scheduler?

app/services/exchange_rates/create_csv.rb
```diff
@@ -0,0 +1,71 @@
+module ExchangeRates
```
willfish, line 1: An idiom in the app is to always name services with a Service suffix

app/services/exchange_rates/upload_monthly_csv.rb
```diff
@@ -0,0 +1,47 @@
+module ExchangeRates
+  class UploadMonthlyCsv
+    attr_reader :current_time, :month, :year, :date_string
```
willfish, line 3: Should these be private?

app/services/exchange_rates/create_csv.rb
shjohnson: Yeah i mentioned this to Jeremy because seemed confusing to me? Is this enforced? Also noticed it with the `_factory` in factories? didnt seem much point as already prefixed under the factories folder.

Jeremy mentioned about `_service` in services that its to do with name clashes (potentially models) but then id see that as a code smell if something doesnt look like a service and naming is incorrect or approach? Happy to discuss this if my point doesnt make sense?

spec/services/exchange_rates/upload_monthly_csv_spec.rb
```diff
@@ -0,0 +1,103 @@
+# rubocop:disable RSpec/MultipleMemoizedHelpers
+RSpec.describe ExchangeRates::UploadMonthlyCsv do
+  subject(:upload_csv) { described_class.call }
+
+  let(:current_time) { Time.zone.now }
```
willfish, line 5: I wonder if we could just pass these to the service and then we can avoid all of the memoization

Gemfile
shjohnson: ahhh i didnt see this! yeah happy to use that instead 👍

app/services/exchange_rates/create_csv.rb
willfish: Its mainly for consistency at this point. Not really any good reason. I'd be happy to have them all renamed to not include Service. Just being a mixed bag is odd

willfish: Name clashes makes sense, yeah

app/services/exchange_rates/upload_monthly_csv.rb
shjohnson: Yeah can do just bad habit there 👍

app/services/exchange_rates/create_csv.rb
shjohnson: Yeah makes sense, i can change and if I really throw toys out pram and get round to it propose a folder change PR 😂

willfish: :tada:

spec/services/exchange_rates/upload_monthly_csv_spec.rb
shjohnson: To be honest i quite liked it being like an atomic action so its all held within the class. I did have at first passing in the date as this might help further down the line with other updates to the CSV file but for now we dont have all the full requirements.

shjohnson: Theres some extra memoization to allow the `.for_month` to only expect two integers which could just accept a date i guess? but think nicer to work like this.

Gemfile
shjohnson: Even though realise it was literally the line above 😂

spec/services/exchange_rates/upload_monthly_csv_spec.rb
willfish: Yeah. I think I just noticed the bypassing of the rubocop rules and figured something smells here with the lots of setup.

Gemfile
willfish: No worries :rocket: :smile:

app/services/exchange_rates/upload_monthly_csv.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def data_result
+      @data_result ||= ::ExchangeRateCurrencyRate.for_month(month, year)
+    end
+
+    def file_path
+      "data/exchange_rates/monthly_csv_#{date_string}.csv"
+    end
+
+    def penultimate_thursday?
```
willfish, line 41: I love this. I wouldn't have approached it naturally like this but it makes loads of sense.

app/services/exchange_rates/upload_monthly_csv.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def initialize
+      @current_time = Time.zone.now
+      @date_string = current_time.to_date.to_s
+      @month = current_time.month
+      @year = current_time.year
+    end
+
+    def call
+      # return unless penultimate_thursday?
```
willfish, line 17: Does this need to be uncommented?

spec/services/exchange_rates/upload_monthly_csv_spec.rb
shjohnson: Do you think this is a blocker? i know ive worked where very mixed opinions on this cop? All id be doing is moving around to another class sort of spreading dependency potentially?

shjohnson: Or could just not store some string values and add them inline which seems a but pointless 🤷

app/services/exchange_rates/upload_monthly_csv.rb
shjohnson: Yeah, i was thinking could be done something like this but hadnt given much thought then realise @jebw had already solved it basically this way... So kudos to his idea!

spec/services/exchange_rates/upload_monthly_csv_spec.rb
willfish: TBH, your approach is actually more readable for the things that are conceptually relevant to the test.

app/services/exchange_rates/upload_monthly_csv.rb
willfish: I'd probably have worked out every Thursday and then indexed the one we're targeting.

app/workers/monthly_exchange_rates_worker.rb
```diff
@@ -0,0 +1,9 @@
+class MonthlyExchangeRatesWorker
+  include Sidekiq::Worker
+
+  sidekiq_options retry: false
+
+  def perform
+    ExchangeRates::UploadMonthlyCsv.call
```
maverickvi, line 7: ExchangeRates::UploadMonthlyCsvService.call

app/models/exchange_rate_currency_rate.rb
```diff
@@ -4,6 +4,8 @@ class ExchangeRateCurrencyRate < Sequel::Model
   RATES_FILE = 'data/exchange_rates/all_rates.csv'.freeze
   SPOT_RATES_FILE = 'data/exchange_rates/all_spot_rates.csv'.freeze
 
+  many_to_one :currency, key: :currency_code, primary_key: :currency_code, class_name: ExchangeRateCurrency
```
willfish, line 7:
```suggestion
  many_to_one :exchange_rate_currency, key: :currency_code, primary_key: :currency_code
```

app/models/exchange_rate_currency_rate.rb
```diff
@@ -12,6 +14,10 @@ def spot_rate?
     [3, 12].include?(validity_start_date.month) && validity_start_date.day == 31 && validity_end_date.nil?
   end
 
+  def countries
```
willfish, line 17: Should this be a has many?

app/models/exchange_rate_currency_rate.rb
```diff
@@ -76,6 +92,14 @@ def months_for_year(year)
         .uniq
     end
 
+    def for_month(month, year = Time.zone.today.year)
```
willfish, line 95: You could make this extendable by moving it to the dataset module section and doing the `.all` call elsewhere. No biggie, though, we might never need variations on this scope I guess

app/services/exchange_rates/create_csv_service.rb
```diff
@@ -0,0 +1,92 @@
...
+        ordered_data = order_data
+
+        ordered_data.each do |currency_country_data|
+          csv << build_row(currency_country_data)
+        end
+      end
+    end
+
+    def build_row(currency_country_data)
+      # currency_country_data is an array with 2 elements
```
willfish, line 45: You could use named indexes over first/last? and remove this comment?

willfish: e.g. 

```ruby
COUNTRY_NAME_INDEX = 0
CURRENCY_RATE_INDEX = 1
```

spec/models/exchange_rate_currency_rate_spec.rb
```diff
@@ -7,13 +7,45 @@
...
+    context 'when only a month is provided' do
+      before do
+        travel_to Time.zone.local(2023, 7, 1)
+      end
+
+      after do
+        travel_back
+      end
+
+      # Timecop.freeze(2023, 7, 1) do
```
willfish, line 36: I guess we can remove this?

spec/support/time_helpers.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.configure do |config|
```
willfish, line 1: I guess I have to look at the directory to know this exists :thinking: rather than cursory look the spec helper

willfish: I guess its pretty idiomatic

spec/models/exchange_rate_currency_rate_spec.rb
shjohnson: yes my mistake!

app/models/exchange_rate_currency_rate.rb
shjohnson: Yeah this is probably a good way to go, i guess sometimes i was worried it might end up looking super long in some places. `ExchangeRateCurrencyRate.exchange_rate_currency` but happy either way

app/models/exchange_rate_currency_rate.rb
shjohnson: Yeah i couldnt think any more variations, will have a think though

app/services/exchange_rates/create_csv_service.rb
shjohnson: Happy either way, i wasnt going to leave a comment but just incase got confusing using the hash structure/array

app/models/exchange_rate_currency_rate.rb
willfish: Yeah, fair enough

app/models/exchange_rate_currency_rate.rb
willfish: I'm easy on this really. I just generally expect the name of the method and the name of the thing coming back from the method are the same

app/models/exchange_rate_currency_rate.rb
maverickvi: @willfish we would need a join table to create a many to many here. I think we can keep it like this or do you recommend anything else?

willfish: So I figured this is a one to many but without a through? Looks like Sam has this working now?

If we don't use properly defined associations to pull related resources we end up without the ability to eager load stuff. I'll have a proper look tomorrow

maverickvi: I haven't had time to review these new changes yet. I will take a look tomorrow.

app/models/exchange_rate_currency_rate.rb
```diff
@@ -4,6 +4,9 @@ class ExchangeRateCurrencyRate < Sequel::Model
   RATES_FILE = 'data/exchange_rates/all_rates.csv'.freeze
   SPOT_RATES_FILE = 'data/exchange_rates/all_spot_rates.csv'.freeze
 
+  many_to_one :exchange_rate_currency, key: :currency_code, primary_key: :currency_code, class_name: ExchangeRateCurrency
+  many_to_many :countries, key: :currency_code, primary_key: :currency_code, class: ExchangeRateCountry
```
willfish, line 8:
```ruby
  one_to_many :exchange_rate_countries, key: :currency_code, primary_key: :currency_code
```

app/models/exchange_rate_currency_rate.rb
```diff
@@ -4,6 +4,9 @@ class ExchangeRateCurrencyRate < Sequel::Model
   RATES_FILE = 'data/exchange_rates/all_rates.csv'.freeze
   SPOT_RATES_FILE = 'data/exchange_rates/all_spot_rates.csv'.freeze
 
+  many_to_one :exchange_rate_currency, key: :currency_code, primary_key: :currency_code, class_name: ExchangeRateCurrency
```
willfish, line 7: You really ought not to need the class name now, this is worked out with the equivalent of inflections

willfish (approved):

PR #1380, opened 2023-07-07
HOTT-3576: Adds tradeset descriptions to bulk search index

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3576

### What?

I have added/removed/altered:

- [x] Adds association for tradeset descriptions
- [x] Adds tradeset descriptions to index definition
- [x] Adds utility methods to stream tradeset_descriptions zip file
- [x] Adds tradeset_description model and specs
- [x] Adds tradeset descriptions populator

### Why?

I am doing this because:

- This is required because we've proven that this improves the accuracy of search results based on curated test data that was produced by the classification team


maverickvi (dismissed):

app/lib/tariff_synchronizer/file_service.rb
```diff
@@ -48,6 +48,61 @@ def file_presigned_url(file_path)
...
+            while chunk
+              output.write(chunk)
+              chunk = input.read(16 * 1024)
+            end
+          end
+        end
+      end
+
+      def list_by(prefix:)
+        if Rails.env.production?
```
alexdesi, line 88: same for this.

app/lib/tariff_synchronizer/file_service.rb
```diff
@@ -48,6 +48,61 @@ def file_presigned_url(file_path)
...
+        end
+
+        download(source:, destination:)
+        gunzip(source: destination, destination: destination.to_s.gsub('.gz', ''))
+
+        Dir.glob("#{File.dirname(destination)}/*").reject { |file| file.end_with?('.zip') }
+      end
+
+      def download(source:, destination:)
+        if Rails.env.production?
```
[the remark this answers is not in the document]
willfish: This is actually a great idea. I'll put in a story to move the persistence/reporting bucket interactions to use ActiveStorage

app/services/tradeset_description_populator_service.rb
```diff
@@ -0,0 +1,92 @@
...
+    csv.each do |row|
+      attributes = row.to_hash
+      attributes[:classification_date] = classification_date
+      attributes[:filename] = filename
+      attributes[:created_at] = created_at
+      attributes[:updated_at] = updated_at
+
+      tradeset_description = TradesetDescription.build(attributes)
+
+      unless chunk.size >= CHUNK_SIZE
```
alexdesi, line 43: we could use if instead of unless, it's shorter ;) 

`if chunk.size < CHUNK_SIZE`

app/services/tradeset_description_populator_service.rb
```diff
@@ -0,0 +1,92 @@
...
+
+        next
+      end
+
+      yield chunk
+
+      chunk = []
+    end
+
+    yield chunk unless chunk.empty?
```
alexdesi, line 54: it sounds like a double negative. 
What about: `yield chunk if chunk.any?`

app/services/tradeset_description_populator_service.rb
```diff
@@ -0,0 +1,92 @@
...
+      sample = File.read(file_path, encoding:, lines: 5)
+      sample.valid_encoding?
+    end
+  end
+
+  def classification_date
+    @classification_date ||= filename.match(FILENAME_REGEX) do |match|
+      month = match[:month]
+      year = match[:year]
+      Date.parse("1 #{month} #{year}").end_of_month
```
alexdesi, line 78: Is it oka date format to parse?
usually it goes this way: yyyy-mm-dd, for example:
`Date.parse('2001-02-03')`

alexdesi (commented):
The logic looks good!
I've suggested a few small improvements.

app/models/goods_nomenclature_description.rb
```diff
@@ -30,7 +28,7 @@ def description
   end
 
   def description_indexed
-    SearchNegationService.new(description).call
+    SearchNegationService.new(description.downcase).call
```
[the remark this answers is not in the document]
willfish: So the negation service is for removing negation terms. I figured I could create a new service but it seemed overkill for just downcasing a string

app/services/tradeset_description_populator_service.rb
willfish: It works yeah. But since I'm in control of the format of course I can change it to be more idiomatic

willfish: Hang on. No the reason I'm doing this is because I have DEC/JAN/MAR, etc as the month format and converting these is an extra step I don't need

app/models/tradeset_description.rb
```diff
@@ -0,0 +1,67 @@
...
+      description.updated_at = attributes[:updated_at]
+      description.classification_date = attributes[:classification_date]
+      description.validity_start_date = attributes[:classification_date]
+      description.validity_end_date = nil
+      description
+    end
+
+    private
+
+    def normalise_goods_nomenclature_item_id(goods_nomenclature_item_id)
```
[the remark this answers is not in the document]
willfish: Not for the format from the file we receive, no.

alexdesi (approved):
It looks good! thanks for the changes :)

PR #1389, opened 2023-07-12
iam for access to secret manager

e-d-o-c:
WORK IN PROGRESS -  NOT READY FOR REVIEW

### Jira link

HOTT- 3617

### What?

I have added/removed/altered:

- [ ] Added iam role for the services to gain access to secret manager

### Why?

I am doing this because, it will allow the service to fetch redis url from secret manager


amberstarlight:
Is this ready for review now?

terraform/variables.tf
```diff
@@ -86,5 +86,5 @@ variable "base_domain" {
 
 variable "trade_tariff_oauth_secret" {
   description = "trade tarrif oauth secrets."
-  type        = "map"
+  type        = map(string)
```
amberstarlight, line 89:
```suggestion
  type        = string
```

I think this should just be a `string` here, because this is the value of `TRADE_TARIFF_OAUTH_SECRET`

terraform/iam.tf
```diff
@@ -0,0 +1,37 @@
...
+    resources = [
+      data.aws_kms_key.secretsmanager_key.arn
+    ]
+  }
+}
+
+resource "aws_iam_policy" "secrets" {
+  name   = "backend-execution-role-secrets-policy"
+  policy = data.aws_iam_policy_document.secrets.json
+}
```
amberstarlight, line 37: Can you run the pre-commit hooks? This is missing a newline at the end of the file.

amberstarlight (changes requested):

amberstarlight (approved):
Approved but we'll need to fix the smoketests before merging

PR #1390, opened 2023-07-13
HOTT-3268: This PR allows to generate the quota exclusion misalignment report

alexdesi:
### What?
This PR adds the report generator for the quota exclusion misalignment.

### Jira link

https://transformuk.atlassian.net/browse/HOTT-3268

### Why?

the UK data of measure geographical area exclusions and quota origin exclusions should have the same data.
In case of misalignment, we must report it to HMRC

### DIFF
--- a/app/lib/reporting/differences.rb
+++ b/app/lib/reporting/differences.rb
@@ -78,7 +78,7 @@ module Reporting
       @as_of = Time.zone.today.iso8601
     end
 
-    def generate
+    def generate(only: [])
       total_start = Time.zone.now
 
       methods = %i[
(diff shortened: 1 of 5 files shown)


app/lib/reporting/differences.rb
```diff
@@ -78,7 +78,7 @@ def initialize
       @as_of = Time.zone.today.iso8601
     end
 
-    def generate
+    def generate(only: [])
```
alexdesi, line 81: This is not strictly part of the PR,
but it's handy to generate only the report we need, for example, during development.

willfish (approved):

willfish (approved):

PR #1392, opened 2023-07-14
HOTT-3641: Extract shared component behaviour into concern

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3641

### What?

I have added/removed/altered:

- [x] Added Componentable module and extracted shared behaviour for MeasureConditionComponent and MeasureComponent

### Why?

I am doing this because:

- This is required to make changes in one place and to simplify a duplicated api
- **This is a non-functional refactor**

### DIFF
--- a/app/models/concerns/componentable.rb
+++ b/app/models/concerns/componentable.rb
@@ -1,0 +1,71 @@
+module Componentable
+  SMALL_PRODUCERS_QUOTIENT = 'SPQ'.freeze
+  LITERS_OF_PURE_ALCOHOL_QUALIFIER = 'LPA'.freeze
+  LITERS_QUALIFIER = 'LTR'.freeze
+
+  extend ActiveSupport::Concern
+
+  included do
+    plugin :time_machine
+
+    one_to_one :duty_expression, key: :duty_expression_id,
+                                 primary_key: :duty_expression_id do |ds|
+      ds.with_actual(DutyExpression)
+    end
+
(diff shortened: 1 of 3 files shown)


app/models/concerns/componentable.rb
```diff
@@ -0,0 +1,71 @@
+module Componentable
+  SMALL_PRODUCERS_QUOTIENT = 'SPQ'.freeze
+  LITERS_OF_PURE_ALCOHOL_QUALIFIER = 'LPA'.freeze
+  LITERS_QUALIFIER = 'LTR'.freeze
```
alexdesi, line 4: 👍  for not using just the acronyms!

app/models/concerns/componentable.rb
```diff
@@ -0,0 +1,71 @@
...
+      measurement_unit_code == 'ASV' && measurement_unit_qualifier_code == 'X'
+    end
+
+    def liters_of_pure_alcohol?
+      measurement_unit_code == 'LPA'
+    end
+
+    def ad_valorem?
+      monetary_unit_code.nil? &&
+        measurement_unit_code.nil?
```
alexdesi, line 60: nitpick: on a single line?

alexdesi (approved):
It all makes sense, 
my only concern is that "componentable" is a weak name (what's a component?)
I don't have a better name anyway :)

PR #1398, opened 2023-07-18
BAU: Fixes smoke test failures

willfish:
### Jira link

BAU

### What?

We created a method for presenting the description that is compatible with being indexed (and therefore full-text-searched in opensearch). This method was recently extended to consistently downcase descriptions. It also removes negation terms from a description as part of a negation service.

This highlighted that the wrong description (the indexed description) was being used in certain objects that are being presented in the backend search results api as indicated by smoketest failures.

By correcting this, our smoke tests should now pass since we're returning the correct `formatted_description` rather than the incorrect `description_indexed`

I have added/removed/altered:

- [x] Fixes references to the indexed descriptions of chapters/headings

### Why?

I am doing this because:

- These should always have been the formatted descriptions


amberstarlight (dismissed):
LGTM

PR #1399, opened 2023-07-18
HOTT-3639: Add presenter to override duty amount on alcohol excise units

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3639

https://dev.trade-tariff.service.gov.uk/commodities/2203000100?day=21&month=8&year=2023#vat_excise

UK

image

XI

image

### What?

For ASVX condition components on excise alcohol measures we need to coerce the
first duty amount to be balanced so its no longer making adjustments to handle
liters/hectoliter conversion.

For ASV condition components on excise alcohol measures the ASV part is presenting too small `0.01` instead of `1.00%` and this is incorrect.

I have added/removed/altered:

- [x] Added presentation adjustments to alcohol duty amounts for ASVX
- [x] Added presentation adjustments to alcohol condition duty amounts for ASV

### Why?

I am doing this because:

- This is required so that the presentation of the duty amount reflects the legislation/intended duty amounts

### DIFF
--- a/app/lib/trade_tariff_backend.rb
+++ b/app/lib/trade_tariff_backend.rb
@@ -269,7 +269,12 @@ module TradeTariffBackend
     end
 
     def excise_alcohol_coercian_starts_from
-      Date.parse(ENV.fetch('ALCOHOL_COERCIAN_STARTS_FROM', '2023-08-01'))
+      @excise_alcohol_coercian_starts_from ||= Date.parse(
+        ENV.fetch(
+          'ALCOHOL_COERCIAN_STARTS_FROM',
+          '2023-08-01',
+        ),
+      )
     end
 
     def revision
--- a/app/lib/trade_tariff_backend.rb
+++ b/app/lib/trade_tariff_backend.rb
@@ -268,6 +268,10 @@ module TradeTariffBackend
       ENV.fetch('OPENSEARCH_DEBUG', 'false') == 'true'
     end
 
+    def excise_alcohol_coercian_starts_from
+      Date.parse(ENV.fetch('ALCOHOL_COERCIAN_STARTS_FROM', '2023-08-01'))
+    end
+
     def revision
       @revision ||= begin
         File.read(REVISION_FILE).chomp if File.file?(REVISION_FILE)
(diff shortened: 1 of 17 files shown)


app/lib/trade_tariff_backend.rb
```diff
@@ -268,6 +268,10 @@ def opensearch_debug
       ENV.fetch('OPENSEARCH_DEBUG', 'false') == 'true'
     end
 
+    def excise_alcohol_coercian_starts_from
+      Date.parse(ENV.fetch('ALCOHOL_COERCIAN_STARTS_FROM', '2023-08-01'))
```
willfish, line 272: Memoize me

app/presenters/api/v2/measures/measure_condition_presenter.rb
```diff
@@ -0,0 +1,73 @@
+module Api
+  module V2
+    module Measures
+      class MeasureConditionPresenter < WrapDelegator
+        delegate :excise_alcohol_coercian_starts_from, to: TradeTariffBackend
+
+        ALCOHOL_PERCENTAGE_MEASUREMENT_UNIT_CODE = 'ASV'.freeze
```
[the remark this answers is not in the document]
willfish: The acronym stands for that, yeah.

app/presenters/api/v2/measures/measure_condition_presenter.rb
```diff
@@ -0,0 +1,74 @@
...
+          end
+        end
+
+        private
+
+        attr_reader :measure, :measure_condition
+
+        def apply_coerced_condition_duty_amount_conversion_factor?
+          return false if Time.zone.today < excise_alcohol_coercian_starts_from
+          return false if MeasureCondition.point_in_time.present? && MeasureCondition.point_in_time < excise_alcohol_coercian_starts_from
```
alexdesi, line 63: These are duplicated, but it's now a big deal thb.

alexdesi (dismissed):
nice one!
just a minor small comment, but it's totally fine for now.
🚢 🇮🇹

app/presenters/api/v2/measures/measure_condition_presenter.rb
willfish: I think any solution solving the duplication would be worse/more complex than just having a bit of duplication, here for the moment so going to leave it for now. You are right, though

PR #1407, opened 2023-07-26
API - exchange_rate_period: Monthly exchange rates

shjohnson:
### Jira link

[HOTT-3475](https://transformuk.atlassian.net/browse/HOTT-3475)

### What?

I have added/removed/altered:

- [ ] Created serializers for exchange rate list, exchange rates and files
- [ ] Create api for returner the JSON api schema with tests


app/models/exchange_rate_currency_rate.rb
```diff
@@ -15,6 +15,10 @@ def spot_rate?
     [3, 12].include?(validity_start_date.month) && validity_start_date.day == 31 && validity_end_date.nil?
   end
 
+  def countries
```
shjohnson, line 18: Sorry il remove this, this is from a rebasing issue

app/models/exchange_rates/rates_list.rb
```diff
@@ -0,0 +1,45 @@
...
+    def exchange_rate_ids
+      exchange_rates.map(&:id)
+    end
+
+    class << self
+      def build(month, year)
+        rates_list = new
+        rates_list.year = year
+        rates_list.month = month
+        rates_list.publication_date = '2023-06-22T00:00:00.000Z'
```
shjohnson, line 26: This i need to fix and was going to discuss later as we are adding a new table

shjohnson:
Just noticed few test files might not have been included in the last rebase will check out 👀

app/controllers/api/v2/exchange_rates/monthly_exchange_rates_controller.rb
```diff
@@ -0,0 +1,40 @@
+module Api
+  module V2
+    module ExchangeRates
+      class MonthlyExchangeRatesController < ApiController
+        def show
+          if month.nil? || year.nil?
+            render json: { "errors": [{ "detail": 'not found' }] }, status: :not_found
+          else
+            render json: serialized_exchange_rates_list
+          end
```
alexdesi, line 10: what about positive scenario first?
```
if month && year
              render json: serialized_exchange_rates_list
else
...
```

app/models/exchange_rate_currency_rate.rb
```diff
@@ -44,6 +48,16 @@ def by_year_and_month(month, year = Time.zone.today.year)
...
         .order(Sequel.desc(:validity_start_date))
     end
+
+    def files_for_year_and_month(month, year = Time.zone.today.year)
+      return if month.blank? || year.blank?
+
+      start_of_month = Time.zone.parse("#{year}-#{month}-01").beginning_of_month
+      end_of_month = start_of_month.end_of_month
+
+      where { (validity_start_date >= start_of_month) & (validity_start_date <= end_of_month) }
```
alexdesi, line 58: not sure here,
could we use here the "TimeMachine"?

alexdesi (commented):
need to finish the review, just a couple of comment for now

app/models/exchange_rate_currency_rate.rb
shjohnson: hmm I dont think so, i think this would be slightly different but good to raise!

app/models/exchange_rate_currency_rate.rb
```diff
@@ -81,6 +95,15 @@ def max_year
         &.year.presence || Time.zone.today.year
     end
 
+    def max_month
```
alexdesi, line 98: what about calling this `most_recent_month`,
also, silly question, where is it used?

app/models/exchange_rates/exchange_rate_file.rb
```diff
@@ -0,0 +1,25 @@
+module ExchangeRates
+  class ExchangeRateFile
```
maverickvi, line 2: We already have this class in app/models, why not use that?

app/models/exchange_rates/rates_list.rb
```diff
@@ -0,0 +1,45 @@
...
+        rates_list.year = year
+        rates_list.month = month
+        rates_list.publication_date = exchange_rate_files(month, year).first.publication_date
+        rates_list.exchange_rate_files = exchange_rate_files(month, year)
+        rates_list.exchange_rates = exchange_rates(month, year)
+        rates_list
+      end
+
+      def exchange_rate_files(month, year)
+        files = ExchangeRateCurrencyRate.files_for_year_and_month(month, year)
```
maverickvi, line 33: You can use ExchangeRateFile.where(month:, year:) here and you can delete the duplicate ExchangeRateFile class in app/models/exchange_rates

app/models/exchange_rates/exchange_rate.rb
```diff
@@ -0,0 +1,37 @@
+module ExchangeRates
+  class ExchangeRate
```
maverickvi, line 2: why not use the ExchangeRateCurrencyRate class?

shjohnson: reason for not using the other file is it this one has different attributes and didnt think worth adding them to the save model and the DB?

app/controllers/api/v2/exchange_rates/monthly_exchange_rates_controller.rb
```diff
@@ -0,0 +1,40 @@
+module Api
+  module V2
+    module ExchangeRates
+      class MonthlyExchangeRatesController < ApiController
+        def show
+          if month && year
+            render json: serialized_exchange_rates_list
+          else
+            render json: { "errors": [{ "detail": 'not found' }] }, status: :not_found
```
willfish, line 9: I think we can just raise an exception to be more idiomatic with the rest of the controllers?

willfish: Okay this is a can of  :worm: :worm:

willfish (approved):

willfish (approved):

PR #1410, opened 2023-07-26
HOTT-3642: Adjust SPQ presentation

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3642

image

### What?

I have added/removed/altered:

- [x] Added expanded presentation of SPQ units on condition duty expression components
- [x] Added test coverage for the conditional expansion of SPQ unit components

### Why?

I am doing this because:

- This is required because the default `SPQ  (SPR * Rate)` is just incorrect

### DIFF
--- a/app/presenters/api/v2/measures/measure_condition_presenter.rb
+++ b/app/presenters/api/v2/measures/measure_condition_presenter.rb
@@ -31,15 +31,7 @@ module Api
         end
 
         def duty_expression
-          formatted_components = measure_condition_components.map do |measure_condition_component|
-            if measure_condition_component.small_producers_quotient?
-              measure_condition_component.verbose_duty_expression
-            else
-              measure_condition_component.formatted_duty_expression
-            end
-          end
-
-          formatted_components.join(' ')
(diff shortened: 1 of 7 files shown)


app/presenters/api/v2/measures/measure_condition_presenter.rb
```diff
@@ -31,7 +31,15 @@ def condition_duty_amount
         end
 
         def duty_expression
-          measure_condition_components.map(&:formatted_duty_expression).join(' ')
+          formatted_components = measure_condition_components.map do |measure_condition_component|
+            if measure_condition_component.small_producers_quotient?
+              measure_condition_component.verbose_duty_expression
+            else
+              measure_condition_component.formatted_duty_expression
+            end
```
alexdesi, line 39: It seems we should add this logic into a method in the `measure_condition_component` ,

for example

`measure_condition_components.map(&:presented_duty_expression)`

willfish: Yeah, this is a great suggestion. Implemented this

alexdesi (dismissed):
💚

PR #1411, opened 2023-07-26
HOTT-3667: Show threshold conditions that are negative

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3667

image

### What?

I have added/removed/altered:

- [x] Show measure conditions that are negative in the permutations UI
- [x] Refactor permutations logic to reflect the real world examples

### Why?

I am doing this because:

- This is required as part of the alcohol work to make sure users can see when excise does not apply to them

### DIFF
--- a/app/models/measure_condition.rb
+++ b/app/models/measure_condition.rb
@@ -1,5 +1,6 @@
 class MeasureCondition < Sequel::Model
   CDS_WAIVER_DOCUMENT_CODE = '999L'.freeze
+  INCLUDED_NEGATIVE_ACTIONS = %w[08].freeze
 
   plugin :time_machine
   plugin :national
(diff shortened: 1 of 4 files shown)


app/models/measure_condition.rb
```diff
@@ -172,7 +174,19 @@ def unit_for(measure)
...
   end
 
-private
+  def is_excluded_condition?
+    return true if universal_waiver_applies?
+    return false if negative_action? && is_threshold?
+    return false if negative_action? && action_code.in?(INCLUDED_NEGATIVE_ACTIONS)
+
+    negative_action? && document_code.blank?
+  end
```
alexdesi, line 183: Thinking about how to simplify this.
Maybe this way? 🤔  (it just checks `negative_action?` once)
```
  def is_excluded_condition?
    return true if universal_waiver_applies?

    if negative_action?
      is_threshold? || action_code.in?(INCLUDED_NEGATIVE_ACTIONS)
    else
      document_code.blank?
    end
  end

```

willfish: So we're covering these scenarios:

*globally excluded conditions*

1. universal waiver

*negative conditions that are included*

1. negative action and threshold is positive (really common)
2. negative action and the action code is 08 (common)

*negative conditions that are excluded*

1. negative action and document is blank (really common)

PR #1414, opened 2023-07-28
BAU: Fixes bug in conditional display of alcohol duties

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Fixes conditional display of alcohol duties

### Why?

I am doing this because:

- This is required since we're not showing these display adjustments in production and we should be

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,6 +1,6 @@
 PGHOST=localhost
 AGGREGATED_SYNONYMS_FILE=spec/fixtures/beta/search/goods_nomenclatures/test_synonyms.txt
-ALCOHOL_COERCIAN_STARTS_FROM=2022-01-01
+ALCOHOL_COERCIAN_STARTS_FROM=2023-08-01
 AWS_ACCESS_KEY_ID=xxxxxxxxxxxxxxxxxxxx
 AWS_BUCKET_NAME=trade-tariff-backend
 AWS_REPORTING_BUCKET_NAME=trade-tariff-reporting
--- a/app/presenters/api/v2/measures/measure_condition_component_presenter.rb
+++ b/app/presenters/api/v2/measures/measure_condition_component_presenter.rb
@@ -55,8 +55,7 @@ module Api
         attr_reader :measure
 
         def apply_coerced_duty_amount_conversion_factor?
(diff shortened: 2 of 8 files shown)


app/presenters/api/v2/measures/measure_condition_component_presenter.rb
```diff
@@ -55,8 +55,7 @@ def duty_expression_str
         attr_reader :measure
 
         def apply_coerced_duty_amount_conversion_factor?
-          return false if Time.zone.today < excise_alcohol_coercian_starts_from
-          return false if MeasureCondition.point_in_time.present? && MeasureCondition.point_in_time < excise_alcohol_coercian_starts_from
+          return false if point_in_time.present? && point_in_time < excise_alcohol_coercian_starts_from
```
willfish, line 58: We only need to care about the point in time not the actual current date otherwise these won't display until the 1st of August 2023

.env.test
```diff
@@ -1,6 +1,6 @@
 PGHOST=localhost
 AGGREGATED_SYNONYMS_FILE=spec/fixtures/beta/search/goods_nomenclatures/test_synonyms.txt
-ALCOHOL_COERCIAN_STARTS_FROM=2022-01-01
+ALCOHOL_COERCIAN_STARTS_FROM=2023-08-01
```
willfish, line 3: Just explicitly reflect what go live will look like

spec/models/measure_spec.rb
```diff
@@ -1232,8 +1248,45 @@
       it { expect(measure.units).to eq(expected_units) }
     end
 
-    context 'when there are just measure conditions' do
```
alexdesi, line 1235: Maybe we can keep the context 'when there are just measure conditions' ?
so we can keep just a single ' subject(:measure)' ? 

Nitpick: can we use "only" instead of "just" ... it sounds better 😄

willfish: Yeah, I can do that for sure :)

alexdesi (approved):
🚀 ✅ 🚀 ✅ 🚀 ✅

PR #1418, opened 2023-08-03
HOTT-3766: Extend measurement unit types

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3766

### What?

For a bit of context, these types are used to determine the class of measure condition. I realised the types are actually interesting for validation, too, since they indicate how a measurement unit should be treated.

A whole bunch of these were missing/wrong in the measurement unit file (some were classified as a volume when they are in fact a weight. Some were just missing and these needed populating.

I've also opted to memoize and use a set for 0(1) lookup since the iteration happens a lot (for every measure condition on a returned commodity and for every commodity we're loading).

I have added/removed/altered:

- [x] Added measurement unit types for all measurement units

### Why?

I am doing this because:

- This will be used in the duty calculator to perform proper measurement unit calculations

### DIFF
--- a/app/models/measurement_unit.rb
+++ b/app/models/measurement_unit.rb
@@ -29,15 +29,27 @@ class MeasurementUnit < Sequel::Model
     end
 
     def weight_units
-      measurement_units.select { |_k, v| v['measurement_unit_type'] == 'weight' }.keys
+      @weight_units ||= begin
+        units = measurement_units.select { |_k, v| v['measurement_unit_type'] == 'weight' }.keys
+
+        Set.new(units)
+      end
     end
 
     def volume_units
-      measurement_units.select { |_k, v| v['measurement_unit_type'] == 'volume' }.keys
+      @volume_units ||= begin
+        units = measurement_units.select { |_k, v| v['measurement_unit_type'] == 'volume' }.keys
+
+        Set.new(units)
+      end
     end
 
     def percentage_abv_units
-      measurement_units.select { |_k, v| v['measurement_unit_type'] == 'percentage_abv' }.keys
+      @percentage_abv_units ||= begin
+        units = measurement_units.select { |_k, v| v['measurement_unit_type'] == 'percentage_abv' }.keys
+
+        Set.new(units)
+      end
     end
 
     def measurement_units
--- a/db/measurement_units.json
+++ b/db/measurement_units.json
@@ -26,7 +26,8 @@
     "compound_units": [
       "ASV",
       "HLT"
-    ]
+    ],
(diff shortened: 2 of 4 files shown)


app/models/measurement_unit.rb
```diff
@@ -29,15 +29,27 @@ def units(unit_code, unit_key)
     end
 
     def weight_units
-      measurement_units.select { |_k, v| v['measurement_unit_type'] == 'weight' }.keys
+      @weight_units ||= begin
+        units = measurement_units.select { |_k, v| v['measurement_unit_type'] == 'weight' }.keys
+
+        Set.new(units)
```
willfish, line 35: Use a set and memoise to make frequent lookups less expensive

app/models/measurement_unit.rb
```diff
@@ -29,15 +29,27 @@ def units(unit_code, unit_key)
...
     end
 
     def percentage_abv_units
-      measurement_units.select { |_k, v| v['measurement_unit_type'] == 'percentage_abv' }.keys
+      @percentage_abv_units ||= begin
+        units = measurement_units.select { |_k, v| v['measurement_unit_type'] == 'percentage_abv' }.keys
+
+        Set.new(units)
+      end
     end
```
alexdesi, line 53: I see a similar pattern here, 
maybe we can simplify those `*_units` methods with something like this:

```
    def get_units_for(unit_type)
      units = measurement_units.select { |_k, v| v['measurement_unit_type'] == unit_type }.keys
      Set.new(units)
    end
```
and then call it 
`get_units_for('weight')` 
for example.

Of course, this is valid if the memoisation is not strictly necessary.

db/measurement_units.json
```diff
@@ -613,7 +648,7 @@
     "multiplier": null,
     "coerced_measurement_unit_code": null,
     "original_unit": null,
-    "measurement_unit_type": "volume"
+    "measurement_unit_type": "number"
```
alexdesi, line 651: Are we sure about this?

"cubic metre" sounds like a volume :)

app/models/measurement_unit.rb
willfish: We definitely want the class level memoisation IMO since these will get called super frequently

db/measurement_units.json
willfish: :facepalm:

app/models/measurement_unit.rb
willfish: I think explicit is better for a pattern only used in one file, though

maverickvi (approved):

PR #1429, opened 2023-08-08
BAU: Renames all ns_<method_name> methods to just <method_name>

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [ ] Rename all occurances of `ns_<method_name>` to just `<method_name>`

### Why?

I am doing this because:

- We do not need api users to know about nested set internals


alexdesi (approved):
nice de-ns-tification!

PR #1431, opened 2023-08-08
HOTT-3755: Footnote search enhancements

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3755

### What?

I have added/removed/altered:

- [x] Adds finder for footnote search
- [x] Adds coverage for new finder

### Why?

I am doing this because:

- This is required to enable a better search experience

### DIFF
--- a/app/models/goods_nomenclature.rb
+++ b/app/models/goods_nomenclature.rb
@@ -132,7 +132,7 @@ class GoodsNomenclature < Sequel::Model
       where(producline_suffix: '80')
     end
 
-    def with_footnotes
+    def join_footnotes
       association_right_join(:footnotes)
         .exclude(goods_nomenclatures__goods_nomenclature_item_id: nil)
     end
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -211,13 +211,7 @@ class Measure < Sequel::Model
       where(additional_code_id:)
     end
 
-    def with_all_measure_conditions
-      # There are more measure conditions than measures
-      # and we do not want to miss any measure conditions
-      # when joining them to measures.
-      #
-      # We then reject rows where measure_sid is nil (which is effectively an inner
-      # join on all measure conditions)
+    def join_measure_conditions
       association_right_join(:measure_conditions)
         .where(Sequel.~(measures__measure_sid: nil))
     end
@@ -245,7 +239,7 @@ class Measure < Sequel::Model
       where(combined_conditions)
     end
 
-    def with_footnotes
+    def join_footnotes
       association_right_join(:footnotes)
         .exclude(measures__measure_sid: nil)
     end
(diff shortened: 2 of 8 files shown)


app/models/goods_nomenclature.rb
```diff
@@ -131,6 +131,34 @@ def non_classifieds
...
+
+    def with_footnotes
+      association_right_join(:footnotes)
+        .exclude(goods_nomenclatures__goods_nomenclature_item_id: nil)
+    end
+
+    def with_footnote_type_id(footnote_type_id)
+      return self if footnote_type_id.blank?
+
+      where(footnotes__footnote_type_id: footnote_type_id)
```
alexdesi, line 143: do we need this kind of method?

is not ` where(footnotes__footnote_type_id: footnote_type_id)` self explanatory

willfish: So the `return self` bit preserves the default dataset when the user supplied value is null. It means I can easily chain these methods together to filter in based on the inputs the user provides

app/services/footnote_finder_service.rb
```diff
@@ -0,0 +1,137 @@
...
+  end
+
+  def grouped_goods_nomenclatures_by_measure_association
+    Measure
+      .with_regulation_dates_query
+      .actual
+      .with_footnotes
+      .with_footnote_type_id(type)
+      .with_footnote_id(id)
+      .with_footnote_types_and_ids(description_types_and_ids)
```
alexdesi, line 81: some of these method named "with_..." are join,
others are where, it could be a bit confusing:

for example:
`
.with_footnotes  # this one is a join
.with_footnote_type_id(type) # this one is a where
`
maybe we should keep the `join` naming?
for example:
```      
.with_footnotes # -> join_footnotes
```
or maybe call 
` .with_footnote_id(id)`

` .select_footnote_id(id)`

app/models/measure.rb
```diff
@@ -245,6 +239,34 @@ def with_certificate_types_and_codes(certificate_types_and_codes)
...
+      return self if footnote_id.blank?
+
+      where(footnotes__footnote_id: footnote_id)
+    end
+
+    def with_footnote_types_and_ids(footnote_types_and_ids)
+      return self if footnote_types_and_ids.none?
+
+      conditions = footnote_types_and_ids.map do |type, id|
+        Sequel.expr(footnotes__footnote_type_id: type) & Sequel.expr(footnotes__footnote_id: id)
```
alexdesi, line 263: Is not `&` a bitwise and?
I guess it should be `&&` or `and`

alexdesi (approved):
✅

app/models/measure.rb
willfish: This is a logical operator, not a bitwise operator in Sequel, basically


image

PR #1435, opened 2023-08-09
HOTT-3788 - Create ECS Service (worker)

venkatp227:
### Jira link

HOTT-3788

### What?

I have added:

- [x] Added ECS Service for worker-uk
- [x] Added ECS Service for worker-xi

### Why?

I am doing this because:

- Migration off gov.uk PaaS

### DIFF
--- a/terraform/worker_uk.tf
+++ b/terraform/worker_uk.tf
@@ -1,0 +1,74 @@
+module "worker_uk" {
+  source = "git@github.com:trade-tariff/trade-tariff-platform-terraform-modules.git//aws/ecs-service?ref=aws/ecs-service-v1.8.0"
+
+  service_name  = "worker-uk"
+  service_count = var.service_count
+  environment   = var.environment
+  region        = var.region
+
+  cluster_name              = "trade-tariff-cluster-${var.environment}"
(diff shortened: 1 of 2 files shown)


terraform/worker_uk.tf
```diff
@@ -0,0 +1,74 @@
...
+      {
+        name  = "TARIFF_FROM_EMAIL"
+        value = "Tariff UK [${title(var.environment)}] <${local.no_reply}>"
+      },
+      {
+        name  = "VCAP_APPLICATION"
+        value = "{}"
+      },
+      {
+        name  = "command"
```
willfish, line 65: Oh interesting, you can set a lowercase command to override the original run command?

venkatp227: Yes, we can do that https://aws.amazon.com/blogs/opensource/demystifying-entrypoint-cmd-docker/
Also I am not sure why the workers gracefullyy terminated after few mins of running. So if these are meant to be used for ad-hoc tasks, I have passed the CMD arugument to keep them running always.

venkatp227: https://awscli.amazonaws.com/v2/documentation/api/2.0.33/reference/ecs/run-task.html

PR #1437, opened 2023-08-10
BAU: Adds validations for search

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Added validations for search apis
- [x] Added test coverage for validations

### Why?

I am doing this because:

- This is required to flag to the user the required inputs


app/models/search_validator.rb
```diff
@@ -0,0 +1,16 @@
...
+  include ActiveModel::Model
+
+  attr_accessor :type, :code, :description
+
+  delegate :present?, to: :code, prefix: true
+  delegate :present?, to: :type, prefix: true
+
+  validates :type, presence: true, if: :code_present?
+  validates :code, presence: true, if: :type_present?
+  validates :description, presence: true, if: :code_and_type_blank?
```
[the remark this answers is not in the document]
willfish: We don't require a description if the code and type are blank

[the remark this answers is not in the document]
willfish: How comes? We need to make sure there is at least a description if there's no code or type input

app/services/api/search/error_serialization_service.rb
```diff
@@ -0,0 +1,35 @@
...
+      private
+
+      def errors
+        @records.each_with_object([]) do |record, acc|
+          record.errors.messages.each do |attribute, errors|
+            errors.each do |error|
+              acc << attribute_error(attribute, error)
+            end
+          end
+        end
```
alexdesi, line 23: Does it return `acc` at the end?

app/services/certificate_finder_service.rb
```diff
@@ -6,7 +6,7 @@ def initialize(type, code, description)
   end
 
   def call
-    return [] if description_types_and_codes.empty? && (type.blank? || code.blank?)
+    return [] if description_types_and_codes.empty? && type.blank? && code.blank?
```
alexdesi, line 9: just to say, the logic is different, but I guess is what you want

willfish: Yeah, this is correct - I want the api to do the control of what gets validated and just let users in scripts, etc, be able to filter on any of these fields

alexdesi (approved):
✅

PR #1438, opened 2023-08-14
BAU: Run Worker Commands

amberstarlight:
### What?

I have added/removed/altered:

- Added worker command
- Bumped worker module versions

### Why?

I am doing this because:

- The workers had commands passed as env vars, which is not what we want - we want to use the new module that supports overriding `entrypoint` and providing `command`.


amberstarlight:
Draft until [the module supports it](https://github.com/trade-tariff/trade-tariff-platform-terraform-modules/pull/37).

willfish (approved):

willfish (approved):

PR #1439, opened 2023-08-14
HOTT-3834: Support ancestor measures in declarable headings

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3834

### What?

I have added/removed/altered:

- [x] Added support for declarable heading ancestor measures

### Why?

I am doing this because:

- These were missing previously and were noticed by one of our stakeholders


willfish:
Merging to get the staging regression pack ran against this

PR #1445, opened 2023-08-15
HOTT-3611: Add show_proofs_for_geographical_areas to roo_scheme_uk.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-3611

### What?
To fix where the DCTS instructions are also showing, the attribute `show_proofs_for_geographical_areas` has been added to the roo scheme uk.

### Why?
To fix this:

image

### Have you? (optional)

### DIFF
--- a/app/models/rules_of_origin/scheme.rb
+++ b/app/models/rules_of_origin/scheme.rb
@@ -10,7 +10,7 @@ module RulesOfOrigin
 
     attr_reader :cumulation_methods, :validity_start_date, :validity_end_date
 
-    attr_writer :rule_sets, :proof_codes
+    attr_writer :rule_sets, :proof_codes, :show_proofs_for_geographical_areas
 
     delegate :read_referenced_file, to: :scheme_set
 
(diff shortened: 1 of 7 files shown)


app/models/rules_of_origin/data_set.rb
```diff
@@ -15,7 +15,7 @@ def load_default
     private
 
       def default_scheme_set
-        RulesOfOrigin::SchemeSet.from_default_file(TradeTariffBackend.service)
+        RulesOfOrigin::TradingScheme.from_default_file(TradeTariffBackend.service)
```
willfish, line 18: Is this reflective of the name Matt has chose for scheme of schemes or is this our choice of name?

willfish (dismissed):

app/models/rules_of_origin/data_set.rb
alexdesi: this is my choice, I changed it since Scheme is very generic,
and the proper name is Trading Scheme (see https://www.gov.uk/guidance/understanding-rules-of-origin-under-the-developing-countries-trading-scheme)
but, I realized I removed the "Set" too, which has actually meaning since there is not just a trading scheme

What about renaming it to `TradingSchemeSet`?

willfish (approved):

app/models/rules_of_origin/data_set.rb
alexdesi: we've clarified this point. No need to rename it.

PR #1448, opened 2023-08-15
BAU: Supply account ID to correct bucket names

amberstarlight:
### What?

I have added/removed/altered:

- Corrected the bucket names by using the account ID.
- Uncommented worker common vars.

### Why?

I am doing this because:

- I noticed that the S3 bucket names were missing the account ID suffix, so I've added a data source to fetch the current account ID.
- There were a couple of vars that were commented out but the worker could use having these.


alexdesi (approved):
It looks good ✅

PR #1453, opened 2023-08-17
HOTT-3579: Updates search suggestion primary keys

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3579

### What?

We've moved to using an id, type primary key constraint on the search suggestions since these are necessary to identify a unique row in the search suggestions table and to avoid duplication.

I have added/removed/altered:

- [x] Updated the primary key of search suggestions

### Why?

I am doing this because:

- This is needed to fix duplication patches

### DIFF
--- a/app/services/search_suggestion_populator_service.rb
+++ b/app/services/search_suggestion_populator_service.rb
@@ -5,7 +5,6 @@ class SearchSuggestionPopulatorService
     SearchSuggestion.unrestrict_primary_key
     TimeMachine.now do
       suggestions = SuggestionsService.new.call
-      suggestions = suggestions.uniq { |suggestion| [suggestion[:id], suggestion[:type]] }
 
       suggestions.each_slice(BATCH_SIZE) do |values|
         SearchSuggestion.dataset.insert_conflict(
@@ -51,6 +50,7 @@ class SearchSuggestionPopulatorService
 
     SearchSuggestion
       .where(id: expired_goods_nomenclature_sids.map(&:to_s))
+      .goods_nomenclature_type
       .delete
   end
 end
--- a/app/services/search_suggestion_populator_service.rb
+++ b/app/services/search_suggestion_populator_service.rb
@@ -1,16 +1,17 @@
 class SearchSuggestionPopulatorService
+  BATCH_SIZE = 5000
+
   def call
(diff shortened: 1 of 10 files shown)


db/migrate/20230817135045_adds_primary_key_for_id_and_type.rb
```diff
@@ -0,0 +1,15 @@
...
+  up do
+    alter_table(:search_suggestions) do
+      drop_constraint(:search_suggestions_pkey, type: :primary_key)
+      add_primary_key [:id, :type]
+    end
+  end
+
+  down do
+    alter_table(:search_suggestions) do
+      drop_constraint(:search_suggestions_pkey, type: :primary_key)
```
alexdesi, line 11: This is the same as line#4, is that right?

app/services/search_suggestion_populator_service.rb
```diff
@@ -1,16 +1,17 @@
 class SearchSuggestionPopulatorService
+  BATCH_SIZE = 5000
+
   def call
     SearchSuggestion.unrestrict_primary_key
     TimeMachine.now do
       suggestions = SuggestionsService.new.call
-      suggestions = suggestions.uniq { |suggestion| [suggestion[:id], suggestion[:value]] }
+      suggestions = suggestions.uniq { |suggestion| [suggestion[:id], suggestion[:type]] }
```
alexdesi, line 8: These lines are not affected by the TimeMachine block, aren't they?
Maybe we can move it outside the `TimeMachine.now do`

db/migrate/20230817135045_adds_primary_key_for_id_and_type.rb
willfish: No no, I'm changing from id, type back to id, value, here just in case I need to rollback

app/services/search_suggestion_populator_service.rb
willfish: I could do. I don't know why I would, though? It feels unnaffected/potentially like it would negatively impact the readability of the block :thinking: . I actually thing I can probably remove this line, anyway

willfish: Removed it

alexdesi (approved):
looks good to me ✅

PR #1454, opened 2023-08-17
HOTT-3705: Create rake task to create old files for previous data

shjohnson:
### Jira link

HOTT-3705

### What?

I have added/removed/altered:

- [x] Added rake task that loads old file dates and will fetch exchange rate correct data to load into files

### DIFF
--- a/app/services/exchange_rates/create_old_files_service.rb
+++ b/app/services/exchange_rates/create_old_files_service.rb
@@ -11,41 +11,50 @@ module ExchangeRates
     def call
       old_file_dates.each do |date_hash|
         month_name = date_hash[:edition_date].split(' ')[MONTH_INDEX]
-        month = Date::MONTHNAMES.index(month_name)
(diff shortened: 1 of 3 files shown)


app/services/exchange_rates/create_old_files_service.rb
```diff
@@ -0,0 +1,188 @@
...
+                    ExchangeRates::CreateXmlService,
+                    created_date,
+                    month,
+                    year)
+      end
+    end
+
+  private
+
+    def upload_data(data_result, format, service_class, created_date, month, year)
```
maverickvi, line 38: a lot of this code is duplicated from app/services/exchange_rates/upload_monthly_file_service.rb, wouldn't it be better to remove some of this duplication?

shjohnson: The reason i moved it here was mainly because i thought this file would probably be deleted once it had been run once as shouldnt need to be run again?

willfish (approved):

PR #1456, opened 2023-08-21
HOTT-3890: Expand countries for a given rate

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3890

### What?

Previously we would have returned Abu Dhabi as having multiple countries in a single row

image

I have added/removed/altered:

- [x] Expanded exchange rates with multiple countries so they appear as separate rows in the index of monthly exchange rates
- [x] Added a default order to exchange rates
- [x] Formatted the rate to 4 decimal places
- [x] Replaced non-restful apis with restful ones (e.g. show and use a composite id)
- [x] Added/updated all of the tests to reflect these changes

### Why?

I am doing this because:

- This is required to make sure we're returning single rows for the exchange rate UI and so that we don't have to do custom behaviour due to normalisation of the backend tables.


app/models/exchange_rate_currency_rate.rb
```diff
@@ -14,10 +14,24 @@ class ExchangeRateCurrencyRate < Sequel::Model
            to: :exchange_rate_currency,
            allow_nil: true
 
+  def_column_accessor :country, :country_code
```
willfish, line 17: For the unfamiliar, this essentially makes these fields that come from an association right join serializable in the api.

maverickvi (approved):

PR #1461, opened 2023-08-22
HOTT-3827: Additional CSV layout for HMRC

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3827

### What?

I have added/removed/altered:

- [ ] Added additional CSV layout for HMRC

### Why?

I am doing this because:

- We need to send this layout to HMRC


app/services/exchange_rates/upload_monthly_file_service.rb
```diff
@@ -35,15 +37,21 @@ def call
     def upload_data(data_result, format, service_class)
       data_string = service_class.call(data_result)
       file_path = "data/exchange_rates/monthly_#{format}_#{year}-#{month}.#{format}"
+      file_path = "data/exchange_rates/#{year}#{month_hmrc}MonthlyRates.csv" if type == :csv_hmrc
+
       TariffSynchronizer::FileService.write_file(file_path, data_string)
 
       file_size = TariffSynchronizer::FileService.file_size(file_path)
-      ::ExchangeRateFile.create(period_year: year, period_month: month, format:, file_size:, publication_date: current_time)
+      ::ExchangeRateFile.create(period_year: year, period_month: month, format:, file_size:, publication_date: current_time) unless type == :csv_hmrc
```
willfish, line 45: Is this a type or format?

willfish: So it definitely feels like we need to extend the api of the create method to include the report type, here

PR #1473, opened 2023-08-30
Better handling when no periods exist for exchange rates

shjohnson:
### Jira link

HOTT-3910

### What?

I have added/removed/altered:

- [ ] Added tests for handling no periods returned in the period list controller and the period list builder
- [ ] Added code to pass the tests

### Why?

This should not happen in a real world situation but handles if the user requests a year that has no periods in the exchange rate api


willfish:
Do we need a dedicated not found page: 
image

willfish (approved):

PR #1477, opened 2023-09-01
BAU: Run database migrations in AWS

amberstarlight:
### What?

I have added/removed/altered:

- Added a CircleCi job to run database migrations in the new AWS.
- Added shell script to do this.

### Why?

I am doing this because:

- We need to make sure we run the database migrations so that we don't run into situations where tables are missing or the application expects database configuration that is not yet reflected.


bin/migrate
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env sh
```
willfish, line 1: Do we want this to run in .circleci/bin/migrate?

amberstarlight: We could do. If so, I'd probably want to move `sync-opensearch-packages` too.

PR #1478, opened 2023-09-04
HOTT-3892: Retry job later if XE connection fails

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3892

### What?

I have added/removed/altered:

- [ ] Retry job later if XE connection fails

### Why?

I am doing this because:

- We want to retry the background job if XE API is down


alexdesi (approved):
✅

willfish (approved):

PR #1484, opened 2023-09-05
HOTT-2789: Run init container for all backends

amberstarlight:
### Jira link

[HOTT-2789](https://transformuk.atlassian.net/browse/HOTT-2789)

### What?

I have added/removed/altered:

- Upgraded the backend module versions to v1.12.0

### Why?

See https://github.com/trade-tariff/trade-tariff-platform-terraform-modules/pull/41

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -19,10 +19,10 @@ Terraform to deploy the service into AWS.
 
 | Name | Source | Version |
 |------|--------|---------|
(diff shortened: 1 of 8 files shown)


willfish:
Is it possible to make both the workers and the apis depend on the migration definition having been executed? I think concurrent migrations might potentially lead to odd behaviour. I now appreciate that this is what you probably mean with the `if I run this multiple times` question.

PR #1487, opened 2023-09-06
HOTT-3947: Make date passable for manual running and testing

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3947

### What?

I have added/removed/altered:

- [ ] Make date passable for manual running and testing

### Why?

I am doing this because:

-  We sometimes need to call the mailer manually and need to pass a date

### DIFF
--- a/app/mailers/exchange_rates_mailer.rb
+++ b/app/mailers/exchange_rates_mailer.rb
@@ -3,7 +3,10 @@ class ExchangeRatesMailer < ApplicationMailer
 
   default to: TradeTariffBackend.management_email
 
-  def monthly_files
+  attr_reader :date
+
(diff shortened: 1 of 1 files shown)


app/mailers/exchange_rates_mailer.rb
```diff
@@ -13,10 +16,6 @@ def monthly_files
 
   private
 
-  def date
-    @date ||= Time.zone.today
```
alexdesi, line 17: this way we loose the memoization, sure we do not need it?

alexdesi (approved):
it looks good, but verify we don't need the memoization (which allow to return the same date)

PR #1488, opened 2023-09-07
HOTT-3958: Remove New Relic

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3958

### What?

I have added/removed/altered:

- [x] Removed new relic from application

### Why?

I am doing this because:

- We're moving to data dog


amberstarlight (approved):
bye bye NewRelic 👋🏻

PR #1489, opened 2023-09-07
BAU: Adds variables for opensearch analyzer packages

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Added variables and configuration for analyzer packages

### Why?

I am doing this because:

- This is required for production and staging to interface with opensearch


amberstarlight (approved):
Awesome, this will fix those test failures

PR #1494, opened 2023-09-12
HOTT-3962: Changed file names on download endpoint and updated emails

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3962

### What?

I have added/removed/altered:

- [ ] Changed file names on download endpoint and updated emails

### Why?

I am doing this because:

- Filenames needed to match HMRC requirements
- Email should only send HMRC_CSV type


willfish (approved):
LGTM

PR #1498, opened 2023-09-13
HOTT-3975: Add rake tasks `import_agv_ex_rates` to import average exchange rates

alexdesi:
### What?
This PR imports data from the  CSV https://www.gov.uk/government/publications/exchange-rates-for-customs-and-vat-yearly
into ExchangeRateCurrencyRate table

### Jira link
https://transformuk.atlassian.net/browse/HOTT-3975

### DIFF
--- a/app/models/exchange_rate_currency_rate.rb
+++ b/app/models/exchange_rate_currency_rate.rb
@@ -1,6 +1,8 @@
 require 'csv'
 
 class ExchangeRateCurrencyRate < Sequel::Model
+  plugin :validation_helpers
+
   SCHEDULED_RATE_TYPE = 'scheduled'.freeze
   SPOT_RATE_TYPE = 'spot'.freeze
 
@@ -19,6 +21,12 @@ class ExchangeRateCurrencyRate < Sequel::Model
   include ContentAddressableId
 
   content_addressable_fields :currency_code, :validity_start_date, :validity_end_date, :country, :country_code
+
+  def validate
+    super
+
+    validates_unique(%i[currency_code validity_start_date validity_end_date rate_type])
+  end
 
   def presented_rate
     sprintf('%.4f', rate)
--- a/lib/tasks/exchange_rate_importer.rake
+++ b/lib/tasks/exchange_rate_importer.rake
@@ -1,0 +1,76 @@
+# Import TARIC or CDS file manually. Usually for initial seed files.
+namespace :importer do
+  namespace :rates do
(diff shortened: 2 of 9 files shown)


app/models/exchange_rate_currency_rate.rb
```diff
@@ -20,6 +22,12 @@ class ExchangeRateCurrencyRate < Sequel::Model
 
   content_addressable_fields :currency_code, :validity_start_date, :validity_end_date, :country, :country_code
 
+  def validate
+    super
+
+    validates_unique(%i[currency_code validity_start_date validity_end_date rate_type])
```
alexdesi, line 28: This validation is to avoid adding the same rate for the same period, dates and types more than once.

lib/tasks/exchange_rate_importer.rake
```diff
@@ -0,0 +1,76 @@
...
+
+      puts '- - - Outcome - - -'
+      puts "New exchage rates imported: #{new_records_count}"
+      puts "Invalid CSV rows: #{invalid_rows.count}"
+      puts "Invalid records: #{invalid_records.count}"
+    end
+
+    # rubocop:disable Rake/MethodDefinitionInTask
+    def invalid?(row)
+      row.to_hash.values.any?(&:empty?)
```
alexdesi, line 62: A CSV row must contain data for all the columns, otherwise, it is invalid.

lib/tasks/exchange_rate_importer.rake
```diff
@@ -0,0 +1,76 @@
+# Import TARIC or CDS file manually. Usually for initial seed files.
+namespace :importer do
+  namespace :rates do
+    desc 'Import Average Exchange Rates'
+
+    task import_agv_ex_rates: %i[environment] do
+      average_type = 'average'
+
+      url = 'https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1154375/Average_for_the_year_to_31_March_2023.csv'
```
maverickvi, line 9: we need to import all the rates from 2020 onwards. This only imports 1 file, It should be 2 files per year

app/models/exchange_rate_currency_rate.rb
shjohnson: Period wise do we need to include both start and end? considering so many are nil?

willfish: Don't we already handle this at the database level?

PR #1504, opened 2023-09-19
HOTT-3809: Support surfacing average and spot rates

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3809

### What?

I have added/removed/altered:

- [x] Added an exchange_rates api (supporting average, spot and scheduled types)
- [x] Altered the period list api (supporting average, spot and scheduled types)
- [x] Altered the files api  (supporting average, spot and scheduled types)
- [x] Altered the unique index on exchange rates to include rate_type
- [x] Removed a bunch of unused methods

### Why?

I am doing this because:

- All of this is required so we can surface exchange rates of different types (spot, average and the existing scheduled types)


app/controllers/api/v2/exchange_rates/period_lists_controller.rb
```diff
@@ -20,12 +20,26 @@ def serialized_period_list
...
           (params[:year].presence || ExchangeRateCurrencyRate.max_year).to_i
         end
+
+        def filter_params
+          params.fetch(:filter, {}).permit(:type)
+        end
+
+        def type
+          filter_params[:type]
+        end
```
alexdesi, line 36: I guess you should remove this

maverickvi: yeah I need to refactor and add tests, so don't bother reviewing yet. this is just to unblock you

app/controllers/api/v2/exchange_rates/period_lists_controller.rb
```diff
@@ -20,12 +20,26 @@ def serialized_period_list
...
+
+        def type
+          type_param = filter_params[:type]
+
+          type_param if [ExchangeRateCurrencyRate::SCHEDULED_RATE_TYPE, ExchangeRateCurrencyRate::SPOT_RATE_TYPE, ExchangeRateCurrencyRate::AVERAGE_RATE_TYPE].include?(type_param)
+        end
```
alexdesi, line 42: Good that you are checking the type here!
We could be more explicit if you like:

```
            type_param = filter_params[:type]

          return nill unless [ExchangeRateCurrencyRate::SCHEDULED_RATE_TYPE, ExchangeRateCurrencyRate::SPOT_RATE_TYPE, ExchangeRateCurrencyRate::AVERAGE_RATE_TYPE].include?(type_param)
          
          type_param
        end
```

app/models/exchange_rate_currency_rate.rb
```diff
@@ -3,6 +3,7 @@
 class ExchangeRateCurrencyRate < Sequel::Model
   SCHEDULED_RATE_TYPE = 'scheduled'.freeze
   SPOT_RATE_TYPE = 'spot'.freeze
+  AVERAGE_RATE_TYPE = 'average'.freeze
```
shjohnson, line 6: Not sure if this can be done but its a same we cant group these constants into `RATE_TYPE_CONSTANTS` in case more are added or we can pass that group of constant around.. probably a simple way but brain not working right now or its over engineering.

willfish (approved):

app/models/exchange_rate_currency_rate.rb
willfish: We could move to using an enum but I'd argue we can do that in a separate PR.

We have a couple of options, there:

1. Build a simple enum plugin which includes validations of whether the rate type value matches one of the available options and adds convenience methods `average?`
2. Add an enum type to the database and use that in the rate type implementation. This would be painful to alter/extend in future but would guarantee that we're meeting the contract we've laid out

PR #1511, opened 2023-09-27
HOTT-3266: Adds measure quota coverage worksheet

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4054

### What?

I have added/removed/altered:

- [x] Adds date formatters
- [x] Adds brexit date dataset method
- [x] Adds measure quota coverage worksheet

### Why?

I am doing this because:

- This is required as part of the differences and anomalies reporting

### DIFF
--- a/app/lib/reporting/differences.rb
+++ b/app/lib/reporting/differences.rb
@@ -61,6 +61,8 @@ module Reporting
       @regular_style = workbook.styles.add_style(
         alignment: {
           wrap_text: true,
+          horizontal: :left,
+          vertical: :top,
         },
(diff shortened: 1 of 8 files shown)


app/lib/reporting/differences.rb
```diff
@@ -270,6 +278,8 @@ def handle_csv(csv)
 
     class << self
       def generate(only: [])
+        only = [:add_measure_quota_coverage_worksheet]
```
willfish, line 281: I'll remove this when I've validated the report outputs are 100%

PR #1515, opened 2023-09-29
HOTT-3976: Fix the support for type average 

alexdesi:
### What?
Fix the support for for type `average` on the endpoint 'exchange_rates/period_lists'.

### Jira link
https://transformuk.atlassian.net/browse/HOTT-3976

### Why?
To simplify/unify FE exchange rate pages.

### DIFF
--- a/app/models/exchange_rates/period.rb
+++ b/app/models/exchange_rates/period.rb
@@ -2,12 +2,7 @@ module ExchangeRates
   class Period
     attr_accessor :month, :year, :files, :has_exchange_rates
 
-    def initialize(month:, year:, has_exchange_rates:, files:)
-      @month = month
-      @year = year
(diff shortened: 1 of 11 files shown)


app/models/exchange_rates/period.rb
```diff
@@ -1,6 +1,13 @@
 module ExchangeRates
   class Period
-    attr_accessor :month, :year, :files
+    attr_accessor :month, :year, :files, :has_exchange_rates
+
+    def initialize(month:, year:, has_exchange_rates:, files:)
+      @month = month
+      @year = year
+      @has_exchange_rates = has_exchange_rates
+      @file = files
```
willfish, line 9:
```suggestion
      @files = files
```

PR #1524, opened 2023-10-04
HOTT-4023: Populate day zero exchange_rate_countries_currencies

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4023

### What?

I have added/removed/altered:

- [x] Added day zero country currencies

### Why?

I am doing this because:

- This is needed to make sure we can surface exchange rates


db/data_migrations/20231004142615_populate_country_currencies.rb
```diff
@@ -0,0 +1,197 @@
...
+            "created_at" => now,
+            "updated_at" => now
+          )
+        )
+      end
+    end
+  end
+
+  down do
+    # Won't rollback
```
alexdesi, line 18: We still need to implement this, don't we?

alexdesi (approved):
great! ✅

PR #1525, opened 2023-10-06
HOTT-4023: Implement new exchange rate data model

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4023

### What?

I have added/removed/altered:

- [x] Adds joining and filtering behaviour to handle overlapping exchange rates
- [x] Updates all implementations to use the new country currency data model
- [x] Updates all specs to make sure these reflect the new model
- [x] Updates controller api serializers to use new model
- [x] Rejigged services to share a common monthly rate implementation
- [x] Update all spot rates to have an end date

### Why?

I am doing this because:

- This is required so that we can keep track of the history of exchange rate country/currency descriptions


alexdesi (approved):
lovely :green_heart:

PR #1529, opened 2023-10-10
Speed up BE tests.

alexdesi:
### What?
Speed up rollback_spec, rollback_worker_spec.
Those tests used to run in minutes. They run in seconds now.

### Jira link

https://transformuk.atlassian.net/browse/HOTT-3952

### Why?
Fast test == Happy developers, Happy QA testers, Higher productivity

### DIFF
--- a/app/lib/tariff_synchronizer.rb
+++ b/app/lib/tariff_synchronizer.rb
@@ -178,33 +178,37 @@ module TariffSynchronizer
         Sequel::Model.db.transaction do
           # Delete actual data
           oplog_based_models.each do |model|
-            model.operation_klass.where { operation_date > date_for_rollback }.delete
+            model.operation_klass.where(Sequel.lit('operation_date > ?', date_for_rollback)).delete
           end
 
           if keep
             # Rollback TARIC
-            TariffSynchronizer::TaricUpdate.applied_or_failed.where { issue_date > date_for_rollback }.each do |taric_update|
-              instrument('rollback_update.tariff_synchronizer',
-                         update_type: :taric,
-                         filename: taric_update.filename)
+            TariffSynchronizer::TaricUpdate.applied_or_failed
+              .where(Sequel.lit('issue_date > ?', date_for_rollback))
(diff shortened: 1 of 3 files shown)


app/lib/tariff_synchronizer/logger.rb
```diff
@@ -143,4 +143,4 @@ def rollback_update(event)
   end
 end
 
-TariffSynchronizer::Logger.attach_to :tariff_synchronizer
+# TariffSynchronizer::Logger.attach_to :tariff_synchronizer
```
alexdesi, line 146: remove this comment

app/lib/tariff_synchronizer.rb
```diff
@@ -178,33 +178,37 @@ def rollback(rollback_date, keep: false)
         Sequel::Model.db.transaction do
           # Delete actual data
           oplog_based_models.each do |model|
-            model.operation_klass.where { operation_date > date_for_rollback }.delete
+            model.operation_klass.where(Sequel.lit('operation_date > ?', date_for_rollback)).delete
           end
 
           if keep
             # Rollback TARIC
-            TariffSynchronizer::TaricUpdate.applied_or_failed.where { issue_date > date_for_rollback }.each do |taric_update|
```
alexdesi, line 186: In this case, Sequel used **virtual row blocks** syntax.
For example:

    Book.where{stars >= 3}
That is not efficient, since behind the scenes, Sequel uses method_missing is to instantiate identifiers and function objects.

Most of the time it does not matter,
but in this case, it is inside a loop, inside a loop, so it does matter (I've tested it).

alexdesi: Since there are other broken tests, I rolled back this improvement to exclude this as a possible cause.

alexdesi:
Note:  rspec test run in 2 min and 40 sec:

Finished in 2 minutes 40.2 seconds (files took 6.61 seconds to load)

willfish (approved):

PR #1532, opened 2023-10-12
BAU: Refactor exchange rates services

willfish:
### Jira link

BAU

### What?

The plan is to remove the create all monthly file service once we've ran this in all environments

I have added/removed/altered:

- [x] Refactored services to share rates and publication date and generally to have reduced responsibilities
- [x] Updated specs to reflect new api
- [x] Added specs for the worker
- [x] Added a service to recreate and upload all files

### Why?

I am doing this because:

- This is required so we can more simply regenerate all of the monthly files


alexdesi (approved):
:+1:

PR #1533, opened 2023-10-12
BAU: Removes old exchange rates services

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Removed old exchange rate services

### Why?

I am doing this because:

- These are no longer needed


alexdesi (approved):
:broom:

PR #1541, opened 2023-10-19
HOTT-3953: Fix synchronizer coupling between CDS and TARIC

rasikasri:
### Jira link

[HOTT-<TODO>](https://transformuk.atlassian.net/browse/HOTT-3953)

### What?

I have added/removed/altered:

- [ ] Refactored Tariff synchronizer to separate CDS and TARIC APIs

### Why?

I am doing this because:
- Having a one service for both CDS and Taric synchronizer doesn't comply with single responsibility principle. In this pr, I created two separate infrastructures for those two operation.
-
-

### Have you? (optional)


app/lib/cds_synchronizer.rb
```diff
@@ -0,0 +1,113 @@
...
+  self.cds_logger_enabled = (ENV['TARIFF_CDS_LOGGER'].to_i == 1)
+
+  # set initial update date
+  # Initial dump date + 1 day
+  cattr_accessor :initial_update_date
+  self.initial_update_date = Date.new(2020, 9, 1)
+
+  class << self
+    def download
+      return Rails.logger.error 'Missing: Tariff sync enviroment variables: TARIFF_SYNC_USERNAME, TARIFF_SYNC_PASSWORD, TARIFF_SYNC_HOST and TARIFF_SYNC_EMAIL.' unless sync_variables_set?
```
willfish, line 16: This is actually incorrect, the variables that are loaded for CDS are:

```
ENV['HMRC_API_HOST'].present? && ENV['HMRC_CLIENT_ID'].present? && ENV['HMRC_CLIENT_SECRET'].present?
```

app/lib/cds_synchronizer.rb
```diff
@@ -0,0 +1,113 @@
...
+    end
+
+    def apply
+      check_tariff_updates_failures
+      check_sequence
+
+      applied_updates = []
+      import_warnings = []
+
+      # The sync task is run on multiple machines to avoid more than on process
```
willfish, line 36:
```suggestion
      # The sync task is run on multiple machines to avoid more than one process
```

app/lib/cds_synchronizer.rb
```diff
@@ -0,0 +1,113 @@
...
+        applied_updates.flatten!
+
+        if applied_updates.any? && BaseUpdate.pending_or_failed.none?
+          TariffLogger.apply(applied_updates.map(&:filename))
+
+          true
+        end
+      end
+    rescue Redlock::LockError
+      TariffLogger.apply_lock_error
```
willfish, line 61: This still feels like an extra bit of misdirection just to log a line :thinking:

app/lib/cds_synchronizer.rb
```diff
@@ -0,0 +1,113 @@
...
+          end
+
+          updates.each do |cds_update|
+            cds_update.mark_as_pending
+            cds_update.clear_applied_at
+            cds_update.cds_errors_dataset.destroy
+            cds_update.delete unless keep
+          end
+
+          # Requeue data migrations
```
willfish, line 92:
```suggestion
          # Run missing data migrations at next deploy
```

app/lib/cds_synchronizer.rb
```diff
@@ -0,0 +1,113 @@
...
+          # Rollback leaves 'date_for_rollback's data intact, it removes only
```
willfish, line 93: I could do with @alexdesi @shjohnson reviewing these comments for comprehensibility

app/lib/cds_synchronizer.rb
```diff
@@ -0,0 +1,113 @@
...
+    rescue Redlock::LockError
+      Rails.logger.warn("Failed to acquire Redis lock for rollback to #{rollback_date}. Keep records: #{keep}")
+    end
+
+    def sync_variables_set?
+      ENV['HMRC_API_HOST'].present? && ENV['HMRC_CLIENT_ID'].present? && ENV['HMRC_CLIENT_SECRET'].present?
+    end
+
+    def downloaded_todays_file?
+      CdsUpdate.downloaded_todays_file?
```
willfish, line 110: This is just a delegator, I wonder if we can just delegate to the CdsUpdate constant:

```ruby
    # ... at the top of the module/class
    delegate :download_todays_file?, to: CdsUpdate
```

app/lib/taric_synchronizer.rb
```diff
@@ -0,0 +1,137 @@
...
+  class << self
+    # Download pending updates for TARIC and CDS data
+    # Gets latest downloaded file present in (inbox/failbox/processed) and tries
+    # to download any further updates to current day.
+    def download
+      return Rails.logger.error 'Missing: Tariff sync enviroment variables: TARIFF_SYNC_USERNAME, TARIFF_SYNC_PASSWORD, TARIFF_SYNC_HOST and TARIFF_SYNC_EMAIL.' unless sync_variables_set?
+
+      TradeTariffBackend.with_redis_lock do
+        begin
+          TradeTariffBackend.patch_broken_taric_downloads? ? TariffSynchronizer::TaricUpdate.sync_patched : TariffSynchronizer::TaricUpdate.sync(initial_date: initial_update_date)
```
willfish, line 40: We can kill the non-patched path, here.

They never fixed the api enough so that we can rely on the non-patched call _and_ they're migrating over to just dumping all of the downloader code in its current form for Taric code by the end of December

app/lib/taric_synchronizer.rb
```diff
@@ -0,0 +1,137 @@
...
+    end
+
+    def apply
+      check_tariff_updates_failures
+      check_sequence
+
+      applied_updates = []
+      import_warnings = []
+
+      # The sync task is run on multiple machines to avoid more than on process
```
willfish, line 56:
```suggestion
      # The sync task is run on multiple machines to avoid more than one process
```

app/lib/tariff_synchronizer/mailer.rb
```diff
@@ -58,10 +57,9 @@ def applied(update_names, import_warnings)
       @update_names = update_names
       @import_warnings = import_warnings
       # if 'presence errors' are ignored during tariff update then we can display them in email body
-      if TariffSynchronizer.ignore_presence_errors
+      if TaricSynchronizer.ignore_presence_errors
```
willfish, line 60: Curious, I suspect we can actually delete the presence error feature

willfish: We've not used it in the three years I've been here

app/workers/taric_updates_synchronizer_worker.rb
```diff
@@ -0,0 +1,41 @@
+class TaricUpdatesSynchronizerWorker
+  include Sidekiq::Worker
+
+  TRY_AGAIN_IN = 20.minutes
+  CUT_OFF_TIME = '10:00'.freeze
+
+  sidekiq_options queue: :sync, retry: false
+
+  def perform(_check_for_todays_file = true, reapply_data_migrations = false)
```
willfish, line 9:
```suggestion
  def perform(reapply_data_migrations = false)
```

spec/integration/tariff_synchronizer/rollback_cds_spec.rb
```diff
@@ -1,5 +1,5 @@
-RSpec.describe TariffSynchronizer, '.rollback_cds' do
-  subject(:rollback_cds) { described_class.rollback_cds('2021-01-01') }
+RSpec.describe CdsSynchronizer, '.rollback' do
```
willfish, line 1: These filenames aren't quite right.

We'll want to rename these to be `spec/integration/cds_synchronizer/rollback_spec.rb`

spec/support/shared_examples/base_update_shared_examples.rb
```diff
@@ -7,12 +7,14 @@
     end
 
     context 'when last update is out of date' do
-      let!(:example_taric_update) { create :taric_update, example_date: Time.zone.yesterday }
+      # let!(:example_taric_update) { create :taric_update, example_date: Time.zone.yesterday }
```
willfish, line 10: Remove me

spec/unit/cds_synchronizer_spec.rb
```diff
@@ -0,0 +1,184 @@
+# rubocop:disable RSpec/MultipleExpectations
```
willfish, line 1: We don't need this with the :aggregate_failures switch on the context/example with multiple expectations

willfish: Generally we shouldn't have these file-specific overrides of rubocop rules

spec/unit/taric_importer/record_processor/destroy_operation_spec.rb
```diff
@@ -94,17 +128,21 @@
                                      'description' => 'French')
         end
 
-        it 'returns a model record' do
+        it 'returns a model record of' do
```
willfish, line 131:
```suggestion
        it 'returns a model record' do
```

spec/unit/taric_importer/record_processor/destroy_operation_spec.rb
```diff
@@ -116,26 +154,32 @@
         expect { operation.call }.to change(LanguageDescription, :count).from(1).to(0)
       end
 
-      it 'returns the model record' do
+      it 'returns the model record of' do
```
willfish, line 157:
```suggestion
      it 'returns the model record' do
```

spec/unit/taric_importer_spec.rb
```diff
@@ -1,5 +1,4 @@
 # rubocop:disable RSpec/MultipleExpectations
-# rubocop:disable RSpec/InstanceVariable
```
willfish, line 2: :raised_hands:

spec/unit/tariff_synchronizer/base_update_spec.rb
```diff
@@ -110,26 +120,29 @@
...
+        expect(TariffSynchronizer::CdsUpdateDownloader).to have_received(:new).with(download_date)
+      end
     end
   end
 
   describe '#last_updates_are_missing?' do
     context 'with weekends' do
       before do
         travel_to Date.parse('21-05-2017')
+        @update1 = create(:taric_update, :missing, example_date: Time.zone.today)
```
willfish, line 138: update1 and update2 aren't meaningfully named:

```ruby
missing_update = create :taric_update, :missing, example_date: Time.zone.today
pending_update = create :taric_update, example_date: Time.zone.yesterday
```

Also, instance variables in tests are frowned on generally https://www.betterspecs.org/

willfish: Missing updates are not actually possible any more so we can actually probably just remove any tests of missing updates

spec/unit/tariff_synchronizer/cds_update_spec.rb
```diff
@@ -51,29 +53,27 @@
...
-      tariff_synchronizer_logger_listener
-      allow_any_instance_of(CdsImporter).to \
-        receive(:import).and_return inserted_oplog_records
+      allow(Rails.logger).to receive(:info)
+
       cds_update.import!
-      expect(@logger.logged(:info).size).to eq 1
-      expect(@logger.logged(:info).last).to match(/Applied CDS update/)
+
+      expect(Rails.logger).to have_received(:info)
```
willfish, line 76: I'd just remove info log tests. They add no value whatsoever and bog down our suite

spec/unit/tariff_synchronizer/logger_spec.rb
```diff
@@ -1,40 +1,44 @@
...
 
     it 'logs a warn event' do
-      TariffSynchronizer.apply
+      TaricSynchronizer.apply
+      expect(Rails.logger).to have_received(:warn)
+    end
 
-      expect(@logger.logged(:warn).size).to be >= 1
-      expect(@logger.logged(:warn).first.to_s).to match(/acquire Redis lock/)
+    it 'logs warn event message' do
```
willfish, line 39: I'd just delete this

spec/workers/cds_updates_synchronizer_worker_spec.rb
```diff
@@ -29,51 +29,9 @@
...
-        context 'with reapply_data_migrations option' do
-          subject(:perform) { described_class.new.perform(true, true) }
-
-          it { expect(TariffSynchronizer).to have_received(:download) }
-          it { expect(DataMigrator).not_to have_received(:migrate_up!) }
-        end
-      end
-    end
-
     context 'when on the uk service' do
```
willfish, line 32: I think we can delete this surrounding context

spec/workers/cds_updates_synchronizer_worker_spec.rb
```diff
@@ -122,11 +80,11 @@
...
           let(:cut_off_time) { 5.minutes.ago }
 
-          it { expect(TariffSynchronizer).to have_received(:download_cds) }
-          it { expect(TariffSynchronizer).to have_received(:apply_cds) }
+          it { expect(CdsSynchronizer).to have_received(:download) }
+          it { expect(CdsSynchronizer).to have_received(:apply) }
 
-          it { expect(TariffSynchronizer).not_to have_received(:download) }
-          it { expect(TariffSynchronizer).not_to have_received(:apply) }
+          it { expect(TaricSynchronizer).not_to have_received(:download) }
```
willfish, line 86: I don't think these tests are relevant anymore since this is the Cds synchronizer

willfish:
It would be good if we could add a rationale in the pull request template for future developers to know why we've done this.

I'm wondering what manually testing you've done?

I'd run sidekiq and see if the schedules for the workers look right:

```bundle exec sidekiq```

For xi/uk services, I'd manually run:

- a rollback (both in the admin UI and in a ruby console)
- a download
- an apply

For the download step I'd make sure that we're actually physically downloading a file (we can delete one of the files in the s3 bucket)

File: app/lib/taric_synchronizer.rb (hunk 7 above)
Comment (rasikasri) [reply app/lib/taric_synchronizer.rb:40]: This will be done in a separate ticket

File: spec/workers/cds_updates_synchronizer_worker_spec.rb (hunk 21 above)
Comment (rasikasri) [reply spec/workers/cds_updates_synchronizer_worker_spec.rb:86]: It is basically testing taric synchronizer doesn't get called by cds update worker. So I think it is relevent negative test.

File: spec/unit/cds_synchronizer_spec.rb (hunk 13 above)
Comment (rasikasri) [reply spec/unit/cds_synchronizer_spec.rb]: I didn't find a way to do bellow test without multiple expectation in the same example.

```
context 'when pending CDS update does not respect the sequence' do
      let(:pending_date) { applied_date + 2.days }

      it 'raises wrong sequence error and notifies Slack app' do
        allow(SlackNotifierService).to receive(:call)

        expect { described_class.apply }.to raise_error(TariffSynchronizer::FailedUpdatesError)
        # allow(described_class).to receive(:apply).and_raise(TariffSynchronizer::FailedUpdatesError)

        described_class.apply

        expect(SlackNotifierService).to have_received(:call)
      end
    end
```

Tried with commented stub, but it doesn't work in this test, any few other similar tests in same spec. This might be the reason why ovverrides added originally.

File: app/lib/tariff_synchronizer/mailer.rb (hunk 9 above)
Comment (rasikasri) [reply app/lib/tariff_synchronizer/mailer.rb:60]: I'd prefer to do this in separate ticket

File: app/lib/cds_synchronizer.rb (hunk 5 above)
Comment (shjohnson) [reply app/lib/cds_synchronizer.rb:95]: I think extra description after `it removes only..` like `removes only updates after the rollback date` or something?

Comment (shjohnson) [reply app/lib/cds_synchronizer.rb:95]: Or maybe thats what you meant below..

Comment (shjohnson) [reply app/lib/cds_synchronizer.rb:95]: This also feels like a description for the method so could be moved to just under the definition then smaller comments for the methods describing the commands?

File: app/lib/taric_synchronizer.rb [hunk 22]
```diff
@@ -0,0 +1,137 @@
+class TaricSynchronizer
+  extend TariffSynchronizer
+
+  # 1 - does not raise an exception when record does not exist on TARIC DESTROY operation
+  #   - does not raise an exception when record does not exist on TARIC UPDATE operation
+  #   - creates new record when record does not exist on TARIC UPDATE operation
+  cattr_accessor :ignore_presence_errors
```
shjohnson, line 7: I know these where inherited from the previous code but not a massive fan seeing as these are not thread safe although not sure how much an issue this considering they are more like setup code when the process is run right?

rasikasri: These variables are defined under self, so I assume  they are in singleton context and would not changed unless you run the application again. So I don't think there is a thread issue here. Please make me correct if I am wrong.

willfish: Just to weigh in here, technically its not possible for these to be read by two concurrent processes at the same time since there is effectively a redis lock/mutex in place that restricts work happening concurrently.

PR #1550, opened 2023-10-24
Create twice-yearly exchange rate AVERAGE rate data: required in CSV and 'view online'

shjohnson:
### Jira link

HOTT-3401

### What?

I have added/removed/altered:

- [ ] Created method for returning the live currency_codes for countries
- [ ] Created method for returning the live countries for last year
- [ ] Created method for returning by currency
- [ ] Method for returning monthly rates by currency last year
- [ ] Average rates service creating and instructing the file upload 
- [ ] Create CSV of the average rates 
- [ ] Orchestration of the file upload
- [ ] Changes file name of the monthly file service to just file service
- [ ] Ensured the monthly exchange rates service has a single method call
- [ ] Added extra factories for testing
- [ ] Orchestration for when the job is kicked off and worker with test

See screenshots taken on local env after running `ExchangeRates::AverageExchangeRatesService.call(force_run: true, selected_date: '2023-10-25')`
 
Screenshot 2023-10-25 at 12 27 53

Screenshot 2023-10-25 at 12 27 24

### DIFF
--- a/app/models/exchange_rate_country_currency.rb
+++ b/app/models/exchange_rate_country_currency.rb
@@ -16,8 +16,7 @@ class ExchangeRateCountryCurrency < Sequel::Model(:exchange_rate_countries_curre
 
   def self.live_currency_codes
     # Criteria:
-    # End date within the current month
-    # Start date less than or equal to the end of the current month
+    # Countries that are live in the current month
 
     where(
       (Sequel[:validity_end_date] =~ (Time.zone.today.beginning_of_month..Time.zone.today.end_of_month)) |
@@ -28,8 +27,8 @@ class ExchangeRateCountryCurrency < Sequel::Model(:exchange_rate_countries_curre
 
   def self.live_last_twelve_months(date)
     # Criteria:
-    # Start date less than then end of the current month and end date nil or more than
-    # End date more than then beginning of the current month 12 months ago
+    # Countries that have had a live rate in the last 12 months whole months
+    # So end of the current month minus 12 months
 
     where(
       (Sequel[:validity_start_date] <= date.end_of_month) &
--- a/app/models/exchange_rate_country_currency.rb
+++ b/app/models/exchange_rate_country_currency.rb
@@ -15,6 +15,27 @@ class ExchangeRateCountryCurrency < Sequel::Model(:exchange_rate_countries_curre
   end
 
   def self.live_currency_codes
-    distinct(:currency_code).select_map(:currency_code)
+    # Criteria:
+    # End date within the current month
+    # Start date less than or equal to the end of the current month
+
+    where(
+      (Sequel[:validity_end_date] =~ (Time.zone.today.beginning_of_month..Time.zone.today.end_of_month)) |
+      (Sequel[:validity_end_date] =~ nil) &
+      (Sequel[:validity_start_date] <= Time.zone.today.end_of_month),
+    ).distinct(:currency_code).select_map(:currency_code)
+  end
+
+  def self.live_last_twelve_months
+    # Criteria:
+    # Start date less than then end of the current month and end date nil or more than
+    # End date more than then beginning of the current month 12 months ago
+
+    where(
+      (Sequel[:validity_start_date] <= Time.zone.today.end_of_month) &
+      (Sequel[:validity_end_date] =~ nil) |
+      (Sequel[:validity_start_date] <= Time.zone.today.end_of_month) &
+      (Sequel[:validity_end_date] >= Time.zone.today.beginning_of_month - 11.months),
+    ).order_by(:country_description)
   end
 end
--- a/app/models/exchange_rate_country_currency.rb
+++ b/app/models/exchange_rate_country_currency.rb
@@ -13,4 +13,8 @@ class ExchangeRateCountryCurrency < Sequel::Model(:exchange_rate_countries_curre
     validates_presence :currency_code
     validates_presence :validity_start_date
   end
+
+  def self.live_currency_codes
+    distinct(:currency_code).select_map(:currency_code)
+  end
 end
--- a/app/models/exchange_rate_country_currency.rb
+++ b/app/models/exchange_rate_country_currency.rb
@@ -26,16 +26,16 @@ class ExchangeRateCountryCurrency < Sequel::Model(:exchange_rate_countries_curre
     ).distinct(:currency_code).select_map(:currency_code)
   end
 
-  def self.live_last_twelve_months
+  def self.live_last_twelve_months(date)
     # Criteria:
     # Start date less than then end of the current month and end date nil or more than
     # End date more than then beginning of the current month 12 months ago
 
     where(
-      (Sequel[:validity_start_date] <= Time.zone.today.end_of_month) &
+      (Sequel[:validity_start_date] <= date.end_of_month) &
       (Sequel[:validity_end_date] =~ nil) |
-      (Sequel[:validity_start_date] <= Time.zone.today.end_of_month) &
-      (Sequel[:validity_end_date] >= Time.zone.today.beginning_of_month - 11.months),
+      (Sequel[:validity_start_date] <= date.end_of_month) &
+      (Sequel[:validity_end_date] >= date.beginning_of_month - 11.months),
     ).order_by(:country_description)
   end
 end
--- a/app/services/exchange_rates/average_exchange_rates_service.rb
+++ b/app/services/exchange_rates/average_exchange_rates_service.rb
@@ -1,0 +1,67 @@
+module ExchangeRates
+  class AverageExchangeRatesService
+    VALID_MONTHS = [4, 12].freeze
+    VALID_DAY = 31
+
+    class << self
+      def call
+        return unless valid_date
+
+        avg_rates = create_average_rates
+        log_to_slack
+        build_average_rate_file(avg_rates)
+      end
+
+      private
+
+      def build_average_rate_file(avg_rates)
+        ExchangeRates::UploadFileService.new(
+          avg_rates,
+          Time.zone.today.iso8601,
+          :average_csv,
+        ).call
+      end
+
+      def valid_date
+        return false unless Time.zone.today.day == VALID_DAY
+        return false unless VALID_MONTHS.include?(Time.zone.today.month)
+      end
+
+      def create_average_rates
+        currency_codes = ExchangeRateCountryCurrency.live_currency_codes
+
+        currency_codes.map do |currency_code|
+          next unless valid_average_rate(currency_code)
+
+          avg_rate = valid_average_rate(currency_code)
+
+          ExchangeRateCurrencyRate.create(
+            currency_code:,
+            validity_start_date: Time.zone.today.beginning_of_month,
+            validity_end_date: nil,
+            rate_type: ExchangeRateCurrencyRate::AVERAGE_RATE_TYPE,
+            rate: avg_rate,
+          )
+        end
+      end
+
+      def valid_average_rate(currency_code)
+        rates = ExchangeRateCurrencyRate.by_currency_and_last_year(currency_code)
+
+        return unless rates.count >= 12
+        return unless rates.first.validity_start_date.month == Time.zone.today.month
+        return unless VALID_MONTHS.include?(rates.first.validity_start_date.month)
+
+        rates.pluck(:rate).sum.fdiv(rates.size)
+      end
+
+      def log_to_slack
+        message = 'Average exchange rates for the current period have been added and are accessible for viewing at /exchange_rates.'
+
+        logger.info message
+
+        SlackNotifierService.call(message)
+      end
+    end
+  end
+end
--- a/app/services/exchange_rates/average_exchange_rates_service.rb
+++ b/app/services/exchange_rates/average_exchange_rates_service.rb
@@ -3,11 +3,16 @@ module ExchangeRates
     VALID_MONTHS = [4, 12].freeze
     VALID_DAY = 31
 
-    def self.call(force_run)
-      new.call(force_run)
+    def self.call(force_run:, selected_date:)
+      new(force_run, selected_date).call
     end
 
-    def call(force_run)
+    def initialize(force_run, selected_date)
+      @force_run = force_run
+      @selected_date = Date.parse(selected_date)
+    end
+
+    def call
       return argument_error unless valid_date || force_run
 
       avg_rates = create_average_rates
@@ -17,6 +22,8 @@ module ExchangeRates
     end
 
     private
+
+    attr_reader :force_run, :selected_date
 
     def match_country_with_rates(avg_rates)
       live_countries_last_twelve_months.each_with_object({}) do |country, h|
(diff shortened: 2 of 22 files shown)


spec/services/exchange_rates/average_exchange_rates_service_spec.rb
```diff
@@ -0,0 +1,355 @@
...
+
+      context 'with force run enabled' do
+        let(:force_run) { true }
+
+        it_behaves_like 'creating the average rates'
+      end
+    end
+  end
+
+  def setup_data
```
shjohnson, line 79: I know lots of setup data but this is to ensure all test cases are covered, it could be done with a massive sql insert query but i think this is clearer in what objects are being created

shjohnson: Plus this is like an integration test for ensuring all the correct data is created

app/models/exchange_rate_country_currency.rb
```diff
@@ -13,4 +13,29 @@ def validate
...
+      (Sequel[:validity_end_date] =~ (Time.zone.today.beginning_of_month..Time.zone.today.end_of_month)) |
+      (Sequel[:validity_end_date] =~ nil) &
+      (Sequel[:validity_start_date] <= Time.zone.today.end_of_month),
+    ).distinct(:currency_code).select_map(:currency_code)
+  end
+
+  def self.live_last_twelve_months(date)
+    # Criteria:
+    # Start date less than then end of the current month and end date nil or more than
+    # End date more than then beginning of the current month 12 months ago
```
alexdesi, line 32: Maybe we could make this phrase more readable. or, if we can't,  even remove it since it does not add much to the code?

app/services/exchange_rates/average_exchange_rates_service.rb
```diff
@@ -0,0 +1,97 @@
+module ExchangeRates
+  class AverageExchangeRatesService
+    VALID_MONTHS = [4, 12].freeze
+    VALID_DAY = 31
+
+    def self.call(force_run:, selected_date:)
```
alexdesi, line 6: it's probably good to have
`force_run: false` here,
just to be safe.

app/services/exchange_rates/average_exchange_rates_service.rb
```diff
@@ -0,0 +1,97 @@
...
+      return unless rates.last.validity_start_date <= selected_date
+
+      # Ensure the first month is not greater than 12 months before selected date
+      return unless rates.first.validity_end_date > selected_date - 12.months
+
+      rates.pluck(:rate).sum.fdiv(rates.count)
+    end
+
+    def argument_error
+      error_message = 'Argument error, invalid date, average exchange rate creation'
```
alexdesi, line 90: A better error message should provide the date that is not valid.
something like:
"Argument error:  #{selected_date} is not a valid date"

Note: I did not  test it

app/services/exchange_rates/average_exchange_rates_service.rb
```diff
@@ -0,0 +1,97 @@
+module ExchangeRates
+  class AverageExchangeRatesService
```
alexdesi, line 2: I was wondering if we could find a more explicit name for the name of the service.
I guess you are following the convention: for example `spot_exchange_rates_services`
What does this actually do?
create the average and upload the file?
maybe we can add a comment? :thinking:

app/services/exchange_rates/average_exchange_rates_service.rb
```diff
@@ -0,0 +1,97 @@
...
+
+    def match_country_with_rates(avg_rates)
+      live_countries_last_twelve_months.each_with_object({}) do |country, h|
+        # Need to ensure unique rates are being collected
+        avg_rate = avg_rates
+                    .select { |rate| rate.currency_code == country.currency_code }
+                    .first
+                    .rate
+
+        # { ExchangeRateCountryCurrency => avg_rate }
```
alexdesi, line 36: do we need this comment?

app/services/exchange_rates/average_exchange_rates_service.rb
```diff
@@ -0,0 +1,97 @@
...
+    private
+
+    attr_reader :force_run, :selected_date
+
+    def match_country_with_rates(avg_rates)
+      live_countries_last_twelve_months.each_with_object({}) do |country, h|
+        # Need to ensure unique rates are being collected
+        avg_rate = avg_rates
+                    .select { |rate| rate.currency_code == country.currency_code }
+                    .first
```
alexdesi, line 33: Is not `select ... first...` equal to the `find` method? :thinking:

app/services/exchange_rates/average_exchange_rates_service.rb
```diff
@@ -0,0 +1,97 @@
...
+      new(force_run, selected_date).call
+    end
+
+    def initialize(force_run, selected_date)
+      @force_run = force_run
+      @selected_date = Date.parse(selected_date)
+    end
+
+    def call
+      return argument_error unless valid_date || force_run
```
alexdesi, line 16: if `force_run == true` you get an argument error?
this does not sound right, but maybe I am not considering something

app/services/exchange_rates/average_exchange_rates_service.rb
```diff
@@ -0,0 +1,97 @@
...
+        selected_date,
+        :average_csv,
+      ).call
+    end
+
+    def valid_date
+      return false unless selected_date.day == VALID_DAY
+      return false unless VALID_MONTHS.include?(selected_date.month)
+
+      true
```
alexdesi, line 53: In case we want to use a  shorter version :smile: 
```
selected_date.day == VALID_DAY &&
  VALID_MONTHS.include?(selected_date.month)
```

app/services/exchange_rates/average_exchange_rates_service.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    def create_average_rates
+      live_countries_last_twelve_months.select_map(:currency_code).uniq.map do |currency_code|
+        avg_rate = valid_average_rate(currency_code)
+
+        next unless avg_rate
```
alexdesi, line 60: We should avoid using a next inside a map.
I'd suggest to use reject block before, to filter out the elements you want to exclude. 
```
something.reject{ |currency_code| valid_average_rate(currency_code).nil?}
```

app/services/exchange_rates/average_exchange_rates_service.rb
```diff
@@ -0,0 +1,97 @@
...
+        )
+      end
+    end
+
+    def live_countries_last_twelve_months
+      # This needs to be all countries that have had a rate for a currency last 12 months
+      @live_countries_last_twelve_months ||= ExchangeRateCountryCurrency.live_last_twelve_months(selected_date)
+    end
+
+    def valid_average_rate(currency_code)
```
alexdesi, line 77: idea: it would be nice to have a method called
`valid_average_rate?(...)`
that you can also use to select the valid ones in `create_average_rates` :

```
something.select{ |currency_code| valid_average_rate?(currency_code)}
```

app/workers/average_exchange_rates_worker.rb
```diff
@@ -0,0 +1,19 @@
...
+  sidekiq_options retry: 1, retry_in: 1.hour
+
+  def perform
+    ExchangeRates::AverageExchangeRatesService.call(force_run: false, selected_date: Time.zone.today.iso8601)
+
+    notify
+  end
+
+  def notify
+    message = 'Average exchange rates for the current period have been added and are accessible for viewing at /exchange_rates.'
```
alexdesi, line 13: Use active form, instead of passive, when you can: 
'Average exchange rates for the current period are ready. You can view them at  /exchange_rates.'

See: https://www.grammarly.com/blog/active-vs-passive-voice/

spec/models/exchange_rate_currency_rate_spec.rb
```diff
@@ -345,6 +345,69 @@
...
+      let(:currency_code) { 'usd' }
+
+      it { expect(dataset.pluck(:currency_code)).to eq(%w[USD]) }
+    end
+
+    context 'when currency code is upper case' do
+      let(:currency_code) { 'USD' }
+
+      it { expect(dataset.pluck(:currency_code)).to eq(%w[USD]) }
+    end
```
alexdesi, line 366: Maybe this is redundant, up to you.

spec/models/exchange_rate_currency_rate_spec.rb
```diff
@@ -345,6 +345,69 @@
...
+             validity_end_date: today.end_of_month)
+      create(:exchange_rate_currency_rate, :monthly_rate, :with_usa,
+             validity_start_date: today.beginning_of_month - 11.months,
+             validity_end_date: today.end_of_month - 11.months)
+      create(:exchange_rate_currency_rate, :monthly_rate, :with_usa,
+             validity_start_date: today.end_of_month - 12.months,
+             validity_end_date: today.end_of_month - 12.months)
+      create(:exchange_rate_currency_rate, :monthly_rate, :with_eur,
+             validity_start_date: today.beginning_of_month,
+             validity_end_date: today.end_of_month)
```
alexdesi, line 389: I cannot say if we need all of them. Let's check if we need them all.

spec/models/exchange_rate_currency_rate_spec.rb
```diff
@@ -345,6 +345,69 @@
...
+      end
+    end
+
+    context 'when currency code is upper case' do
+      let(:currency_code) { 'USD' }
+
+      it 'will return the correct results', :aggregate_failures do
+        expect(dataset.count).to eq(2)
+        expect(dataset.pluck(:currency_code).uniq).to eq(%w[USD])
+      end
```
alexdesi, line 407: maybe redundant?

spec/services/exchange_rates/average_exchange_rates_service_spec.rb
alexdesi: yeah, up to you Sam, have a look if we can use just the bare minimum factories,
the more we use, the slower are the tests.

app/services/exchange_rates/average_exchange_rates_service.rb
shjohnson: Confused about this comment?

app/services/exchange_rates/average_exchange_rates_service.rb
shjohnson: This was just to show what the data structure is tbh

app/services/exchange_rates/average_exchange_rates_service.rb
shjohnson: will think about this one

spec/models/exchange_rate_currency_rate_spec.rb
shjohnson: Possibly but it was put in place as the code provides giving lower upper case options?

spec/services/exchange_rates/average_exchange_rates_service_spec.rb
```diff
@@ -0,0 +1,357 @@
...
+        avg_rates = ExchangeRateCurrencyRate.by_type(ExchangeRateCurrencyRate::AVERAGE_RATE_TYPE)
+
+        expect(avg_rates.all? { |r| r.validity_start_date == Time.zone.today.beginning_of_month }).to be true
+        expect(avg_rates.all? { |r| r.validity_end_date == Time.zone.today.end_of_month }).to be true
+
+        expect(avg_rates.by_currency('USD').first.rate).to eq(1.2264056116916666)
+        expect(avg_rates.by_currency('EUR').first.rate).to eq(1.1491571170166666)
+        expect(avg_rates.by_currency('AUD').first.rate).to eq(1.8997111260800001)
+        expect(avg_rates.by_currency('AED').first.rate).to eq(4.3747)
+        expect(avg_rates.by_currency('KZT').first.rate).to eq(545.0922)
```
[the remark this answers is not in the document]
shjohnson: I actually only took realistic values and like 95% of them are actual rates used.. Not sure it matters either way if they are high precision or not?

spec/models/exchange_rate_currency_rate_spec.rb
shjohnson: same as above

app/models/exchange_rate_country_currency.rb
```diff
@@ -13,4 +13,29 @@ def validate
     validates_presence :currency_code
     validates_presence :validity_start_date
   end
+
+  def self.live_currency_codes
```
willfish, line 17: Is this actually used?

shjohnson: Technically not since last iteration but kept it in as could be useful but appreciate could be taken out

app/services/exchange_rates/average_exchange_rates_service.rb
shjohnson: Let me know if this comment is ok and adds anything, i only liked it because it shows the structure of the hash better, agreed it was a little more complex when i first did it but happy either way

app/services/exchange_rates/average_exchange_rates_service.rb
shjohnson: Oh i think what your saying is because valid_date will be something but will be `false` or `true`

There is a test for this

app/services/exchange_rates/average_exchange_rates_service.rb
shjohnson: Think i can change the method name maybe to make it clearer what im trying to do

app/services/exchange_rates/average_exchange_rates_service.rb
shjohnson: I did try to implement something like this but forgot to say its an array rather than a collection so find wont work hence using select

app/services/exchange_rates/average_exchange_rates_service.rb
shjohnson: I changed the logic in the end, reject wouldnt work as thats on the a different element so hopefully it works better now

app/services/exchange_rates/create_average_exchange_rates_service.rb
```diff
@@ -0,0 +1,93 @@
...
+      selected_date.day == VALID_DAY && VALID_MONTHS.include?(selected_date.month)
+    end
+
+    def create_average_rates
+      live_countries_last_twelve_months.select_map(:currency_code).uniq.map do |currency_code|
+        rates = ExchangeRateCurrencyRate.monthly_by_currency_last_year(currency_code, selected_date)
+
+        next unless rates_valid?(rates)
+
+        rates.pluck(:rate).sum.fdiv(rates.count)
```
maverickvi, line 58: you have this line in 2 places

app/services/exchange_rates/create_average_exchange_rates_service.rb
```diff
@@ -0,0 +1,93 @@
...
+
+        ExchangeRateCurrencyRate.create(
+          currency_code:,
+          validity_start_date: selected_date.beginning_of_month,
+          validity_end_date: selected_date.end_of_month,
+          rate_type: ExchangeRateCurrencyRate::AVERAGE_RATE_TYPE,
+          rate: rates.pluck(:rate).sum.fdiv(rates.count),
```
maverickvi, line 65: you have this line in 2 places

shjohnson: good spot this was a refactor 🙌

alexdesi (approved):
:white_check_mark:

PR #1555, opened 2023-10-27
HOTT-4261: Adds seasonal measures worksheet

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4261

### What?

I have added/removed/altered:

- [x] Added seasonal measures worksheet
- [x] Fixed some presentational issues with quota coverage and overview worksheets

### Why?

I am doing this because:

- This is require for feature completeness


app/lib/reporting/differences/seasonal.rb
```diff
@@ -0,0 +1,220 @@
...
+        acc = []
+
+        each_seasonal_measure do |measure|
+          row = build_row_for(measure)
+
+          acc << row unless row.nil?
+        end
+
+        acc
+      end
```
alexdesi, line 94: a more compact for would be:
```
each_seasonal_measure.map{|m| build_row_for(m)}
                                            .compact
```

app/lib/reporting/differences/seasonal.rb
```diff
@@ -0,0 +1,220 @@
...
+
+          yield seasonal_measure if seasonal_measure.failed_duty_status.present?
+        end
+      end
+
+      def applicable_measures
+        @applicable_measures ||= begin
+          measures = Measure
+          .with_seasonal_measures(measure_type_ids, geographical_area_ids)
+          .all
```
alexdesi, line 125: indentation does not seem right here

alexdesi (dismissed):
It looks good, minor comments.

app/lib/reporting/differences/seasonal.rb
willfish: I think the terser version is harder to read, personally

willfish: I agree actually, I'll change this

PR #1571, opened 2023-11-07
HOTT-4323: Fixes candidate supplementary units

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4323

### What?

I have added/removed/altered:

- [x] Fixes a bunch of measurement units in the json file
- [x] Adds support for working out which units are coerced
- [x] Adds support for working out the type of a given full unit code
- [x] Fix logic in supplementary unit candidate worksheet

### Why?

I am doing this because:

- This is required to unblock the delivery of the differences report


alexdesi (approved):
it looks good :rocket:

PR #1573, opened 2023-11-07
BAU: Adds 8080 to signon links

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Added missing 8080 on signon internal url

### Why?

I am doing this because:

- This is what the container exposes

### DIFF
--- a/terraform/locals.tf
+++ b/terraform/locals.tf
@@ -51,7 +51,7 @@ locals {
     },
     {
       name  = "PLEK_SERVICE_SIGNON_URI"
-      value = "http://signon.tariff.internal"
+      value = "http://signon.tariff.internal:8080"
     },
     {
       name  = "RACK_TIMEOUT_SERVICE"


terraform/locals.tf
```diff
@@ -51,7 +51,7 @@ locals {
     },
     {
       name  = "PLEK_SERVICE_SIGNON_URI"
-      value = "http://signon.tariff.internal"
+      value = "http://signon.tariff.internal:8080"
```
amberstarlight, line 54:
```suggestion
      value = "http://signon.tariff.internal:8080/"
```

amberstarlight: Do we need to append this or does the routing deal with this?

willfish: I'll test it in a minute but I think it should be fine as long as it routes to the app

amberstarlight (approved):

PR #1574, opened 2023-11-08
HOTT-2796: Enable reporting in the new world

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2796

### What?

I have added/removed/altered:

- [x] Removes the env var used to toggle whether reporting is enabled
- [x] Adds the configuration required so that we can enable reporting in all environments

### Why?

I am doing this because:

- This is required to avoid special casing features only in production - we can run reports in development or staging and verify they integrate cleanly now


amberstarlight (approved):
Looks good to me! 🐑 🇮🇹

PR #1578, opened 2023-11-13
HOTT-4377: Fix bug for the endpoint /api/v2/exchange_rates/files/monthly_xml_2023-<??>.xml

alexdesi:
### What?
Fix bug for the endpoint:
/api/v2/exchange_rates/files/monthly_xml_2023-<??>.xml 

endpoint, which happens only for the months "08" and "09". The reason for the bug is that 08 and 09 are interpreted as not valid octal numbers.

### Jira link
https://transformuk.atlassian.net/browse/HOTT-4350

### Why?
It is a bug, 
To reproduce it:

This works:
https://www.trade-tariff.service.gov.uk/api/v2/exchange_rates/files/monthly_xml_2023-08.xml

These do not work:
https://www.trade-tariff.service.gov.uk/api/v2/exchange_rates/files/monthly_xml_2023-08.xml
https://www.trade-tariff.service.gov.uk/api/v2/exchange_rates/files/monthly_xml_2023-09.xml

But they should work!

### DIFF
--- a/app/models/exchange_rate_file.rb
+++ b/app/models/exchange_rate_file.rb
@@ -51,7 +51,8 @@ class ExchangeRateFile < Sequel::Model
     end
 
     def filename_for_download(type, format, year, month)
-      month_with_zero = sprintf('%02i', month)
+      # `to_i` is vital, since a number starting with 0 is interpreted as octal.
+      month_with_zero = sprintf('%02i', month.to_i(10))
 
       case type
       when 'monthly_csv_hmrc'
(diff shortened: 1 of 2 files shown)


app/models/exchange_rate_file.rb
```diff
@@ -51,7 +51,8 @@ def filename_for(type, format, year, month)
     end
 
     def filename_for_download(type, format, year, month)
-      month_with_zero = sprintf('%02i', month)
+      # `to_i` is vital, since a number starting with 0 is interpreted as octal.
+      month_with_zero = sprintf('%02i', month.to_i(10))
```
willfish, line 55: Isn't base 10 the default?

alexdesi: Good point, yes! it is, I was thinking of removing it,
I kept it to make it clear that if we remove it,
the string might be interpreted as octal (if a string starts with '0').

willfish (approved):

PR #1582, opened 2023-11-14
Rubocop all files

shjohnson:
Add back in the rubocop step removed in previous PR

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -361,7 +361,7 @@ jobs:
   ruby-checks:
     docker:
       - image: cimg/ruby:3.2.2
-    resource_class: small
+    resource_class: medium
     steps:
       - checkout
       - run:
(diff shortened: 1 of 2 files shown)


willfish (approved):

.circleci/config.yml
```diff
@@ -361,7 +361,7 @@ jobs:
   ruby-checks:
     docker:
       - image: cimg/ruby:3.2.2
-    resource_class: small
+    resource_class: medium
```
willfish, line 364: Do we need it to be medium?

shjohnson: Yes it kept dying when checking the files before and even tried it here in a separate PR and still got the kill signal from CI

jebw: Its hitting memory limits otherwise - probably number of changed files so whilst this is a bit PR, going forward it'll be scanning all files anyway so running out of memory will become an common issue

This step runs for all of 20 seconds so the impact on cost is tiny

jebw (commented):
I'm a bit late on this but would be nice to make the linting job a blocker on deploying

More importantly this would prevent introducing brakeman failures into the codebase

PR #1584, opened 2023-11-17
HOTT-4350: Green Lanes: Create Stub API end point

HWallenberg:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-4350

### What?
Added Green Lanes Subheading Stub API end point

I have added/removed/altered:

- [x] Created API Stub Endpoint and relevant tests
- [x] Added Serializer and routes

### Why?

I am doing this because:

- To start green lanes work
- I should be able to query an end point which will eventually provide 6 digit permutations information but is currently just a stub

### DIFF
--- a/app/controllers/api/v2/green_lanes/subheadings_controller.rb
+++ b/app/controllers/api/v2/green_lanes/subheadings_controller.rb
@@ -2,13 +2,11 @@ module Api
   module V2
     module GreenLanes
       class SubheadingsController < ApiController
-
         def show
           subheading = Subheading.actual.where(goods_nomenclature_item_id: "#{params[:id]}0000", producline_suffix: '80').take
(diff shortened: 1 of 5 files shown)


app/controllers/api/v2/green_lanes/subheadings_controller.rb
```diff
@@ -0,0 +1,13 @@
+module Api
+  module V2
+    module GreenLanes
+      class SubheadingsController < ApiController
```
willfish, line 4: Is it actually a subheading?

alexdesi: well, I think so, it should be just 6 digits

jebw (approved):
LGTM - can we just change the ticket heading to mention green lanes since

maverickvi (approved):

PR #1588, opened 2023-11-20
Hott 4409 refactor green lanes controllers

HWallenberg:
### Jira link
https://transformuk.atlassian.net/jira/software/projects/HOTT/boards/96?selectedIssue=HOTT-4409&text=4409

### What?
Move green lanes goods nomenclature retrieval to its own service

I have added/removed/altered:

- [x] Added Subheading_Search Service that checks for 6, 8 10 digit nomenclature ID

### Why?

To refactor the controller


jebw (changes requested):
Looks good, queries are around naming and constrain which GNs can be retrieved

jebw (approved):
LGTM

maverickvi (approved):

app/services/green_lanes/fetch_subheadings_service.rb
```diff
@@ -0,0 +1,21 @@
+module GreenLanes
+  class FetchSubheadingsService
+    ITEM_ID_LENGTH = 10
+    def initialize(goods_nomenclature_item_id)
```
rasikasri, line 4: We had this discussion about writing a service class with a constructor that needs all the parameters for the service method. In this way, you need to create an instance of service class for each request and one instance can only process one goods_nomenclature_item_id. Is there a real reason for following this pattern in ruby?

rasikasri: Generally in OOP, you pass constructor parameters, when the service cannot do it's job without those instances or values passed in.

alexdesi: Good point. In this case, I think that's fine since we are going to expand the service, and we'll need `goods_nomenclature_item_id` from some private methods.

PR #1594, opened 2023-11-23
Add simple class for avoiding searching for eroneous terms

shjohnson:
This PR avoids searching for certain terms that we dont want to show to the user and currently returns a nill search for for iterations of gifts.

### DIFF
--- a/app/controllers/api/v2/search_controller.rb
+++ b/app/controllers/api/v2/search_controller.rb
@@ -14,7 +14,9 @@ module Api
       private
 
       def matching_suggestions
-        return SearchSuggestion.fuzzy_search(params[:q]) if params[:q].present?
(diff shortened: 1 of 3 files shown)


app/controllers/api/v2/search_controller.rb
```diff
@@ -14,7 +14,9 @@ def suggestions
       private
 
       def matching_suggestions
-        return SearchSuggestion.fuzzy_search(params[:q]) if params[:q].present?
+        if !SearchService::RogueSearchService.call(params[:q]) && params[:q].present?
```
ashward, line 17: Should check whether `params[:q]` exists before using it
```suggestion
        if params[:q].present? && !SearchService::RogueSearchService.call(params[:q])
```

ashward (approved):
lgtm

PR #1603, opened 2023-11-28
Block deploying to dev before linting code

shjohnson:
### Jira link

HOTT-4392

### What?

Ensuring we arent wasting processing power when the ruby linting fails early on in the development pipeline.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -623,8 +623,6 @@ workflows:
       - test:
           context: trade-tariff
           <<: *filter-not-main
-          requires:
-            - ruby-checks
 
       - write-docker-tag:
           name: write-docker-tag-dev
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -652,6 +653,8 @@ workflows:
           environment: development
           image_name: tariff-backend
           ssm_parameter: "/development/BACKEND_ECR_URL"
+          requires:
```
jebw, line 656: This isn't needed - since it's blocked at the plan-terraform stage and would mean build and cannot happen concurrently with the rubocop/brakeman checks

jebw (changes requested):
Looks great, really pleased to see this go in - finally brakeman blocks deploy :raised_hands: 

_but_ we can still run the checks concurrently with the build and push

.circleci/config.yml
shjohnson: Ok i only included this not to waste a tiny bit of processing if it doesnt pass
Happy to remove

willfish (approved):

PR #1607, opened 2023-11-29
BAU: Fix production configuration

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Altered the production configuration

### Why?

I am doing this because:

- This now reflects current production/supports a different frontend host (e.g. www)
- It also adds a missing email address needed to send CSV files of exchange rates to HMRC

### DIFF
--- a/terraform/config_production.tfvars
+++ b/terraform/config_production.tfvars
@@ -1,6 +1,7 @@
 region                                = "eu-west-2"
 environment                           = "production"
-base_domain                           = "tradetesting.net"
+base_domain                           = "tradetesting.net" # trade-tariff.service.gov.uk
+frontend_base_domain                  = "tradetesting.net" # www.trade-tariff.service.gov.uk
 cpu                                   = 1024
 memory                                = 2048
 alcohol_coercian_starts_from          = "2022-01-01"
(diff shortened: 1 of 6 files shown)


terraform/locals.tf
```diff
@@ -77,6 +77,10 @@ locals {
       name  = "TARIFF_SUPPORT_EMAIL"
       value = "hmrc-trade-tariff-support-g@digital.hmrc.gov.uk"
     },
+    {
+      name  = "TARIFF_MANAGEMENT_EMAIL"
```
willfish, line 81: This is needed to send custom CSV emails of exchange rates to the HMRC team

terraform/config_production.tfvars
```diff
@@ -1,6 +1,7 @@
 region                                = "eu-west-2"
 environment                           = "production"
-base_domain                           = "tradetesting.net"
+base_domain                           = "tradetesting.net" # trade-tariff.service.gov.uk
+frontend_base_domain                  = "tradetesting.net" # www.trade-tariff.service.gov.uk
```
amberstarlight, line 4: We'll have the root working after Saturday's migration, so do we need this?

willfish: Its a noop atm

amberstarlight (approved):

PR #1630, opened 2023-12-13
HOTT-4632 Removal of wrong test search serialisation

shjohnson:
### Jira link

HOTT-4632

### What?

I have added/removed/altered:

- [ ] Changed the serialiser used in the search service test as not used anymore
- [ ] Removed the old serialiser and error serialiser

### Why?

The wrong serialisation was being used in the tests as the V! is not used anymore

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -38,8 +38,8 @@ private
   end
 
   def serialize_errors(errors)
-    if request.headers['Accept'] == 'application/vnd.uktt.v2'
-      Api::V2::ErrorSerializationService.new.serialized_errors(errors)
+    # if request.headers['Accept'] == 'application/vnd.uktt.v2'
+    Api::V2::ErrorSerializationService.new.serialized_errors(errors)
     # else
     #   Api::V1::ErrorSerializationService.new.serialized_errors(errors)
     # end
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -38,11 +38,11 @@ private
   end
 
   def serialize_errors(errors)
-    # if request.headers['Accept'] == 'application/vnd.uktt.v2'
-    Api::V2::ErrorSerializationService.new.serialized_errors(errors)
-    # else
-    #   Api::V1::ErrorSerializationService.new.serialized_errors(errors)
-    # end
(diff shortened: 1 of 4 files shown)


app/controllers/errors_controller.rb
```diff
@@ -38,11 +38,11 @@ def respond_to_error(status, message)
   end
 
   def serialize_errors(errors)
-    if request.headers['Accept'] == 'application/vnd.uktt.v2'
-      Api::V2::ErrorSerializationService.new.serialized_errors(errors)
-    else
-      Api::V1::ErrorSerializationService.new.serialized_errors(errors)
-    end
+    # if request.headers['Accept'] == 'application/vnd.uktt.v2'
```
shjohnson, line 41: Commented this out for now because im pretty sure this can be removed but wanted to get some eyes on it first

shjohnson: re-added as still needed

spec/services/search_service_spec.rb
```diff
@@ -441,16 +445,16 @@
         }.ignore_extra_keys!
       end
 
-      it 'returns goods code if search date falls within validity period' do
-        result = TimeMachine.at('2005-01-01') { described_class.new(data_serializer, q: 'water').to_json }
+      context 'with search date falls within validity period returns goods code' do
```
willfish, line 448: Small thing but I think you don't want your expectation here
```suggestion
      context 'with search date falls within validity period' do
```

app/controllers/errors_controller.rb
willfish: This is one for @jebw, really, but I don't think we can remove it since we sometimes have v1 in the Accept header as per https://github.com/trade-tariff/trade-tariff-frontend/blob/main/lib/trade_tariff_frontend/request_forwarder.rb#L30

spec/services/search_service_spec.rb
shjohnson: Yeah i thought these phrases could be improved

app/controllers/errors_controller.rb
shjohnson: Yeah think for now ive added back in so makes sense to leave and get the other change out 🙌

willfish (approved):

PR #1631, opened 2023-12-13
HOTT-4557: Publication date code fix

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4557

### What?

I have added/removed/altered:

- [ ] updated the code to use the correct publication date when creating ExchangeRateFiles

### Why?

I am doing this because:

- We were saving/displaying the incorrect date on ExchangeRates

Screenshot 2023-12-13 at 16 51 37

### DIFF
--- a/app/services/exchange_rates/monthly_exchange_rates_service.rb
+++ b/app/services/exchange_rates/monthly_exchange_rates_service.rb
@@ -19,18 +19,21 @@ module ExchangeRates
         rates,
         date,
         :monthly_csv,
+        sample_date,
       ).call
 
       ExchangeRates::UploadFileService.new(
         rates,
         date,
         :monthly_xml,
+        sample_date,
       ).call
 
       ExchangeRates::UploadFileService.new(
         rates,
         date,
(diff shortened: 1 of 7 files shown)


app/services/exchange_rates/monthly_exchange_rates_service.rb
```diff
@@ -19,18 +19,21 @@ def call
...
       ExchangeRates::UploadFileService.new(
         rates,
         date,
         :monthly_xml,
+        sample_date,
       ).call
 
       ExchangeRates::UploadFileService.new(
         rates,
         date,
```
shjohnson, line 34: I wondered if we can have better naming for `date` this is the same as the sample date right?

shjohnson: Think its just a bit confusing where we have `selected_date`, `date` and `sample_date` unless im not understanding

willfish (approved):
LGTM

jebw (approved):

PR #1633, opened 2023-12-14
HOTT-4561: Categorisation API

rasikasri:
### Jira link

[HOTT-4561]<https://transformuk.atlassian.net/browse/HOTT-4561>

### What?

I have added/removed/altered:

- [ ] Added a simple API to serialize all of the categorisation records

### Why?

I am doing this because:

- This will facilitate service like TSS to implement green lanes if they plan to fully recalculate the logic and use over night scrapes
-
-

### Have you? (optional)

### DIFF
--- a/spec/serializers/api/v2/green_lanes/categorisation_serializer_spec.rb
+++ b/spec/serializers/api/v2/green_lanes/categorisation_serializer_spec.rb
@@ -1,0 +1,35 @@
+RSpec.describe Api::V2::GreenLanes::CategorisationSerializer do
+  subject(:serialized) do
+    described_class.new(::GreenLanes::Categorisation.load_from_string json_string).serializable_hash.as_json
+  end
+
+  let(:json_string) do
+    '[{
+          "category": "1",
+          "regulation_id": "D0000001",
+          "measure_type_id": "400",
+          "geographical_area": "1000",
+          "document_codes": [],
+          "additional_codes": []
+        }]'
+  end
+
+  let(:expected_pattern) do
+    {
+      data: [
+        id: be_a(String),
+        type: 'green_lanes_categorisation',
+        attributes: {
+          category: '1',
+          regulation_id: 'D0000001',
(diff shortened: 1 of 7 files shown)


spec/serializers/api/v2/green_lanes/categorisation_serializer_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  let(:json_string) do
+    '[{
+          "category": "1",
+          "regulation_id": "D0000001",
+          "measure_type_id": "400",
+          "geographical_area": "1000",
+          "document_codes": [],
+          "additional_codes": []
+        }]'
```
alexdesi, line 14: the indentation looks strange here

alexdesi (dismissed):
It looks good to me—just a minor comment.

spec/serializers/api/v2/green_lanes/categorisation_serializer_spec.rb
alexdesi: I safer approach is to use Ruby structure and convert it `to_json` :
```
[{
  "category"=>"1", 
  "regulation_id"=>"D0000001", 
  "measure_type_id"=>"400", 
  "geographical_area"=>"1000", 
  "document_codes"=>[], 
  "additional_codes"=>[]
}].to_json
```
This way the object is parsed and the json format is correct.

jebw (changes requested):
LGTM - can you tweak the source JSON data in the spec - either amend the indendation or use a ruby hash with `.to_json` as @alexdesi suggested

If you're wanting to use a string, you can generate multiline strings with heredoc (but `.to_json` is a better approach), eg

```ruby
class SomeClass
  def some_method
    <<~ EOS
      {
        "key" => "value"
      }
    EOS
  end
end
```
```

jebw (approved):
LGTM :raised_hands:

PR #1634, opened 2023-12-14
HOTT-4578: Enable ancestor search on negated descriptions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4578

### What?

I have added/removed/altered:

- [x] Adds negated ancestor descriptions to the commodities index
- [x] Adds negated descriptions to the chapters index
- [x] Adds negated descriptions to the headings index
- [x] Extended negated descriptions handling to support more sophisticated scenarios
- [x] Adds ancestor descriptions to the commodity index

### Why?

I am doing this because:

- This significantly improves the way the results are returned in legacy
search and is being driven by our client.
- The ancestor descriptions will enable us to contextualise commodity descriptions that are simply "Other"


app/services/search_service/fuzzy_search/goods_nomenclature_query.rb
```diff
@@ -6,70 +6,82 @@ def query(query_opts = {})
...
               },
             },
             size: INDEX_SIZE_MAX,
           },
         }
       end
+
+      def multi_match_part(query_opts = {})
+        query = {
+          # match the search phrase
```
willfish, line 65: This comment is a lie - we'd want shingles if we wanted to match a phrase. Remove me

spec/factories/goods_nomenclature_factory.rb
```diff
@@ -38,6 +38,10 @@
     after(:create) do |gono, _evaluator|
       if gono.associations[:goods_nomenclature_indents]
         gono.associations[:goods_nomenclature_indents].all?(&:save)
+        # We clear associations that rely on nested set that might have been
+        # loaded before the nested set tree node view was refreshed
+        gono.associations.clear
```
jebw, line 43: Would it make sense to only clear the tree nodes and do it after the refresh has happened?

willfish: We could but it makes no difference.

shjohnson (dismissed):
LGTM

PR #1636, opened 2023-12-15
HOTT-4628: Extend Backend ErrorsController

rasikasri:
### Jira link

<[HOTT-4623](https://transformuk.atlassian.net/browse/HOTT-4628)>

### What?

I have added/removed/altered:

- [ ] Added actions in the controller for the missing error handling actions

### Why?

I am doing this because:

- There are certain exceptions in which are mapped in the default Rails behaviour to actions we don’t currently support
-
-

### Have you? (optional)


spec/requests/errors_controller_spec.rb
```diff
@@ -75,4 +75,28 @@
   describe 'GET #serice_unavailable' do
     it_behaves_like 'a csv or json error response', 503, 'Service is unavailable'
   end
+
+  describe 'GET #maintenance' do
```
alexdesi, line 79: `def maintenance` didn't have a spec.

willfish (dismissed):
LGTM

spec/requests/errors_controller_spec.rb
jebw: We've already got a separate feature spec for maintenance mode so this isn't required, see

`spec/features/maintenance_mode_spec.rb`

jebw (changes requested):
Looks good but we have already got a separate feature spec for maintenance mode?

willfish (approved):

willfish (approved):

willfish:
The feature spec for the maintenance mode is a bit different from the plumbing tests of the errors even though it piggy backs off of the same errors

PR #1644, opened 2023-12-21
HOTT-4446: Adds HEALTHCHECK

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4449

### What?

ECS healthchecks are managed separately:

https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_HealthCheck.html

I have added/removed/altered:

- [x] Added HEALTHCHECK to Dockerfile
- [x] Make Dockerfile run locally with minimum-required configuration

### Why?

I am doing this because:

- This is remediation work from our pentest report

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -15,6 +15,7 @@ MERGED_XI_UK_DATABASES=true
 NESTED_SET_SUBHEADINGS=true
 PERMUTATIONS=true
 PGHOST=localhost
+PORT=3000
 REPORTING_CDN_HOST=https://reporting.trade-tariff.service.gov.uk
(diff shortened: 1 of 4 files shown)


.env.development
```diff
@@ -15,6 +15,7 @@ MERGED_XI_UK_DATABASES=true
 NESTED_SET_SUBHEADINGS=true
 PERMUTATIONS=true
 PGHOST=localhost
+PORT=3000
```
willfish, line 18: Explicit, here, so we have a port to override the default image port of 8080

amberstarlight (approved):
Great stuff!

PR #1650, opened 2024-01-03
Add filter for GreenLanes::Categorisation.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-4351

### What?
Add filter for GreenLanes::Categorisation.

### DIFF
--- a/app/models/green_lanes/categorisation.rb
+++ b/app/models/green_lanes/categorisation.rb
@@ -31,6 +31,18 @@ module GreenLanes
       def all
         @all ||= load_from_file
       end
+
+      def filter(regulation_id:, measure_type_id:, geographical_area: nil)
+        return [] if regulation_id.blank? || measure_type_id.blank?
+
+        all.select { |c| c.match?(regulation_id:, measure_type_id:, geographical_area:) }
+      end
+    end
+
+    def match?(regulation_id:, measure_type_id:, geographical_area: nil)
+      regulation_id == self.regulation_id &&
+        measure_type_id == self.measure_type_id &&
+        (geographical_area.nil? || geographical_area == self.geographical_area)
     end
 
     class InvalidFile < RuntimeError; end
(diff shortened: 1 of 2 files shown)


app/models/green_lanes/categorisation.rb
```diff
@@ -31,6 +31,18 @@ def load_from_string(data)
       def all
         @all ||= load_from_file
       end
+
+      def filter(regulation_id:, measure_type_id:, geographical_area: nil)
+        return [] if regulation_id.blank? || measure_type_id.blank?
+
+        all.select { |c| c.match?(regulation_id:, measure_type_id:, geographical_area:) }
```
jebw, line 38: super minor suggestion but you could just use `...` to pass all params through?

spec/models/green_lanes/categorisation_spec.rb
```diff
@@ -101,4 +101,46 @@
...
+      subject(:categorisation_filter) do
+        described_class.filter regulation_id: 'D000004', measure_type_id: '430', geographical_area: 'US'
+      end
+
+      it { expect(categorisation_filter.first).to have_attributes regulation_id: 'D000004', measure_type_id: '430' }
+    end
+
+    context 'when geographical_area does not match' do
+      subject(:categorisation_filter) do
+        described_class.filter regulation_id: 'D000004', measure_type_id: '430', geographical_area: 'unexisting'
```
jebw, line 139: how about `non-existant` ? `unexisting` isn't really a word AFAIK?

spec/models/green_lanes/categorisation_spec.rb
```diff
@@ -101,4 +101,46 @@
...
+      end
+
+      it { is_expected.to be_an Array }
```
jebw, line 142: TIL - `be_an` :smile:

jebw (approved):
LGTM - couple of really minor suggestions

app/models/green_lanes/categorisation.rb
alexdesi: in this case I'd prefer to be more explicit

spec/models/green_lanes/categorisation_spec.rb
alexdesi: good point, I'll chat it.
Strangely some dictionaries still report it as a proper word, but they consider "non-existent"   wrong now.

jebw (approved):
Nice, super clean!

PR #1652, opened 2024-01-03
HOTT-4737: Move storage of Permutations data to S3 bucket

rasikasri:
### Jira link

[HOTT-4737](https://transformuk.atlassian.net/browse/HOTT-4737)

### What?

I have added/removed/altered:

- [x] Added functionality to read categorisation file from s3 bucket

### Why?

I am doing this because:

- We need to keep the permutations json file private longer term so plan is to add it to the existing persistence S3 bucket
-
-

### Have you? (optional)


jebw (dismissed):
LGTM :raised_hands:

spec/models/green_lanes/categorisation_spec.rb
```diff
@@ -104,6 +104,45 @@
...
+          "geographical_area": "1000",
+          "document_codes": [],
+          "additional_codes": []
+        }]'
+    end
+
+    context 'when the file exists in S3' do
+      subject(:s3_categories) { described_class.load_from_s3 }
+
+      # rubocop:disable RSpec/MessageChain
```
willfish, line 123: So I personally don't mind message chains. Should we just add this to the rubocop config file?

rasikasri: Here I tried to avoid message chains and allow any instance by mocking `Rails.application.config.persistence_bucket`, but it is not possible as it is a frozen object. In other places probably you can avoid message chain by using mock object. So I not sure if we need to disable this cop globally. Having said that, I am not sure what is the issue of using message chain either.

jebw: Yeh, I'd agree @willfish  we could just disable that rubocop rule globally?

willfish: Yeah, I also can't see how that object is frozen in the initialiser but probably I'm missing something and it's not super important anyway

rasikasri: OK, I'll move it to global level then.

jebw (changes requested):
Sorry to be a pain - looks like we've talked at cross purposes around rspec rules

* AnyInstance rule - this should still be enforced globally with rubycop comments in specs where unavoidable
* MessageChain - this should be disabled globally because we don't mind this one

rasikasri:
> Sorry to be a pain - looks like we've talked at cross purposes around rspec rules
> 
> * AnyInstance rule - this should still be enforced globally with rubycop comments in specs where unavoidable
> * MessageChain - this should be disabled globally because we don't mind this one

I think I missundersoot that. Now it is corrected

jebw (approved):
LGTM :raised_hands:

PR #1655, opened 2024-01-04
Fix average rate creation bugs

shjohnson:
### Jira link

HOTT-4754

### What?

I have added/removed/altered:

- [ ] Improved avg rate service to catch instances where a country might have been a live country but actually the test cases didnt cover if a rate wasnt valid for that period (ie not covering the full 12 periods)
- [ ] Improved some wording around comments
- [ ] Fixed bug where service would have created a rate no matter when the country was live or not
- [ ] Remove uneeded method `live_last_twelve_months`
- [ ] Changed validity start period to be the beginning of the first month 11 months previous to the run date

### Why?

I am doing this because:

- Previously the avg rate task would have generated a rate for any countries that had a live rate in the last 12 months
-
-

### DIFF
--- a/app/services/exchange_rates/create_average_exchange_rates_service.rb
+++ b/app/services/exchange_rates/create_average_exchange_rates_service.rb
@@ -2,7 +2,7 @@ module ExchangeRates
   class CreateAverageExchangeRatesService
     # Creates the average rates for countries that have had a live rate the last 12 months and upload the file
 
-    VALID_MONTHS = [4, 12].freeze
+    VALID_MONTHS = [3, 12].freeze
     VALID_DAY = 31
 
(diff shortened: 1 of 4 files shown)


app/services/exchange_rates/create_average_exchange_rates_service.rb
```diff
@@ -2,7 +2,7 @@ module ExchangeRates
   class CreateAverageExchangeRatesService
     # Creates the average rates for countries that have had a live rate the last 12 months and upload the file
 
-    VALID_MONTHS = [4, 12].freeze
+    VALID_MONTHS = [3, 12].freeze
```
willfish, line 5: :raised_hands:

willfish (approved):

PR #1658, opened 2024-01-08
HOTT-4736: Restrict access to new APIs

rasikasri:
### Jira link

[HOTT-4736](https://transformuk.atlassian.net/browse/HOTT-4736)

### What?

I have added/removed/altered:

- [ ] Added access control to green lane api

### Why?

I am doing this because:

- we need to restrict access to green lanes api for selected users

### Have you? (optional)


jebw (commented):
Looks good. You could avoid the duplication on the auth specs by moving them to a shared example? No strong feeling on that though because at this point thats only one level of duplication.

Unless I'm missing something though the auth specs are missing a check for no auth header? I've done some manual testing and it works the right way but should still have a spec for it?

jebw (approved):
Awesome! nice pattern we can use for future in development APIs as well :raised_hands:

spec/requests/api/v2/green_lanes/categorisation_controller_spec.rb
```diff
@@ -1,15 +1,25 @@
 require 'rails_helper'
 
 RSpec.describe Api::V2::GreenLanes::CategorisationsController do
+  before do
```
willfish, line 4: We ought to not mock the env if we can help it

willfish: Can we maybe encapsulate the env var in the application config module TradeTariffBackend and mock that?

rasikasri: Done this with moving to TradeTariffBackend and mocking it instead of ENV

willfish (approved):

PR #1659, opened 2024-01-08
HOTT-4814 Disable ETag caching for exchange rates

jebw:
### Jira link

HOTT-4814

### What?

I have added/removed/altered:

- [x] Prevents ExchangeRates giving out invalid cache info to the frontend

### Why?

I am doing this because:

- ExchangeRates API responses are getting cached and should not be

- May increase backend load on server

### DIFF
--- a/app/controllers/api/v2/exchange_rates/base_controller.rb
+++ b/app/controllers/api/v2/exchange_rates/base_controller.rb
@@ -2,6 +2,8 @@ module Api
   module V2
     module ExchangeRates
       class BaseController < ApiController
+        include NoCaching
+
         before_action :validate_exchange_rate_type
 
         private


shjohnson (approved):

app/controllers/api/v2/exchange_rates/base_controller.rb
```diff
@@ -2,6 +2,8 @@ module Api
   module V2
     module ExchangeRates
       class BaseController < ApiController
+        include NoCaching
```
willfish, line 5: I guess this is easier than tracking the events that drive the exchange rate changes

willfish (approved):

app/controllers/api/v2/exchange_rates/base_controller.rb
jebw: Yes, it is possible to generate appropriate ETags but its a bigger change since at the moment we lack an identifier for when the S3 contents are updated. 

Hence conclusion was to just remove the caching - that being said it sounds like the underlying issue wasn't ETags anyway but this is lowest hanging fruit

PR #1661, opened 2024-01-09
HOTT-4816 remove error handling duplication

jebw:
### Jira link

HOTT-4816

### What?

I have added/removed/altered:

- [x] Removed the duplicate error handling in api controller
- [x] Fixed up various controller tests which expected exceptions to be rescued which is no longer the case

### Why?

I am doing this because:

- This is duplicating rails' own functionality
- This means we also need to try and keep this in sync with Errors Controller
- It is reporting our internal exception messages to users which seems like a bad idea
- It shows the minimal error message rather than the stock rails exception screen in development mode making it harder for developers to discover the origin of an error

- Changes exception handling code
- May result in increased error reporting if there is something which was previously getting rescued and which has not been added to the Sentry exclusions lists

### DIFF
--- a/config/initializers/exceptions.rb
+++ b/config/initializers/exceptions.rb
@@ -1,7 +1,9 @@
 ActionDispatch::ExceptionWrapper.rescue_responses.merge!(
   'ArgumentError' => :bad_request,
-  'NotImplementedError' => :bad_request,
+  'NotImplementedError' => :unprocessable_entity,
+  'ActionController::ParameterMissing' => :unprocessable_entity,
   'Sequel::RecordNotFound' => :not_found,
+  'Sequel::NoMatchingRow' => :not_found,
   'BulkSearch::ResultCollection::RecordNotFound' => :not_found,
   'ActionController::RoutingError' => :not_found,
   'AbstractController::ActionNotFound' => :not_found,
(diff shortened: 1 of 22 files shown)


config/initializers/exceptions.rb
```diff
@@ -1,7 +1,9 @@
 ActionDispatch::ExceptionWrapper.rescue_responses.merge!(
   'ArgumentError' => :bad_request,
-  'NotImplementedError' => :bad_request,
+  'NotImplementedError' => :unprocessable_entity,
+  'ActionController::ParameterMissing' => :unprocessable_entity,
```
alexdesi, line 4: I guess you removed this because it is the default behaviour.
Correct?

spec/controllers/api/admin/chapters/chapter_notes_controller_spec.rb
```diff
@@ -30,9 +30,9 @@
       let(:chapter) { create :chapter }
 
       it 'returns not found if record was not found' do
-        get :show, params: { chapter_id: chapter.to_param }, format: :json
-
-        expect(response.status).to eq 404
```
alexdesi, line 35: I think the advantage of texting the 404 status,
is that it's closer to the production behaviour.

We test the behaviour we expect in production.

alexdesi (approved):
Nice cleaning!
I approved, but I am not sure about those changes in the specs.
I think we should test the behaviour we expect in production.

PR #1668, opened 2024-01-11
Fix Broken RoO Proofs ORD URL

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4516

### What?

I have

 - [x]  Updated the Origin Reference Document URL in the origin-declaration.md files with a placeholder of {ord_url} so this can be dynamically inserted by the Front End with app/helpers/rules_of_origin_helper.rb>insert_ord_url for part of the proofs section.
 
### Why?

I am doing this because:

The URL to an ORD in the RoO proofs section was never correct and has to be dynamic

### Have you?
- [x] Reviewed view changes with stake holders

### Deployment risks
Possible backend files with updated placeholder may not deploy but this is very rare and won't cause an error.

Screenshot of origin reference document link at bottom of the screen
image


jebw (approved):
LGTM

maverickvi (approved):

PR #1669, opened 2024-01-11
HOTT-4831: Extend terraform scripts to assign auth token

rasikasri:
### Jira link

[HOTT-4831](https://transformuk.atlassian.net/browse/HOTT-4831)

### What?

I have added/removed/altered:

- [x] Added terraform scripts to extract green lanes tokens

### Why?

I am doing this because:

- Now that we have auth support for the green lanes APIs, we need to assign a token to the env var so they can use it
-
-

### Have you? (optional)


alexdesi (approved):
It looks all right to me, but I am not 100% we are not missing something, I am not a terraform expert :)

PR #1677, opened 2024-01-17
HOTT-4236-Update-DCTS-PSRs

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4236

### What?

I have altered:

- [x] db/rules_of_origin/roo_schemes_uk/rule_sets/dcts-general-enhanced.json 

### Why?

I am doing this because:

- The rule set for DCTS countries has been updated for 2024

PSR for Commodity  3903110000 Polystyrene 

**BEFORE**

image

**AFTER**

image


alexdesi (approved):
It looks good, but I cannot say if all the changes in the JSON make sense.
Probably we should rely on manual testing on this.

PR #1685, opened 2024-01-23
HOTT-4949: Fix tests and remove the "flaky" flag from rspec.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-4949

### What?

Fix tests and remove the "flaky" flag from rspec serializers/search/goods_nomenclature_serializer_spec.rb.

### DIFF
--- a/app/serializers/search/goods_nomenclature_serializer.rb
+++ b/app/serializers/search/goods_nomenclature_serializer.rb
@@ -15,7 +15,7 @@ module Search
           description:,
           description_indexed:,
           description_indexed_shingled: description_indexed,
-          formatted_description:,
+          formatted_description: formatted_description || '',
           search_references:,
           search_intercept_terms:,
           ancestors:,
@@ -23,7 +23,7 @@ module Search
           validity_end_date:,
           guides:,
           guide_ids:,
-          declarable: true,
+          declarable: declarable?,
         }
 
         1.upto(MAX_ANCESTORS) do |i|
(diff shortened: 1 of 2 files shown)


app/serializers/search/goods_nomenclature_serializer.rb
```diff
@@ -15,15 +15,15 @@ def serializable_hash(_opts = {})
           description:,
           description_indexed:,
           description_indexed_shingled: description_indexed,
-          formatted_description:,
+          formatted_description: formatted_description || '',
```
alexdesi, line 18: For the formatted one, we expect empty space and not nil.

app/serializers/search/goods_nomenclature_serializer.rb
```diff
@@ -15,15 +15,15 @@ def serializable_hash(_opts = {})
...
           search_references:,
           search_intercept_terms:,
           ancestors:,
           validity_start_date:,
           validity_end_date:,
           guides:,
           guide_ids:,
-          declarable: true,
+          declarable: declarable?,
```
alexdesi, line 26: Not sure about this?
The tests 
created a non-declarable and was expecting a declarable false in the serialised, which makes sense.

But the code returned `true` hardcoded! 
My hunch is that this has not been run, since it was marked as "flaky".

HWallenberg (approved):

shjohnson (approved):

PR #1687, opened 2024-01-26
Update sidekiq and rails to 7.1.3

shjohnson:
### Jira link

HOTT-

### What?

These gem upgrades help update sidekiq to fix the error: undefined method `broadcast'

I have added/removed/altered:

- [ ] update rails and sidekiq gems


alexdesi (approved):
:green_heart:

PR #1691, opened 2024-01-29
HOTT-4994 Added theme to categorisation model

jebw:
### Jira link

HOTT-4994

### What?

I have added/removed/altered:

- [x] Added theme attribute to green lanes `Categorisation` model

### Why?

I am doing this because:

- We will want to expose this in the API

- Low, private API - specs test both with and without key in JSON

### DIFF
--- a/lib/tasks/green_lanes.rake
+++ b/lib/tasks/green_lanes.rake
@@ -7,11 +7,12 @@ namespace :green_lanes do
     @json = @data.map do |row|
       {
         category: row['Primary category'],
-        regulation_id: row['Regulation id'].presence.strip,
-        measure_type_id: row['Measure type ID'].presence.strip,
+        regulation_id: row['Regulation id'].presence&.strip,
+        measure_type_id: row['Measure type ID'].presence&.strip,
         geographical_area: row['Geographical area']&.presence&.strip,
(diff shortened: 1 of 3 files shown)


lib/tasks/green_lanes.rake
```diff
@@ -7,11 +7,12 @@ namespace :green_lanes do
...
       {
         category: row['Primary category'],
-        regulation_id: row['Regulation id'].presence.strip,
-        measure_type_id: row['Measure type ID'].presence.strip,
+        regulation_id: row['Regulation id'].presence&.strip,
+        measure_type_id: row['Measure type ID'].presence&.strip,
         geographical_area: row['Geographical area']&.presence&.strip,
         document_codes: row['Document codes'].to_s.split.map(&:strip),
         additional_codes: row['Additional codes'].to_s.split.map(&:strip),
+        theme: row['Theme'].to_s,
```
alexdesi, line 15: just to check,
Do we need `.strip` here?

alexdesi (approved):
It looks good :white_check_mark: 
just check we do not need "strip" o the new field

alexdesi (approved):
:rocket: :white_check_mark:

PR #1693, opened 2024-01-31
HOTT-4996 Add green lanes db tables

jebw:
### Jira link

HOTT-4996

### What?

I have added/removed/altered:

- [x] Added a `green_lanes_themes` table
- [x] Added a `green_lanes_category_assessments` table

### Why?

I am doing this because:

- A theme implies a category, we need to track these if we are to categorise commodities
- A CategoryAssessment is derived from the measure type and (optionally) the regulation of the measures against a commodity - so track these combinations together with their corresponding theme (and so implied category)

- Adds new database tables, in its own PR so it can be deployed independently of anything depending upon it due to current complications in deployment process around migrations


jebw:
Wrong ticket/branch number - superceded by #1695

PR #1696, opened 2024-02-01
HOTT-4930 Added simple check that sync is happening

jebw:
### Jira link

HOTT-4930

### What?

I have added/removed/altered:

- [x] Added worker task which checks QuotaBalanceEvents have updated in the last 5 days and notifies Sentry if that is not the case

### Why?

I am doing this because:

- It should alert us if data is getting out of sync - QuotaBalanceEvent is our most frequently changing table

- None - background job, just checks the data is recent

### DIFF
--- a/app/workers/synchronizer_check_worker.rb
+++ b/app/workers/synchronizer_check_worker.rb
@@ -1,7 +1,7 @@
 class SynchronizerCheckWorker
   include Sidekiq::Worker
 
-  RECENCY_THRESHOLD = 5 # days ago
+  RECENCY_THRESHOLD = 4 # days ago
 
   def perform(check_since = RECENCY_THRESHOLD)
     latest_qbe = QuotaBalanceEvent::Operation.order_by(Sequel.desc(:oid)).first
--- a/app/workers/synchronizer_check_worker.rb
+++ b/app/workers/synchronizer_check_worker.rb
@@ -6,11 +6,21 @@ class SynchronizerCheckWorker
   def perform(check_since = RECENCY_THRESHOLD)
     latest_qbe = QuotaBalanceEvent::Operation.order_by(Sequel.desc(:oid)).first
 
-    unless latest_qbe && latest_qbe.created_at >= check_since.days.ago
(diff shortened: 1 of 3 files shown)


app/workers/synchronizer_check_worker.rb
```diff
@@ -0,0 +1,16 @@
+class SynchronizerCheckWorker
+  include Sidekiq::Worker
+
+  RECENCY_THRESHOLD = 5 # days ago
+
+  def perform(check_since = RECENCY_THRESHOLD)
+    latest_qbe = QuotaBalanceEvent::Operation.order_by(Sequel.desc(:oid)).first
+
+    unless latest_qbe && latest_qbe.created_at >= check_since.days.ago
```
alexdesi, line 9: this is correct,
but a guard clause simplifies the condition:

```
return unless latest_qbe

if  latest_qbe.created_at < check_since.days.ago
...
```

app/workers/synchronizer_check_worker.rb
```diff
@@ -0,0 +1,16 @@
...
+      msg = "Potential sync problem on #{TradeTariffBackend.service&.upcase} service - last update: #{latest_qbe&.created_at}"
```
alexdesi, line 10: The message is generic. I wonder if we could help with more insight. But probably we can't say extract much info from here.

spec/workers/synchronizer_check_worker_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    context 'with no data' do
+      before { described_class.new.perform }
+
+      it 'alerts with the failing service' do
+        expect(Sentry).to have_received(:capture_message).with \
+          %r{Potential sync problem on UK service}
+      end
+    end
+  end
+end
```
alexdesi, line 41: nice tests :100:

alexdesi (approved):
Approved because it's correct.
I'd suggest the guard clause (see comment) to simplify the condition.

alexdesi (approved):
lgtm :white_check_mark:

PR #1697, opened 2024-02-01
Add country of Origin filter to NomenclaturesController

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-4838

### What?
Add origin filter to `GoodsNomenclaturesController
This PR also contains small refactors to clean up the code/specs.

### Why?
It's a requirement. We need to be able to filter by country of origin.


app/controllers/api/v2/green_lanes/goods_nomenclatures_controller.rb
```diff
@@ -4,12 +4,20 @@ module GreenLanes
...
-          serializer = Api::V2::GreenLanes::GoodsNomenclatureSerializer.new(presented_gn, include: %w[applicable_measures possible_categorisations])
+          serializer = Api::V2::GreenLanes::GoodsNomenclatureSerializer.new(
+            presented_gn, include: %w[applicable_measures possible_categorisations]
+          )
 
           render json: serializer.serializable_hash
         end
+
+        def param_origin
+          params.fetch(:origin, nil)
```
jebw, line 19: This doesn't handle a blank param, ie `?origin=`  - can this append `.presence`

jebw (changes requested):
Code looks great - minor comment on the param

_but_ I forgot to update the ticket when I changed the API spec so can the param name change from

`origin` to `filters[geographical_area_id]` - this brings us inline with other parts of the API

my bad :man_facepalming:

spec/requests/api/v2/green_lanes/goods_nomenclatures_controller_spec.rb
```diff
@@ -82,13 +87,31 @@
...
     end
+
+    context 'when origin is provided' do
+      let(:params) { { origin: 'AU' } }
+
+      before do
+        allow(GreenLanes::FindCategorisationsService).to receive(:call).and_call_original
+      end
+
+      it 'call FindCategorisationsService with correct params' do
```
alexdesi, line 98: ops:
```
it 'calls ...`
...
```

rasikasri (approved):

app/controllers/api/v2/green_lanes/goods_nomenclatures_controller.rb
alexdesi: this is not outdated,  we are using `filter` now.

PR #1700, opened 2024-02-01
HOTT-4995: Allow localhost access without auth token

rasikasri:
### Jira link

[HOTT-4995](https://transformuk.atlassian.net/browse/HOTT-4995)

### What?

I have added/removed/altered:

- [ ] Removed green_lanes api authentication on local servers

### Why?

I am doing this because:

- we can test locally from within a browser without requiring curl or postman
-
-

### Have you? (optional)


app/controllers/api/v2/green_lanes/base_controller.rb
```diff
@@ -15,9 +15,11 @@ def check_service
         end
 
         def authenticate
-          authenticate_or_request_with_http_token do |provided_token, _options|
-            Rails.logger.debug provided_token
-            api_tokens.any? { |token| ActiveSupport::SecurityUtils.secure_compare(provided_token, token) }
+          unless TradeTariffBackend.deployed_environment == 'development'
```
jebw, line 18: So this isn't quite what we were discussing - currently this is checking the production server name, not the rails environment - I can chat this through if it helps

In general this should still allow setting the tokens locally for manually testing (in case we need to do any other development on this) - so the condition should be something similar to

```ruby
unless Rails.env.development? and TradeTariffBackend.green_lanes_api_tokens.blank?
```

jebw (changes requested):
This doesn't implement quite what we'd discussed yesterday, I've added a suggestion inline

jebw (approved):
LGTM

PR #1702, opened 2024-02-05
HOTT-4991: Restructure CategoryAssessment entity

rasikasri:
### Jira link

[HOTT-4991](https://transformuk.atlassian.net/browse/HOTT-4991)

### What?

I have added/removed/altered:

- [ ] Restructure CategoryAssessment entity

### Why?

I am doing this because:

- To match with updated api requirements 
-
-

### Have you? (optional)


app/models/green_lanes/category_assessment.rb
```diff
@@ -72,6 +72,10 @@ def match?(regulation_id:, measure_type_id:, geographical_area: nil)
           self.geographical_area == GeographicalArea::ERGA_OMNES_ID)
     end
 
+    def excluded_geographical_areas
```
alexdesi, line 75: Should we add a small test for this method?
(on ../green_lanes/category_assessment_spec.rb)

alexdesi (approved):
lgtm :white_check_mark:

PR #1706, opened 2024-02-06
Allow the worker to have access to use the delete object method on the AWS servers for exchange rate path.

shjohnson:
### Jira link

HOTT-5069

### What?

I have added/removed/altered:

- [ ] Allow the workers to use delete object off the S3 buckets under the exchange rate path.

### Why?

I am doing this to allow the rake tasks for rebuilding the exchange rate files to have delete access.


amberstarlight (approved):
💯

shjohnson:
Just FYI this has been tested on DEV aswell

PR #1708, opened 2024-02-06
HOTT-4992:Replace `geographical_area` with an instance of a `GeographicalArea` entity

rasikasri:
### Jira link

[HOTT-4992](https://transformuk.atlassian.net/browse/HOTT-4992)

### What?

I have added/removed/altered:

- [ ] Linked CategoryAssessment geographical_area relationship to GeographicalArea database model 

### Why?

I am doing this because:

- Match with green_lanes api requirements
-
-

### Have you? (optional)


spec/serializers/api/v2/green_lanes/category_assessment_serializer_spec.rb
```diff
@@ -37,6 +40,12 @@
               { id: '1', type: 'additional_code' },
             ],
           },
+          "geographical_area": {
+            "data": {
+              "id": 'EU',
+              "type": 'geographical_area',
+            },
+          },
```
jebw, line 48: This looks like it should include the following as well - not sure if this is just the test still passing but its there, or if the content is missing?

```json
"excluded_geographical_areas": {
  "data": [
  
  ],
},
```

app/controllers/api/v2/green_lanes/goods_nomenclatures_controller.rb
```diff
@@ -15,6 +15,7 @@ def show
               applicable_measures
               applicable_category_assessments
               applicable_category_assessments.exemptions
+              applicable_category_assessments.geographical_area
```
jebw, line 18: can this also include excluded areas thanks

app/controllers/api/v2/green_lanes/category_assessments_controller.rb
```diff
@@ -4,7 +4,7 @@ module GreenLanes
       class CategoryAssessmentsController < BaseController
         def index
           category_assessment = ::GreenLanes::CategoryAssessment.load_category_assessment
-          serializer = Api::V2::GreenLanes::CategoryAssessmentSerializer.new(category_assessment)
+          serializer = Api::V2::GreenLanes::CategoryAssessmentSerializer.new(category_assessment, include: %w[geographical_area])
```
jebw, line 7: can this include excluded areas

jebw (changes requested):
Looks good but is either missing tests for excluded areas or the excluded areas segment itself is missing (I'd guess the former)

Thanks

app/controllers/api/v2/green_lanes/category_assessments_controller.rb
rasikasri: done

app/controllers/api/v2/green_lanes/goods_nomenclatures_controller.rb
rasikasri: done

.env.development
```diff
@@ -7,7 +7,7 @@ BULK_SEARCH_API_ENABLED=true
 DEFAULT_API_VERSION=2
 DIFFERENCES_TO_EMAILS=Online Trade Tariff Support <hmrc-trade-tariff-support-g@digital.hmrc.gov.uk>
 DUMP_CDS_DATA_AS_JSON=false
-EXCESS_QUERY_THRESHOLD=1000
+EXCESS_QUERY_THRESHOLD=2000
```
jebw, line 10: This was a suggested local variable change - this will change for all developers

Please use `.env.development.local`

jebw (commented):
Looks good, can you drop the env file change and just use a local one

Could you also tidy up the PR message, cheers - either remove the bits you don't need or correct them - kinda looks like this has destructive changes which I don't think is the case?

.env.development
rasikasri: Done

jebw (approved):

PR #1709, opened 2024-02-06
HOTT-4962 Added document describing our caching

jebw:
### Jira link

HOTT-4962

### What?

I have added/removed/altered:

- [x] Added a document describing how caching works within our app

### Why?

I am doing this because:

- It came up during a retro that there is limited information on this within the team

- Non, just a document


amberstarlight (approved):
Great writeup! We can put `[ci skip]` in the merge commit description to instruct CircleCi not do anything when this is merged (as we don't need to).

PR #1714, opened 2024-02-08
HOTT-4957 Separated and delayed generation of differences report

jebw:
### Jira link

HOTT-4957

### What?

I have added/removed/altered:

- [x] Separated the generation of Differences report from the other reports which it depends upon
- [x] Deferred the generation of the differences report to allow for reports from both environments to be completed
- [x] Added rake tasks to manually trigger the generation of reports

### Why?

I am doing this because:

- We have had the differences report fail, this allows it to retry
- This avoids it trying to create a differences report on UK before the source reports have finished generating on XI
- This allows manually trigger report generation on days other than monday

- Low, impacts generation of reports

### DIFF
--- a/app/workers/report_worker.rb
+++ b/app/workers/report_worker.rb
@@ -3,7 +3,7 @@ class ReportWorker
 
   sidekiq_options retry: false
 
-  def perform
+  def perform(trigger_differences_report = true)
     Reporting::Commodities.generate
     Reporting::Basic.generate
     Reporting::SupplementaryUnits.generate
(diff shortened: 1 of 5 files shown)


app/workers/report_worker.rb
```diff
@@ -3,29 +3,27 @@ class ReportWorker
...
-  end
-
-  def mail_differences?
+  def generate_differences?
     TradeTariffBackend.uk? && monday?
   end
 
-  def differences
-    Reporting::Differences.generate
+  def schedule_differences_generation
```
shjohnson, line 23: 👍

PR #1720, opened 2024-02-12
HOTT-4852: Make deployments depend on workers

amberstarlight:
### Jira link

[HOTT-4852](https://transformuk.atlassian.net/browse/HOTT-4852)

### What?

I have added/removed/altered:

- Added explicit `depends_on` to API and admin APIs on the worker services.

### Why?

I am doing this because:

- The current problem statement is that we need to ensure that database migrations have applied (or failed) before attempting to launch the applications that rely on these migrations having happened.

  Our current approach to deploying the backend is that we deploy three separate sets of services:
  
  1: An API backend, with read-only access to the database.
  2: A worker, with init container which runs database migrations.
  3: An admin backend with write access to the database.

  These services start, run, and are updated in parallel. This can lead to a condition where a database migration required for new, changed, or removed code in the API could have failed to complete - but the API containers are still attempting to start.

- By adding an explicit `depends_on` argument to the APIs, forcing them to depend on the worker's successful deployment, we can avoid this issue from occuring.

  After a discussion with @jebw on #1698, we agreed that this was a lower-risk approach to solving the problem.

### Notes

- This will inevitably make the deployment of the backend service longer.


jebw (dismissed):
Looks good _but_ AFAICT we need to know the worker service will have failed if its been rolled back so I think it needs the postcondition PR to be in place before this can get merged?

jebw (approved):
LGTM

PR #1721, opened 2024-02-12
HOTT-4957 Document reporting

jebw:
### Jira link

HOTT-4957

### What?

I have added/removed/altered:

- [x] Added a document explaining the daily reports

### Why?

I am doing this because:

- During the XI sync discussion it transpired out that this was not understood and lacked docs

- Non - adds a markdown file


shjohnson (approved):
LGTM to me, i was going to say about adding more detail on times it runs or the full name of reporting bucket but i think this is a good improvement of not having anything currently

jebw:
@shjohnson Cheers for the reason, I excluded them because that info tends to get outdated where as just telling the dev its scheduled should mean they know to look in sidekiq.yml :+1:

PR #1724, opened 2024-02-14
HOTT-5114 add measures to category assessments

jebw:
### Jira link

HOTT-5114

### What?

I have added/removed/altered:

- [x] Reworked to FindCategoryAssessmentService to also return matching measures
- [x] Added a serialization of the matching measures
- [x] Added a serialization of the measure groups
- [x] Added a serialization of the footnotes for those measures

### Why?

I am doing this because:

- Our proposed API needs to serialize out the measures relating to CategoryAssessments

- Low, private (for now) API

### DIFF
--- a/app/models/green_lanes/category_assessment.rb
+++ b/app/models/green_lanes/category_assessment.rb
@@ -50,7 +50,7 @@ module GreenLanes
 
       def load_from_string(data)
         json_array = JSON.parse(data)
-        @all = json_array.map { |json| new(json) }
+        @all = json_array.map { |json| new(json).tap(&:id).tap(&:freeze) }
       end
 
       def all
--- a/app/services/green_lanes/find_category_assessments_service.rb
+++ b/app/services/green_lanes/find_category_assessments_service.rb
@@ -2,11 +2,11 @@ module GreenLanes
   class FindCategoryAssessmentsService
     class << self
       def call(goods_nomenclature:, geographical_area_id: nil)
-        goods_nomenclature.applicable_measures.flat_map do |measure|
+        goods_nomenclature.applicable_measures.flat_map { |measure|
           CategoryAssessment.filter(regulation_id: measure.measure_generating_regulation_id,
                                     measure_type_id: measure.measure_type_id,
                                     geographical_area: geographical_area_id)
-        end
+        }.uniq
       end
     end
   end
--- a/app/services/green_lanes/find_category_assessments_service.rb
+++ b/app/services/green_lanes/find_category_assessments_service.rb
@@ -6,7 +6,7 @@ module GreenLanes
           CategoryAssessment.filter(regulation_id: measure.measure_generating_regulation_id,
                                     measure_type_id: measure.measure_type_id,
                                     geographical_area: geographical_area_id)
-        }.uniq
+        }.uniq(&:id)
       end
     end
   end
--- a/app/services/green_lanes/find_category_assessments_service.rb
+++ b/app/services/green_lanes/find_category_assessments_service.rb
@@ -9,7 +9,9 @@ module GreenLanes
                                        geographical_area: geographical_area_id)
           end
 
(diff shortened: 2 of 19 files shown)


app/models/green_lanes/category_assessment.rb
```diff
@@ -50,7 +50,7 @@ def load_from_file(file = DEFAULT_JSON)
 
       def load_from_string(data)
         json_array = JSON.parse(data)
-        @all = json_array.map { |json| new(json) }
+        @all = json_array.map { |json| new(json).tap(&:id).tap(&:freeze) }
```
alexdesi, line 53: so, to check, this is returning the same object as before,
but it freezes the IDs so they cannot be changed?

why do we need it?

app/services/green_lanes/find_category_assessments_service.rb
```diff
@@ -2,11 +2,15 @@ module GreenLanes
...
-        end
+        CategoryAssessment.all.map { |category_assessment|
+          matches = goods_nomenclature.applicable_measures.select do |measure|
+            category_assessment.match?(regulation_id: measure.measure_generating_regulation_id,
+                                       measure_type_id: measure.measure_type_id,
+                                       geographical_area: geographical_area_id)
+          end
+
+          matches.any? ? [category_assessment, matches] : nil
+        }.compact
```
alexdesi, line 13: it looks a bit complex,
I was thinking if we split this in two sections:
1. select the categories
2.  add the category assessment (map)   

it would be simpler

spec/models/green_lanes/category_assessment_spec.rb
```diff
@@ -74,6 +74,7 @@
 
       it { is_expected.to be_an Array }
       it { is_expected.to all be_instance_of described_class }
+      it { is_expected.to all be_frozen }
```
alexdesi, line 77: This makes me think the second `tap` works on the categories, not the IDs.
but why do we need to freeze them?

spec/serializers/api/v2/green_lanes/goods_nomenclature_serializer_spec.rb
```diff
@@ -1,19 +1,23 @@
 RSpec.describe Api::V2::GreenLanes::GoodsNomenclatureSerializer do
   subject(:serialized) do
-    described_class.new(gn_presenter, include: %w[applicable_measures applicable_category_assessments applicable_category_assessments.geographical_area]).serializable_hash
+    assessments_with_measures = [[category_assessment, [first_measure]]]
+
+    gn_presenter = \
+      Api::V2::GreenLanes::GoodsNomenclaturePresenter.new(subheading,
+                                                          assessments_with_measures)
```
alexdesi, line 7: a bit of strange indentation here :smile_cat:

alexdesi (commented):
The logic looks great.
The code is becoming more complex in a couple of parts.

It would be great to simplify those parts.

app/models/green_lanes/category_assessment.rb
jebw: The ContentAddressableId module stores the id in `@id` after generating it -> can't do this on a frozen class so we force generation prior to freezing

spec/models/green_lanes/category_assessment_spec.rb
jebw: They are not intended to be mutatable data - so this avoids developers mutating them by mistake and introducing hard to undestand bugs

Took 5 minutes to add but could save us a load of debugging in the future

spec/serializers/api/v2/green_lanes/goods_nomenclature_serializer_spec.rb
jebw: Really? I've seen it lots in code bases and is actually what the govuk'ised rubocop recommends to use?

alexdesi (approved):
nice one :rocket:  :white_check_mark:

PR #1727, opened 2024-02-16
HOTT-5144 Renamed category assessment model

jebw:
### Jira link

HOTT-5114

### What?

I have added/removed/altered:

- [x] Renamed the CategoryAssessment model to CategoryAssessmentJson
- [x] Renamed the category_assessment factory to category_assessment_json
- [x] Changed other code to use the new models/factory names
- [x] Refactored a spec that was overwriting the stateful data on the CategoryAsssesmentJson model
- [x] Renamed the category_assessments_controller_controller_spec

### Why?

I am doing this because:

- This free's up the name for the Sequel ORM backed model
- It free's up the factory name for a future orm backed factory
- Removing the modification of stateful data removes a possible source of inconsistency in specs
- The request spec for the controller was incorrectly named

- Low, no external impact on APIs


alexdesi (approved):
It all makes sense here! :man_dancing:

PR #1728, opened 2024-02-16
HOTT-4996 import green lanes themes data

jebw:
### Jira link

HOTT-4996

### What?

I have added/removed/altered:

- [x] Added a GreenLanes::Theme model
- [x] Added a rake task and (limited subset) data file to import themes

### Why?

I am doing this because:

- We will require this data as part of our db backed implementation of the API

- Low


lib/tasks/green_lanes.rake
```diff
@@ -22,4 +22,34 @@ namespace :green_lanes do
...
+        theme = theme.strip
+        section = section.strip.gsub(/^(\d+\.\d+)\./, '\1')
+        section, subsection = section.split('.').map(&:to_i)
+
+        instance = @existing_themes[[section, subsection]]
+        instance ||= GreenLanes::Theme.new(section:, subsection:)
+        instance.theme = theme
+        instance.description = row['Full description']
+        instance.category = row['Category implied']
+        if instance.valid?
```
alexdesi, line 47: instead of checking `valid?`,  we can just: 
`save(raise_on_failure: true)`

without even rescuing the error,
but in case you want:

```
rescue Sequel::ValidationFailed => e
  raise "Import error! #{instance.inspect}"
  ```

**Review by alexdesi — DISMISSED** [review]
Approved. 
I added a minor comment to remove a conditional.

File: lib/tasks/green_lanes.rake [hunk 2]
```diff
@@ -22,4 +22,30 @@ namespace :green_lanes do
...
+        theme = theme.strip
+        section = section.strip.gsub(/^(\d+\.\d+)\./, '\1')
+        section, subsection = section.split('.').map(&:to_i)
+
+        instance = @existing_themes[[section, subsection]]
+        instance ||= GreenLanes::Theme.new(section:, subsection:)
+        instance.theme = theme
+        instance.description = row['Full description']
+        instance.category = row['Category implied']
+        instance.save(raise_on_failure: true)
```
alexdesi, line 47: :rocket:

alexdesi (dismissed):
:white_check_mark:

lib/tasks/green_lanes.rake
jebw: good suggestion!

rasikasri (approved):

PR #1729, opened 2024-02-20
BAU: Remove init container config from API app

amberstarlight:
### What?

I have added/removed/altered:

- Removed `init_container` config from `backend_xi`.

### Why?

I am doing this because:

- It has no effect but produces errors because the database connection string is read-only.


jebw (approved):
LGTM

PR #1752, opened 2024-03-07
HOTT-5215: Add filtering the Goods Nomenclature page by geo area

rasikasri:
### Jira link

[HOTT-5215](https://transformuk.atlassian.net/browse/HOTT-5215)

### What?

I have added/removed/altered:

- [ ] Added empty geographical area filter into category_assessment model
- [ ] Removed baz

### Why?

I am doing this because:

- When front end sends api request to green lanes api, it sends country='' paramter for default All Country option in country drop down.

### Have you? (optional)

### DIFF
--- a/app/models/green_lanes/category_assessment_json.rb
+++ b/app/models/green_lanes/category_assessment_json.rb
@@ -69,6 +69,7 @@ module GreenLanes
         measure_type_id == self.measure_type_id &&
         (geographical_area == geographical_area_id ||
           geographical_area.nil? ||
+          geographical_area.empty? ||
           geographical_area_id == GeographicalArea::ERGA_OMNES_ID)
     end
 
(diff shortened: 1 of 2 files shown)


app/models/green_lanes/category_assessment_json.rb
```diff
@@ -69,6 +69,7 @@ def match?(regulation_id:, measure_type_id:, geographical_area: nil)
         measure_type_id == self.measure_type_id &&
         (geographical_area == geographical_area_id ||
           geographical_area.nil? ||
+          geographical_area.empty? ||
```
jebw, line 72: These two should be replaced with `.blank?` which checks for nils, empty string and whitespace only strings

jebw (changes requested):
Good spot, really minor suggestion of improving the check

app/models/green_lanes/category_assessment_json.rb
rasikasri: done

jebw (approved):

jebw (approved):
LGTM

PR #1780, opened 2024-03-18
GL-105: Add backend admin API to list all category assessment records

rasikasri:
### Jira link

[GL-105](https://transformuk.atlassian.net/browse/GL-105)

### What?

I have added/removed/altered:

- [ ] Added admin api for category assessments

### Why?

I am doing this because:

- we need to build admin screens to manage category assessments
-
-

### Have you? (optional)


jebw (approved):
LGTM

config/routes.rb
```diff
@@ -62,6 +62,8 @@
 
         resources :news_items, only: %i[index show create update destroy],
                                controller: 'news/items'
+
+        resources :category_assessments, module: 'green_lanes', only: %i[index]
```
jebw, line 66: We shouldn't be exposing green lanes admin API for uk only? If anything I think it should be XI only sicnce that is where the data is

jebw (changes requested):
Needs to be XI only, not UK only?

jebw (approved):
LGTM

PR #1784, opened 2024-03-20
BAU: Round the condition amount for the excise ASV measure conditions

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Added handling for floating point arithmetic issue
- [x] Added test coverage for this

### Why?

```ruby
0.145 * 100 => 14.499999999999998
(0.145 * 100).round(2) => 14.5
```

Floating point numbers in many programming languages, including Ruby, are represented in a binary format that can lead to precision errors for certain decimal numbers. This is because some decimals do not have an exact binary representation. For example, the decimal 0.145 cannot be precisely represented in binary; it's stored as a close approximation instead. When we perform arithmetic operations on these approximations, like multiplying 0.145 by 100, the small inaccuracies in their binary representation can result in unexpected results, such as 14.499999999999998 instead of 14.5.

To address this, we can round the result to a specific number of decimal places using Ruby's round method. This method adjusts the floating point number to the nearest value with the specified number of digits after the decimal point, providing a more accurate and expected representation of the value for our use case.


jebw (approved):
LGTM

PR #1795, opened 2024-04-03
GL-296: Hide Cat 3 CAs

rasikasri:
### Jira link

[GL-296](https://transformuk.atlassian.net/browse/GL-296)

### What?

I have added/removed/altered:

- [ ] Added a filter to filter out category 3 assessments from goods_nomenclature response

### Why?

I am doing this because:

- We only need to show category 1 and 2 in categorisation response


jebw (approved):
LGTM

PR #1800, opened 2024-04-08
OTT-??? Feature flag admin APIs

jebw:
### Jira link

OTT-???

### What?

I have added/removed/altered:

- [x] Moved the admin APIs behind a feature flag
- [x] Enable flag on the admin backends

### Why?

I am doing this because:

- These routes are inaccessible and so redundant on the other backends

- Should be low


willfish (approved):
Love it

PR #1807, opened 2024-04-10
GL-107: Add Admin API for listing all themes

rasikasri:
### Jira link

[GL-107](https://transformuk.atlassian.net/browse/GL-107)

### What?

I have added/removed/altered:

- [x] Added a GET admin api to list out all themes from the database 

### Why?

I am doing this because:

- This will enable a theme drop down in the category assessments admin area

### Have you? (optional)


app/controllers/api/admin/green_lanes/themes_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+        before_action :check_service, :authenticate_user!
+
+        def index
+          render json: serialize(themes.to_a)
+        end
+
+        private
+
+        def themes
+          @themes ||= ::GreenLanes::Theme.order(Sequel.desc(:id))
```
jebw, line 14: This would make more sense as section and subsection

app/controllers/api/admin/green_lanes/themes_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+        end
+
+        def serialize(*args)
+          Api::Admin::GreenLanes::ThemeSerializer.new(*args).serializable_hash
+        end
+
+        def serialize_errors(theme)
+          Api::Admin::ErrorSerializationService.new(theme).call
+        end
+
```
jebw, line 24: AFAICT this method is unused - I'm surprised rubocop didn't complain about it?

app/serializers/api/admin/green_lanes/theme_serializer.rb
```diff
@@ -0,0 +1,21 @@
...
+
+        set_id :id
+
+        attributes :section,
+                   :subsection,
+                   :theme,
+                   :description,
+                   :category,
+                   :created_at,
+                   :updated_at
```
jebw, line 17: Not convinced we need to be serializing these two? No strong feelings though

config/routes.rb
```diff
@@ -66,6 +66,7 @@
         end
 
         resources :category_assessments, module: 'green_lanes', only: %i[index show create update destroy]
+        resources :themes, module: 'green_lanes', only: %i[index]
```
jebw, line 69: Both these routes could be moved under a nested namespace (like news)

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
        - hmrc-postgres13:/var/lib/postgresql/data
   hmrc-opensearch:
     container_name: hmrc-opensearch
-    image: opensearchproject/opensearch:2
+    image: opensearchproject/opensearch:2.11.0
```
jebw, line 25: I'd say either leave this out, or move it to its own PR since its not relevant to this PR

In general I'm inclined to say leave it as is, which forces someone to solve the env var question properly rather than just locking us to an outdated version

spec/requests/api/admin/green_lanes/themes_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  describe 'GET to #index' do
+    let(:make_request) do
+      authenticated_get api_admin_themes_path(format: :json)
+    end
+
+    context 'with some themes' do
+      before do
+        theme
+      end
```
jebw, line 19: super minor but this is so short it could just be a one liner `before { theme }` ? no strong feelings either way on this

jebw (changes requested):
Couple of minor suggestions and the compose file change can be dropped but looks great

app/controllers/api/admin/green_lanes/themes_controller.rb
rasikasri: done

app/controllers/api/admin/green_lanes/themes_controller.rb
rasikasri: removed

config/routes.rb
rasikasri: done

spec/requests/api/admin/green_lanes/themes_controller_spec.rb
rasikasri: done

jebw (approved):
LGTM :raised_hands:

bhavanibandla (approved):
Approved to merge this PR to the main branch.

PR #1821, opened 2024-04-15
OTT-139 End Date report added with new flag

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/OTT-139

### What?

I have altered:

- [x] app/lib/reporting/differences/goods_nomenclature_end_date.rb

### Why?

I am doing this because:

- The differences report requires a 'New' tab for issues that have occurred in the past 7 days.

image

image


willfish (approved):

willfish:
Let's chat about these @HWallenberg I think we can do this differently and achieve the outcome we want in one foul swoop

PR #1832, opened 2024-04-22
OTT-180 Use router to parse routes

jebw:
Incorrectly formatted requests now return a 404

### Jira link

OTT-180

### What?

I have added/removed/altered:

- [x] Moved the logic for parsing the exchange rates params out into the router

### Why?

I am doing this because:

- Invalid `:id`'s were being passed to the controller but then could not be parsed because they did not follow the expected format
- This lets the router deal with parsing path params

- Notable, this is a simple change but it affects a crucial service so needs to be thoroughly checked prior to deployment

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -100,7 +100,14 @@ Rails.application.routes.draw do
 
       namespace :exchange_rates do
         get 'period_lists(/:year)', to: 'period_lists#show', as: :period_list
-        resources :files, only: [:show]
+
+        get 'files/(:type)_:year-:month', to: 'files#show',
+                                          as: :file,
+                                          constraints: {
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -100,7 +100,14 @@
 
       namespace :exchange_rates do
         get 'period_lists(/:year)', to: 'period_lists#show', as: :period_list
-        resources :files, only: [:show]
+
+        get 'files/(:type)_:year-:month', to: 'files#show',
+                                          as: :file,
+                                          constraints: {
+                                            type: ExchangeRateFile.type_check_regexp,
```
willfish, line 107: Sexy

willfish (approved):
I love it

PR #1841, opened 2024-04-30
OTT-248: Increase section title limit

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/OTT-248

### What?

I have added/removed/altered:

- [ ] Increase section title limit

### Why?

I am doing this because:

- We need to update a title that is more than 255 characters in length

### DIFF
--- a/db/migrate/20240429125446_change_title_limit_in_sections.rb
+++ b/db/migrate/20240429125446_change_title_limit_in_sections.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    run 'ALTER TABLE sections ALTER COLUMN title TYPE varchar(500)'
+  end
+
+  down do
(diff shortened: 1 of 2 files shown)


HWallenberg (dismissed):

jebw (changes requested):
This works but it might be nicer to use the migrations own DSL rather than raw SQL?

https://sequel.jeremyevans.net/rdoc/files/doc/schema_modification_rdoc.html#label-alter_table

More significantly this does not record the changes to `structure.sql`

maverickvi:
@jebw changes added to structure.sql. I'm skipping the nicer DSL change as I have already tested the raw version and the result would be the same so would like to save some time

willfish (approved):
Small preference for changing the type to TEXT (assuming this is easier) but otherwise this looks great to me

jebw (approved):
LGTM

PR #1844, opened 2024-04-30
GL-364: Add model for exemption overrides, possibly should 'relate' to certificates model

rasikasri:
### Jira link

[GL-364](https://transformuk.atlassian.net/browse/GL-364)

### What?

I have added/removed/altered:

- [ ] Added GreenLane::ExemptingCertificateOverride model
- [ ] Associate it with Certificate model

### Why?

I am doing this because:

- This model represents certificates that included exemption or excluded exemption, exception to its default behaviour


jebw (changes requested):
Looks good but can you add a relationship back from the certificate to the override class

When this is getting loaded in it'll be chaining through from the certificate to the override rather than vice versa

rasikasri:
> Looks good but can you add a relationship back from the certificate to the override class
> 
> When this is getting loaded in it'll be chaining through from the certificate to the override rather than vice versa

I added the relationship now

jebw (dismissed):
LGTM

alexdesi (approved):
:white_check_mark:

PR #1850, opened 2024-05-03
GL-366: Add admin APIs to list out a page of overrides

rasikasri:
### Jira link

[GL-366](https://transformuk.atlassian.net/browse/GL-366)

### What?

I have added/removed/altered:

- [ ] Added admin APIs to list / create / edit / remove overrides

### Why?

I am doing this because:

- So that a screen in the Admin area can be built to manage exempting certificate overrides

### DIFF
--- a/app/controllers/api/admin/green_lanes/category_assessments_controller.rb
+++ b/app/controllers/api/admin/green_lanes/category_assessments_controller.rb
@@ -3,6 +3,7 @@ module Api
     module GreenLanes
       class CategoryAssessmentsController < AdminController
         include Pageable
+        include XiOnly
 
         before_action :check_service, :authenticate_user!
 
@@ -74,12 +75,6 @@ module Api
 
         def serialize_errors(category_assessment)
           Api::Admin::ErrorSerializationService.new(category_assessment).call
-        end
-
-        def check_service
-          if TradeTariffBackend.uk?
-            raise ActionController::RoutingError, 'Invalid service'
-          end
         end
       end
     end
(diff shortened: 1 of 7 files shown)


app/controllers/api/admin/green_lanes/category_assessments_controller.rb
```diff
@@ -3,6 +3,7 @@ module Admin
     module GreenLanes
       class CategoryAssessmentsController < AdminController
         include Pageable
+        include XiOnly
```
jebw, line 6: :raised_hands:

app/controllers/api/admin/green_lanes/exempting_certificate_overrides_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+            render json: serialize(eco),
+                   location: api_admin_green_lanes_exempting_certificate_override_url(eco.id),
+                   status: :created
+          else
+            render json: serialize_errors(eco),
+                   status: :unprocessable_entity
+          end
+        end
+
+        def update
```
jebw, line 32: After refinement I don't think we need to support the `#update` so this action can be dropped

config/routes.rb
```diff
@@ -68,6 +68,7 @@
         namespace :green_lanes do
           resources :category_assessments, only: %i[index show create update destroy]
           resources :themes, only: %i[index]
+          resources :exempting_certificate_overrides, only: %i[index show create update destroy]
```
jebw, line 71: `#update` can get dropped

jebw (changes requested):
Looks good and great to see the XI checks split out

Only query is around the update which I think becomes irrelevant now we don't require a `behaviour` attribute

app/controllers/api/admin/green_lanes/exempting_certificate_overrides_controller.rb
rasikasri: Done

jebw (approved):
LGTM

PR #1864, opened 2024-05-09
GL-365: Use exempting_certificate_override table to control whether a certificate is treated as an exemption

rasikasri:
### Jira link

[GL-365](https://transformuk.atlassian.net/browse/GL-365)

### What?

I have added/removed/altered:

- [ ] Changed GL permutation calculator to check exempting_certificate_override table to detemine if certificate is an exemption
- [ ] Changed GL Category Assessment presenter to check exempting_certificate_override to load exemption certificates

### Why?

I am doing this because:

- We want to override the default exemption behaviour of certificates.


jebw (approved):
LGTM

PR #1868, opened 2024-05-15
GL-369: Add admin API to create and edit pseudo exemptions

rasikasri:
### Jira link

[GL-369](https://transformuk.atlassian.net/browse/GL-369)

### What?

I have added/removed/altered:

- [ ] Added admin api endpoint to create and update green lanes exemptions

### Why?

I am doing this because:

- User should be able to create abd update exemptions in admin UI


jebw (approved):
LGTM

PR #1872, opened 2024-05-21
GL-361: Add admin API to list out pseudo measures

rasikasri:
### Jira link

[GL-361](https://transformuk.atlassian.net/browse/GL-361)

### What?

I have added/removed/altered:

- [ ] Added admin api endpoint to list green lanes measurs
- [ ] Updated admin category_assessment api response to include them information

### Why?

I am doing this because:

- User should be able to manage measures in admin UI

### DIFF
--- a/app/models/green_lanes/measure.rb
+++ b/app/models/green_lanes/measure.rb
@@ -7,8 +7,6 @@ module GreenLanes
     many_to_one :goods_nomenclature, class: 'GoodsNomenclature',
                                      primary_key: %i[goods_nomenclature_item_id producline_suffix],
                                      key: %i[goods_nomenclature_item_id productline_suffix]
-    def goods_nomenclature_id
-      goods_nomenclature_item_id
-    end
+    delegate :goods_nomenclature_sid, to: :goods_nomenclature
   end
 end
--- a/app/models/green_lanes/measure.rb
+++ b/app/models/green_lanes/measure.rb
@@ -7,5 +7,8 @@ module GreenLanes
     many_to_one :goods_nomenclature, class: 'GoodsNomenclature',
                                      primary_key: %i[goods_nomenclature_item_id producline_suffix],
                                      key: %i[goods_nomenclature_item_id productline_suffix]
+    def goods_nomenclature_id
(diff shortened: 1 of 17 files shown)


app/models/green_lanes/measure.rb
```diff
@@ -7,5 +7,8 @@ class Measure < Sequel::Model(:green_lanes_measures)
     many_to_one :goods_nomenclature, class: 'GoodsNomenclature',
                                      primary_key: %i[goods_nomenclature_item_id producline_suffix],
                                      key: %i[goods_nomenclature_item_id productline_suffix]
+    def goods_nomenclature_id
+      goods_nomenclature_item_id
+    end
```
jebw, line 12: I cannot see this used anywhere so can this be removed I think?

jebw (changes requested):
Looks good - there just appears to be one orphaned method?

app/models/green_lanes/measure.rb
rasikasri: This is checked by - app/serializers/api/admin/green_lanes/measure_serializer.rb

`has_one :goods_nomenclature, serializer: GoodsNomenclatureSerializer`

I think that's the convention to map this association as the name is goods_nomenclature. If the name is goods_nomenclature_item, then this wouldn't have been an issue.

jebw: Ah, now I understand - I think you'll want to use a uniq id for the relationship, item id isn't uniq at a given point in time but sid is

`has_one :goods_nomenclature, serializer: GoodsNomenclatureSerializer, id_method_name: :goods_nomenclature_sid`

You'll also want to set the id on GoodsNomenclatureSerializer to be `goods_nomenclature_sid` instead of `goods_nomenclature_item_id`

jebw: https://github.com/jsonapi-serializer/jsonapi-serializer?tab=readme-ov-file#customizable-options

rasikasri: Thanks, changed the id to `goods_nomenclature_sid `

jebw (approved):
:+1:

PR #1877, opened 2024-05-28
Gl 496 ca show api pseudo measure

rasikasri:
### Jira link

[GL-496](https://transformuk.atlassian.net/browse/GL-496)

### What?

I have added/removed/altered:

- [ ] Altered GL CA api show endpoint to include green lanes measures and exemptions

### Why?

I am doing this because:

- In Admin UI, user should be able to associate measures and exemption with category assessments


app/models/green_lanes/category_assessment.rb
```diff
@@ -3,6 +3,7 @@ class CategoryAssessment < Sequel::Model(:green_lanes_category_assessments)
     plugin :timestamps, update_on_create: true
     plugin :auto_validations, not_null: :presence
     plugin :association_pks
+    plugin :association_dependencies
```
jebw, line 6: :raised_hands: Good find!

app/presenters/api/admin/green_lanes/category_assessment_presenter.rb
```diff
@@ -0,0 +1,15 @@
...
+          green_lanes_measures.map(&:id)
+        end
+
+        def exemption_ids
+          exemptions.map(&:id)
+        end
+      end
+    end
+  end
+end
```
jebw, line 15: Is this needed? Doesn't the `association_pks` plugin give you these? That was why I'd included that originally?

jebw (approved):
LGTM but I don't think the presenter is needed - at first glance its replication what `association_pks` does?

app/presenters/api/admin/green_lanes/category_assessment_presenter.rb
rasikasri: OK, `association_pks` plugin adds id method with `_pks` sufix not `_ids` suffix. This can be done with `id_method_name` added to the serializer. This is done now and I removed the presenter now.

jebw (approved):
LGTM

PR #1878, opened 2024-05-28
OTT-264 Origin Tab Update UX

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/OTT-264

### What?

I have removed

- [x] Check rules of origin links. 

### Why?

I am doing this because:

- This is no longer required following UX changes of Origin tab. The link was used at the bottom of the tab replaced by 'Further information on trade with `[_country_]`'
- 
Deployment risks
Requires frontend PR https://github.com/trade-tariff/trade-tariff-frontend/pull/1921

image

image

### DIFF
--- a/spec/models/rules_of_origin/scheme_set_spec.rb
+++ b/spec/models/rules_of_origin/scheme_set_spec.rb
@@ -99,14 +99,6 @@ RSpec.describe RulesOfOrigin::SchemeSet do
     end
   end
 
-  describe '#links' do
-    subject { scheme_set.links }
-
-    it { is_expected.to have_attributes length: 1 }
-    it { is_expected.to all be_instance_of RulesOfOrigin::Link }
-    it { is_expected.to all have_attributes source: 'scheme_set' }
-  end
-
   describe '#read_referenced_file' do
     subject :read_file do
       scheme_set.read_referenced_file('fta_intro', file_name)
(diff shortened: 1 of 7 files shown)


spec/models/rules_of_origin/scheme_set_spec.rb
```diff
@@ -99,14 +99,6 @@
     end
   end
 
-  describe '#links' do
```
willfish, line 102: So my understanding is that we still have a links method but in our uk schemes file this is now empty?

Do we want to remove the corresponding method since it's no longer in use or have I misunderstood this?

HWallenberg: @willfish Only one of the links that is at 'UK' level was removed from the UK Backend json file but there are other links that are iterated over in the page (that are FTA specific) so we still need the method.

Hope this makes sense. :) 

image

willfish: Ah okay, so we have links for specific schemes in the scheme set and a links attribute of the scheme set itself. I'm guessing that these are all merged together at some point. I wonder if it's worth keeping the test and just saying its empty?

```  describe '#links' do
    subject { scheme_set.links }

    it { is_expected.to eq [] }
  end
```

willfish: Not sure if there's value to that, tbh

willfish (approved):

willfish (approved):

PR #1888, opened 2024-06-06
GL-551 Small tweaks to API

jebw:
### Jira link

GL-551

### What?

I have added/removed/altered:

- [x] Include associated entities for descendant Category Assessments

### Why?

I am doing this because:

- These were missed by mistake

- Low

### DIFF
--- a/app/controllers/api/v2/green_lanes/goods_nomenclatures_controller.rb
+++ b/app/controllers/api/v2/green_lanes/goods_nomenclatures_controller.rb
@@ -31,6 +31,8 @@ module Api
               applicable_category_assessments.measures.footnotes
               applicable_category_assessments.theme
               descendant_category_assessments
+              descendant_category_assessments.regulation
+              descendant_category_assessments.measure_type
               descendant_category_assessments.exemptions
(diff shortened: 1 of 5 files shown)


app/models/green_lanes/measure.rb
```diff
@@ -7,7 +7,7 @@ class Measure < Sequel::Model(:green_lanes_measures)
     many_to_one :goods_nomenclature, class: 'GoodsNomenclature',
                                      primary_key: %i[goods_nomenclature_item_id producline_suffix],
                                      key: %i[goods_nomenclature_item_id productline_suffix]
-    delegate :goods_nomenclature_sid, to: :goods_nomenclature
+    delegate :goods_nomenclature_sid, to: :goods_nomenclature, allow_nil: true
```
rasikasri, line 10: How this could be nullable as goods_nomenclature_sid is primary key of the [goods_nomenclature] and part of the composite key of green_lanes_measure

jebw: Its for when the goods nomenclature is nil but this is more thoroughly handled in another PR so I'll just drop this from here

PR #1892, opened 2024-06-10
GL-574 Handle pseudo measures against expired goods nomenclatures

jebw:
### Jira link

GL-574

### What?

I have added/removed/altered:

- [x] Use TimeMachine for GoodsNomenclature relationship on pseudo measures
- [x] Exclude pseudo measures against expired GoodsNomenclatures in APIs
- [x] Fixed `reciprocal` configuration for pseudo measures relationship

### Why?

I am doing this because:

- We should ignore measures (and their associated CAs) when the GN they attach to is not valid for the requested date

- Low, only affects GL APIs


rasikasri (dismissed):

alexdesi (approved):
:white_check_mark:

PR #1894, opened 2024-06-11
BAU: Fixes borked migration

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Fixes migration by removing the explicit schema

### Why?

I am doing this because:

- This is required to make sure the broken data files can be loaded
- We rely on implicit schema choices set by hierarchical search paths as part of our migration process.


jebw (approved):
LGTM

PR #1895, opened 2024-06-11
GL-601 Fix priorisation of regulations for category assessment imports

jebw:
### Jira link

GL-601

### What?

I have added/removed/altered:

- [x] Changed priorisation of Regulation Lookup

### Why?

I am doing this because:

- We should favour ModificationRegulations over BaseRegulations

- Low, means we'll be importing slightly different data into production than staging


alexdesi (dismissed):
It seems that the modification has priority on the base regulation.
It makes sense :+1:

alexdesi (approved):
:white_check_mark:

PR #1896, opened 2024-06-11
GL-562 Uniq certificates to avoid duplicate certs

jebw:
### Jira link

GL-562

### What?

I have added/removed/altered:

- [x] Remove list of duplicate certificates from the exemptions list

### Why?

I am doing this because:

- We were getting duplicate certificates when there was duplicate measure conditions pointing to the same certificate

- Low, small bugfix


alexdesi (dismissed):
:white_check_mark:

alexdesi (approved):
thanks for the changes! :white_check_mark:

PR #1899, opened 2024-06-18
Error in previous data migration file fix

shjohnson:
### Jira link

OTT-309

### What?

There was a slight mistake in the previous migration file that was missed so this is to fix that


alexdesi (approved):
:white_check_mark:

PR #1902, opened 2024-07-02
GL-705: Remove add code from permutation key

rasikasri:
### Jira link

[GL-705](https://transformuk.atlassian.net/browse/GL-705)

### What?

I have added/removed/altered:

- [ ] Removed additional code key from the permutation key

### Why?

I am doing this because:

- We don't need to create separate CA for each additional codes. Instead, we need to group them into a one CA

### DIFF
--- a/app/presenters/api/v2/green_lanes/category_assessment_presenter.rb
+++ b/app/presenters/api/v2/green_lanes/category_assessment_presenter.rb
@@ -17,7 +17,6 @@ module Api
                  :geographical_area,
                  :excluded_geographical_areas,
                  :measure_conditions,
-                 :additional_code,
                  to: :first_measure
 
         content_addressable_fields do |ca|
@@ -77,7 +76,8 @@ module Api
         end
 
         def additional_codes
-          Array.wrap(additional_code)
+          additional_codes = measures.map(&:additional_code).compact
+          additional_codes.empty? ? [] : additional_codes
         end
 
         def pseudo_exemptions
--- a/app/services/green_lanes/find_category_assessments_service.rb
+++ b/app/services/green_lanes/find_category_assessments_service.rb
@@ -25,8 +25,8 @@ module GreenLanes
 
     def compute_assessment_permutations(assessment, assessment_measures)
       permutations_for_assessment(assessment_measures)
-        .map do |key, permutation|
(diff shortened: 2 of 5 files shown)


app/presenters/api/v2/green_lanes/category_assessment_presenter.rb
```diff
@@ -77,7 +76,8 @@ def certificates
         end
 
         def additional_codes
-          Array.wrap(additional_code)
+          additional_codes = measures.map(&:additional_code).compact
```
willfish, line 79: I'm guessing the answer is yes, but has this association for measures been eager loaded upstream of this call?

willfish:
My basic understanding of additional code is they basically decide whether the current measure is applicable (in most cases) or whether a certain duty rate applies.

So what we're saying here is that they're non-deterministic in terms of whether the current category  assessment abstraction applies for a given commodity.

So I guess this means that we're basically assuming that whether the additional code is relevant comes only after we ask the user for more information in the frontend UI and excludes specific measures, here?

app/presenters/api/v2/green_lanes/category_assessment_presenter.rb
rasikasri: Yes, here you can see additional_code is in the eager loading plan
https://github.com/trade-tariff/trade-tariff-backend/blob/GL-705_remove_add_code_from_permutation_key/app/services/green_lanes/fetch_goods_nomenclature_service.rb

rasikasri:
> My basic understanding of additional codes is they basically decide whether the current measure is applicable (in most cases) or whether a certain duty rate applies.
> 
> So what we're saying here is that they're non-deterministic in terms of whether the current category assessment abstraction applies for a given commodity essentially because we need more information.
> 
> So I guess this means that we're basically assuming that whether the additional code is relevant comes only after we ask the user for more information in the frontend UI and excludes specific measures, here?

We have multiple measures for same measure type with different additional_code attached to the measure record. So if we have additional_code in the permutation key, we gonna end up with number of category assessments for the same category per each additional code. Instead we want to have one CA aggigating all the additional_code into it. Then user can choose if there is a additional_code that applicable to the user and then exempt the category.

willfish:
Got it. Okay

PR #1904, opened 2024-07-03
GL-706: Add additional code exemption override schema

rasikasri:
### Jira link

[GL-706](https://transformuk.atlassian.net/browse/GL-706)

### What?

I have added/removed/altered:

- [ ] Added schema changes to add additional code override tables

### Why?

I am doing this because:

- We need to exclude all the additional codes other than few from the exemption list. This table will record the additional codes that are exemptions


shjohnson (dismissed):
LGTM

alexdesi (approved):
:white_check_mark:

PR #1905, opened 2024-07-03
GL-706: Add support to whitelist additional codes as exemptions in the BE

rasikasri:
### Jira link

[GL-706](https://transformuk.atlassian.net/browse/GL-706)

### What?

I have added/removed/altered:

- [ ] Added backend GN api changes to exclude all the additional codes other than the ones white listed

### Why?

I am doing this because:

- We need to exclude all the additional codes other than few from the exemption list. This table will record the additional codes that are exemptions


app/models/green_lanes/exempting_additional_code_override.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    one_to_one :reference_additional_code, class: :AdditionalCode,
+                                           key: %i[additional_code_type_id additional_code],
+                                           primary_key: %i[additional_code_type_id additional_code] do |ds|
+      ds.with_actual(AdditionalCode)
+    end
+
+  private
+
+    def after_create
```
shjohnson, line 14: Remind me the reason we are touching category assessments after any action takes place?

shjohnson (approved):
one question otherwise looks good

app/models/green_lanes/exempting_additional_code_override.rb
shjohnson: Refresh cache 👍  thanks @rasikasri

willfish (approved):

PR #1925, opened 2024-07-16
BAU: Bumps rails

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Fixed sidekiq compatibility with rails
- [x] Bumps rails
- [x] Bumps rubocop
- [x] Lints
- [x] Fixes some test file names and broken tests

### Why?

I am doing this because:

- This was failing locally for Hadleigh
- I'm in the process of moving us to newer versions of rails/ruby/rubocop anyway


app/lib/trade_tariff_backend.rb
```diff
@@ -148,7 +148,7 @@ def check_query_count?
     end
 
     def excess_query_threshold
-      @excess_query_threshold ||= ENV['EXCESS_QUERY_THRESHOLD'].presence&.to_i || 0
+      @excess_query_threshold ||= ENV['EXCESS_QUERY_THRESHOLD'].presence.to_i
```
willfish, line 151:
```ruby
[2] pry(main)> "".presence.to_i
=> 0
```

HWallenberg (approved):

PR #1926, opened 2024-07-16
GL-791: Green Lanes returns GN with the lowest PLS and indents number…

rasikasri:
### Jira link

[GL-791](https://transformuk.atlassian.net/browse/GL-791)

### What?

I have added/removed/altered:

- [ ] Altered the GN select criteria from default PLS 80 to GN with lowest PLS and lowest indent number

### Why?

I am doing this because:

- When that GN has multiple PLSs, We need to returns GN that sits highest in the nested list in green lanes api


shjohnson (approved):
LGTM

PR #1941, opened 2024-08-05
BAU: Include the category assessment id in the serialized output

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Fixes missing serialized attribute in live environments

### Why?

I am doing this because:

- This is needed by the frontend


alexdesi (approved):
:palm_tree:

PR #1946, opened 2024-08-08
GL-921: Change how CAs uniqueness is calculated to factor in theme

rasikasri:
### Jira link

[GL-921](https://transformuk.atlassian.net/browse/GL-921)

### What?

I have added/removed/altered:

- [ ] Updated unique index of category_assessment table add them_id into the index

### Why?

I am doing this because:

- We need to add two different CA for same measure_type and regulation quotable and non-quotable items. These CAs created for different category but enforced by same measure type and regulation.

### DIFF
--- a/lib/tasks/green_lanes.rake
+++ b/lib/tasks/green_lanes.rake
@@ -191,4 +191,54 @@ namespace :green_lanes do
       end
     end
   end
+
+  desc 'Add CategoryAssessments CATEGORY_ASSESSMENT_CSV_FILE=path/to/file.csv'
+  task import_csv_category_assessments: :environment do
+    raise 'Only supported on XI service' unless TradeTariffBackend.xi?
+
+    if Rails.application.config.persistence_bucket.present?
+      CATEGORY_ASSESSMENT_CSV_OBJECT_KEY = 'data/categorisation/category_assessment.csv'.freeze
+      begin
+        csv = Rails.application.config.persistence_bucket.object(CATEGORY_ASSESSMENT_CSV_OBJECT_KEY).get.body.read
+        @data = CSV.parse(csv, headers: true)
+      rescue Aws::S3::Errors::NoSuchKey => e
+        raise InvalidFile, "File not found in S3 (#{e.message})"
+      end
+    else
+      raise "Cannot read file '#{ENV['CATEGORY_ASSESSMENT_CSV_FILE']}'" unless File.file?(ENV['CATEGORY_ASSESSMENT_CSV_FILE'].to_s)
+
+      @data = CSV.read(ENV['CATEGORY_ASSESSMENT_CSV_FILE'], headers: true)
+    end
+
+    @data.map do |row|
+      theme = GreenLanes::Theme.where(section: row['theme_section'], subsection: row['theme_subsection']).first
+      assessments = GreenLanes::CategoryAssessment.all.index_by do |assessment|
+        [assessment.measure_type_id, assessment.regulation_id, assessment.theme_id]
+      end
+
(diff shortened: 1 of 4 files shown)


spec/models/green_lanes/category_assessment_spec.rb
```diff
@@ -22,16 +22,17 @@
     it { is_expected.to include regulation_role: ['is not present'] }
     it { is_expected.to include theme_id: ['is not present'] }
 
-    context 'with duplicate measure_type_id and regulation_id' do
+    context 'with duplicate measure_type_id, regulation_id and theme_id' do
```
shjohnson, line 25: I guess also the regulation role is the same, do we need to mention?

lib/tasks/green_lanes.rake
```diff
@@ -191,4 +191,53 @@ namespace :green_lanes do
       end
     end
   end
+
+  desc 'Add CategoryAssessments CATEGORY_ASSESSMENT_CSV_FILE=path/to/file.csv'
```
shjohnson, line 195: my only worry about long rake tasks like this is its quite a lot going on and whether breaking it down something like this might be more readable:

```
desc 'Add CategoryAssessments CATEGORY_ASSESSMENT_CSV_FILE=path/to/file.csv'
task import_csv_category_assessments: :environment do
  raise 'Only supported on XI service' unless TradeTariffBackend.xi?

  csv_data = fetch_csv_data
  parsed_data = CSV.parse(csv_data, headers: true)

  assessments = GreenLanes::CategoryAssessment.all.index_by do |assessment|
    [assessment.measure_type_id, assessment.regulation_id, assessment.theme_id]
  end

  parsed_data.each do |row|
    process_row(row, assessments)
  end
end

def fetch_csv_data
  if Rails.application.config.persistence_bucket.present?
    fetch_csv_from_s3
  else
    fetch_csv_from_local_file
  end
end

def fetch_csv_from_s3
  object_key = 'data/categorisation/category_assessment.csv'
  Rails.application.config.persistence_bucket.object(object_key).get.body.read
rescue Aws::S3::Errors::NoSuchKey => e
  raise InvalidFile, "File not found in S3 (#{e.message})"
end

def fetch_csv_from_local_file
  file_path = ENV['CATEGORY_ASSESSMENT_CSV_FILE'].to_s
  raise "Cannot read file '#{file_path}'" unless File.file?(file_path)

  File.read(file_path)
end

def process_row(row, assessments)
  theme = GreenLanes::Theme.find_by(section: row['theme_section'], subsection: row['theme_subsection'])
  key = [row['measure_type_id'], row['regulation_id'], theme.id]
  assessment = assessments[key] || initialize_new_assessment(row, theme)

  if assessment.id.nil?
    assessment.theme = theme
    assessment.save(validate: true)
    add_exemption_to_assessment(row['exemption_code'], assessment)
  end

  assessments[key] = assessment
end

def initialize_new_assessment(row, theme)
  GreenLanes::CategoryAssessment.new.tap do |assessment|
    assessment.measure_type_id = row['measure_type_id']
    assessment.regulation_id = row['regulation_id']
    assessment.regulation_role = row['regulation_role']
    assessment.theme = theme
  end
end

def add_exemption_to_assessment(exemption_code, assessment)
  exemption = GreenLanes::Exemption.find_by(code: exemption_code)
  assessment.add_exemption(exemption) if exemption
end

shjohnson: Fyi i havent tested the above hence why might be easier to explain on call tomorrow what going on?

willfish (approved):

shjohnson (approved):
🍏

PR #1949, opened 2024-08-13
GL-935: Extend API to allow categorization of 4 digit commodities that are end leaves

rasikasri:
### Jira link

[GL-935](https://transformuk.atlassian.net/browse/GL-935)

### What?

I have added/removed/altered:

- [x] Updated CC validation in GN api to allow 4 digit declarables.

### Why?

I am doing this because:

- As current validation block 4 digit declarables but they need to be allowed for categorisation
-
-

### Have you? (optional)

### DIFF
--- a/app/services/green_lanes/fetch_goods_nomenclature_service.rb
+++ b/app/services/green_lanes/fetch_goods_nomenclature_service.rb
@@ -68,7 +68,7 @@ module GreenLanes
     private
 
     def invalid_id(id)
-      id.blank? || !is_id_length_greater(id,2)
+      id.blank? || !is_id_length_greater(id, 2)
     end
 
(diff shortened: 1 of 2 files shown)


app/services/green_lanes/fetch_goods_nomenclature_service.rb
```diff
@@ -54,7 +54,7 @@ def call
              .order(Sequel[:goods_nomenclatures][:producline_suffix], Sequel[:goods_nomenclature_indents][:number_indents])
```
shjohnson, line 54: I know not related to this PR but would be good to extract some of the code out of the call method to improve readability and understanding.

shjohnson (approved):

PR #1957, opened 2024-08-20
BAU: Switch off GL update email on staging and dev environment

rasikasri:
### Jira link

BAU

### What?

I have added/removed/altered:

- [ ] Disabled Green Lanes Update Notifications job in dev and staging

### Why?

I am doing this because:

- We only need this in prod env


shjohnson (approved):
🚀

PR #1961, opened 2024-08-28
Allow development to use localstack over AWS

neilmiddleton:
### Jira link

BAU / Experiment.  

### What?

This PR allows local development to use [localstack](https://github.com/localstack/localstack) over AWS for development purposes.  This allows developers to work offline, or seperated from other developers in a non-shared environment.

Note this PR only covers S3 usage rather than other services.

Fails-safe to the current way of working so localstack is an opt-in change

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -31,3 +31,9 @@ XE_API_PASSWORD=pass
 XE_API_URL=url
 XE_API_USERNAME=user
 GREEN_LANES_UPDATE_EMAIL=hmrc@example.com
+## LOCALSTACK SETUP
+# Uncomment the following lines to use localstack
+# AWS_ENDPOINT=http://s3.localhost.localstack.cloud:4566
+# AWS_REGION=eu-west-2
+# AWS_SECRET_ACCESS_KEY=test
+# AWS_ACCESS_KEY_ID=test
--- a/README.md
+++ b/README.md
@@ -84,6 +84,12 @@ TARIFF_SYNC_USERNAME
 GREEN_LANES_UPDATE_EMAIL
 ```
 
+## Localstack
+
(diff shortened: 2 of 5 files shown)


willfish:
Having used localstack a bunch in the developer hub, I'm wondering if we can add a couple of Make targets to get started? And maybe a docker-compose.yml file?

```make
localstack: clean-localstack
	$(eval USAGE_PLAN_ID=$(shell docker-compose up | grep -m1 'usagePlanId=' | awk -F'=' '{print $$2}'))
	@if [ -z "$(USAGE_PLAN_ID)" ]; then \
		echo "Failed to extract USAGE_PLAN_ID"; \
		exit 1; \
	fi
	@sed -i'' -e 's/export USAGE_PLAN_ID=.*/export USAGE_PLAN_ID=$(USAGE_PLAN_ID)/' .env.development
	@echo "Updated USAGE_PLAN_ID in .env.development to $(USAGE_PLAN_ID)"
	@echo "Localstack is running. Run make stop-localstack to stop it."

clean-localstack: stop-localstack
	docker-compose rm -f -s -v

stop-localstack:
	docker-compose down
```

```yaml
version: "2"
services:
  localstack:
    image: localstack/localstack:latest
    ports:
      - 4566:4566           # LocalStack edge port
      - 4510-4559:4510-4559 # maps various AWS services ports
    volumes:
      # Initialization Hook (see https://docs.localstack.cloud/references/init-hooks/)
      - .bin/seed.sh:/etc/localstack/init/ready.d/script.sh
```

neilmiddleton:
Yeah, adding localstack to the local docker-compose so there's essentially zero setup for people would be a good idea.

PR #1962, opened 2024-09-02
GL-919: Implement throttling GL api

rasikasri:
### Jira link

[GL-919](https://transformuk.atlassian.net/browse/GL-919)

### What?

I have added/removed/altered:

- [ ] Implemented throttleing API request base on api key

### Why?

I am doing this because:

- API need to be protected from heavy load, hence it needs a throtteling mechanism

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,31 +84,6 @@ TARIFF_SYNC_USERNAME
 GREEN_LANES_UPDATE_EMAIL
 ```
 
-### API Key Management
-To access GL endpoints in the production environment, clients must provide a valid API key. These API keys are 
-securely stored in AWS Secrets Manager under the secret name backend-green-lanes-api-keys as a JSON blob.
-
-When authorizing a new API client, you should update the JSON file in Secrets Manager by adding the client’s API key 
-along with their details and the appropriate throttling limits. This ensures that each client has the correct permissions 
-and rate limits when interacting with the GL endpoints. Throttling limit is set per number of hours specified in period.
-
-```
-{
-  "api_keys": {
-    "<secret>": {
-      "name": "Descartes Labs",
-      "description": "Descartes Labs integration key",
-      "client_id": "dev",
-      "client_contact": "example@hmrc.com",
-      "client_secret": "<secret>",
-      "t&c_accepted": true,
-      "limit": 100,
-      "period": 1
-    }
-  }
-}
-```
-
 ## Licence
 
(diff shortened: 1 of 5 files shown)


neilmiddleton (commented):
Are there tests coming?

README.md
```diff
@@ -84,6 +84,31 @@ TARIFF_SYNC_USERNAME
 GREEN_LANES_UPDATE_EMAIL
 ```
 
+### API Key Management
```
willfish, line 87: I don't think this will pass the markdown linting rules?

willfish: How do you feel about putting a separate doc under docs/green_lanes_api_keys.md or equivalent?

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    throttle('default', limit: 10, period: 1.hour) do |req|
+      # Apply to all requests with an API key not in the list
+      api_key = req.env['HTTP_X_API_KEY']
+      API_KEY_LIMITS.blank? || (!API_KEY_LIMITS.key?(api_key) && api_key.present?)
+    end
+  end
+else
+  logger.info 'Rack::Attack is disabled in Dev env.'
```
willfish, line 23: I didn't know `logger` was injected into the scope of an initialiser

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,24 @@
+if Rails.env.production?
+  class Rack::Attack
+    Rack::Attack.cache.store = ActiveSupport::Cache::RedisCacheStore.new(url: ENV['REDIS_URL'])
+
+    API_KEY_LIMITS = JSON.parse(ENV['GREEN_LANES_API_KEYS'])
```
willfish, line 5: I'm a little lost on how this controls just for /green_lanes/ scoped api calls?

config/initializers/rack_attack.rb
rasikasri: I double checked and logger is available there

config/initializers/rack_attack.rb
rasikasri: This is a good point, I added logic to only throttle `/green_lanes/goods_nomenclatures` endpoint. Also removed the default throttle method as it doesn't make sense. If the api-key does not exist in the json, the the request will be blocked anyway.

willfish (approved):
It would be good to do a performance benchmark with/without this enabled (e.g. on staging) before/after we merge this

PR #1964, opened 2024-09-03
OTT-124 Seperate differences data/display and add "New?" Column

neilmiddleton:
### Jira link

OTT-124 / OT-457

### What?

There is a lot here code wise, but primarily three changes:

1) Seperates the code which gets hold of the report data from the code concerned with generating the Excel spreadsheet
2) Adds a new column to each sheet "New?" and if the data has changed since the last record in the DB it'll render as "Yes".
    - Stores each generated sheet in the database in it's raw JSON form
    - Adds a "New Items" column to the overview tab which contains a formula enumerating how many new records on each sheet.
 3) Adds a check that the difference report has created logs in the last week and alerts in Slack if not.  This check runs 1am every Tuesday.

Aside from the "New?" column there are no functional changes to the report.  For review purposes, almost none of this code is new even if the diff would have you think otherwise.

Sample differences report:  [differences.xlsx](https://github.com/user-attachments/files/16851216/differences.xlsx)

Note, the data saved to the DB has no egress path so it will build up over time.  We may in future want to purge old data from here, but considering the space requirements I don't see a big need for this.

There is a database migration in this PR, but only adding a table with no other dependencies


db/migrate/20240830142019_add_differences_log.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    create_table :differences_logs do
+      primary_key :id
+      Date :date, null: false
```
willfish, line 7: Should we make this less ambigious and just add an audit timestamp of created_at/updated_at as datetimes?

neilmiddleton: The reason here is that I'm looking to detect logs for the same day which is much easier with just a date field rather than getting into messing about with ranges.

willfish (approved):
Seems reasonable to me. Have you ran this in a live environment?

db/migrate/20240830142019_add_differences_log.rb
willfish: Okay

neilmiddleton:
> Seems reasonable to me. Have you ran this in a live environment?

Not past dev at the moment.  As it happens I'm going to add alerting to this PR to simplify deployment so no mergey yet.

neilmiddleton:
Run check added (OT-457)

willfish (approved):

PR #1965, opened 2024-09-09
Add bin/dev to allow dev web+workers

neilmiddleton:
### Jira link

BAU: Dev cleanup

### What?

I have added `bin/dev` which will use `Procfile.dev` to spin up web + workers as an option.  Currently the README only demonstrates how to start Rails on it's own.

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,0 +1,3 @@
+web: bin/rails server -p 3000
+worker1: bundle exec sidekiq -C ./config/sidekiq.yml
+worker2: bundle exec sidekiq -C ./config/sidekiq.yml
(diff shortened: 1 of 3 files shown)


Procfile.dev
```diff
@@ -0,0 +1,2 @@
+web: bin/rails server -p 3000
```
willfish, line 1: The Procfile isn't actually used if you wanted to steal that.

I doubt we'll ever use it for live envs

willfish (approved):

Procfile.dev
neilmiddleton: I left that as this is purely a dev change, but another PR should remove it if it's not used.

PR #1969, opened 2024-09-10
OTT-502 Remove item counts and rely on Excel

neilmiddleton:
### Jira link

OTT-502

### What?

This removes the manual counting of items in the Excel spreadsheet, and replaces them with Excel functions meaning that the numbers will always correlate with the contents of each sheet.

### Why?

Previously it was wrong


willfish (dismissed):
Amazing. Some leftover local changes but lgtm

HWallenberg (approved):

PR #1970, opened 2024-09-10
GL-785: Enable notifications component to surface what data expired

rasikasri:
### Jira link

[GL-785](https://transformuk.atlassian.net/browse/GL-785)

### What?

I have added/removed/altered:

- [ ] Added updated and expired measures to green lanes update notifications

### Why?

I am doing this because:

- We need to let HMRC know about updates to take relevant actions
-
-


shjohnson (approved):
LGTM guessing its been fully tested in dev

PR #1971, opened 2024-09-11
OTT-326 error message encountered if user enters trade date of more than 6 digits

neilmiddleton:
### Jira link

OTT-326

### What?

Added a constraint around date params (as_of) that it must be inside 20 years either side of today

### Why?

Currently it's possible to pass a date from the frontend such as 202400000-01-01, which is a valid date, but one which is very far in the future.  We already have dates to check that the date is valid, but no check to determine if the date is _reasonable_.

Should the date be unreasonable, todays date is used (which is the same behaviour as an invalid date)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -24,7 +24,10 @@ class ApplicationController < ActionController::API
   private
 
   def actual_date
-    Date.parse(params[:as_of].to_s)
+    date = Date.parse(params[:as_of].to_s)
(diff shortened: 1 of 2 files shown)


HWallenberg (commented):
@neilmiddleton apologies if I'm wrong but am I right in thinking this has some incorrect files committed to the branch? 😬

neilmiddleton:
@HWallenberg whacked the inappropriate commits.

HWallenberg (approved):
lgtm 👍

PR #1972, opened 2024-09-13
BAU: Remove all implicit deletion behaviour (soft/hard)

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Removes all missing deletion behaviour

### Why?

I am doing this because:

- CDS are now moving to a change-only explicit deletion model where they'll produce change secondaries in _all_ situations where they want to delete something and not require us to either infer or make the current state of the records reflect the snapshot in the XML


neilmiddleton (approved):
🟢  but ideally I'd like someone else to also review

HWallenberg (approved):

PR #1973, opened 2024-09-13
GL-1052: Remove exclusion of quota measures for GL categorisation

rasikasri:
### Jira link

[GL-1052](https://transformuk.atlassian.net/browse/GL-1052)

### What?

I have added/removed/altered:

- [ ] Removed quota measure types from applicable measure exclution filter for GL

### Why?

I am doing this because:

- Quota measures are used in categorisation, so we should not exclude them from goods nomencatures' applicable measure list 
-
-


app/lib/trade_tariff_request.rb
```diff
@@ -0,0 +1,4 @@
+# app/models/current.rb
+class TradeTariffRequest < ActiveSupport::CurrentAttributes
+  attribute :green_lanes
```
willfish, line 3: Wow, TIL

app/lib/trade_tariff_request.rb
```diff
@@ -0,0 +1,4 @@
+# app/models/current.rb
```
willfish, line 1: Not needed?

willfish (approved):
Elegant. I _love_ it

alexdesi (approved):

PR #1975, opened 2024-09-16
BAU: Check declarable flag only for four digit item id.

rasikasri:
### What?

I have added/removed/altered:

- [ ] Changed fetch goods nomenclature service to check the gn is declarable only for four digit id

### Why?

I am doing this because:

- Only four digit item id need to checked for declarable.
- There is a performance impact of checking the declarable flag for all the item ids
-


shjohnson (approved):

willfish (approved):
Yes there is a performance impact! Good spot

PR #1980, opened 2024-09-19
OTT-270 remove duplicate text

neilmiddleton:
### Jira link

https://transformuk.atlassian.net/browse/OTT-270

### What?

Removed duplicate text on ROO for Canada (`/rules_of_origin/0603130000/CA/components_definition`)


willfish (approved):

HWallenberg (commented):
I'm not sure if we need this PR as it's a duplicate of this and neutral elements was replaced instead of deleted.

https://transformuk.atlassian.net/browse/OTT-273
https://github.com/trade-tariff/trade-tariff-backend/pull/1979

neilmiddleton:
> I'm not sure if we need this PR as it's a duplicate of this and neutral elements was replaced instead of deleted.
> 
> https://transformuk.atlassian.net/browse/OTT-273 #1979

Not sure - the diff is still showing a difference though so might be worth merging anyway?

neilmiddleton:
Actually, strike that, the conflict removes the change

PR #1990, opened 2024-09-25
BAU: Removes last vestiges of beta search

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Remove last parts of beta search

### Why?

I am doing this because:

- This is now a defunct service and the new approach will be semantic and not token-based. Long-live semantics :black_flag:


neilmiddleton (dismissed):
🙌 
Screenshot 2024-09-25 at 12 19 52

PR #1992, opened 2024-09-25
BAU: Ruby 3.3.5

neilmiddleton:
### Jira link

BAU: Ruby 3.3.5

### What?

This PR updates Ruby to the latest version

### Why?

We should almost always be running the latest version.


HWallenberg (approved):
Grabbed and bundled locally successfully.  Seems good to me. Unit test issue seems like an environmental glitch.

willfish (approved):

PR #1993, opened 2024-09-26
BAU: Change time ReportWorker runs

neilmiddleton:
### Jira link

BAU

### What?

Changed the time the ReportWorker is triggered from 0800 to 1030 UTC.

### Why?

Whilst the UpdatesSynchronizers run at 0500, they re-try every 20 minutes until the file appear.  This continues until 1000, at which point it give us.
Meanwhile, over in reporting land, the ReportsWorker runs at 0800.

However, from the timestamps of files in S3 it seems that files are not appearing until around 0800 so the ReportsWorker is often missing the most recent files.

This change ensures that the ReportsWorker has either received the new files, or has given up trying.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -31,7 +31,7 @@
       cron: "0 18 * * *"
       description: "Populates search suggestions which are used in the frontend"
     ReportWorker:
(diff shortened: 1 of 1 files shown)


willfish (approved):
Makes sense

PR #2022, opened 2024-11-22
HMRC-385  Add AWS X-Ray

neilmiddleton:
### Jira link

HMRC-385

### What?

I have added AWS X-Ray which should give us plenty more tracing opportunities in production.  No effect on other environments outside AWS.


HWallenberg (approved):
thanks Neil, it will be interesting to see a demo of this when it's up and running. 👍

PR #2031, opened 2024-11-29
HMRC-385 Scout APM

neilmiddleton:
### Jira link

HMRC-385

### What?

Another day, another APM trial.  This adds Scout APM and will be reverted if we don't use it.


HWallenberg (approved):
✅

PR #2035, opened 2024-12-02
HMRC-423 Fix 20 Year Problem

HWallenberg:
Added regex check for correct date format, returns today if date is completely wrong

### Jira link

https://transformuk.atlassian.net/browse/HMRC-423

### What?

I have altered:

- [x] Application_Controller>Actual_Date to check for correct date format or return today's date.  

### Why?

I am doing this because:

- returning a date >20 years in the past caused a bug in the front end making a commodity look active when it isn't
- Dates thousands of years in the future are unacceptable however we don't want the app to error as trade dates can be in the future
- possible UX constraint for front end required to restrict date input

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -24,13 +24,25 @@ class ApplicationController < ActionController::API
   private
 
   def actual_date
-    date = Date.parse(params[:as_of].to_s)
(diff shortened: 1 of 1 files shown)


neilmiddleton (approved):

EraliaKats (approved):
Confirmed behaviour locally! All good - thanks Hadleigh!

PR #2049, opened 2024-12-13
 HMRC-418: Add models to the snapshort bulk import process 

rasikasri:
### Jira link

[HMRC-418](https://transformuk.atlassian.net/browse/HMRC-418)

### What?

I have added/removed/altered:

- [ ] Added process to import cds annual data file into a db

### Why?

I am doing this because:

- We need to have a backup db to resolve missing data issue


Dockerfile
```diff
@@ -57,6 +57,8 @@ RUN addgroup -S tariff && \
   chown -R tariff:tariff /app && \
   chown -R tariff:tariff /usr/local/bundle
 
+RUN chmod +x /app/config/init_backup.sh
```
willfish, line 60:
```suggestion
RUN chmod +x config/init_backup.sh
```

PR #2056, opened 2024-12-17
Revert "Allow ERB in ROO JSON files for date changes"

neilmiddleton:
Reverts a previous commit no longer needed.


HWallenberg (approved):
reverted due to caching issue, checked roo_schemes_uk.json with cptpp in and not affecting current code.

PR #2091, opened 2025-01-30
HMRC-576 

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HMRC-576

### What?

I have added 
- [x] noemail to env vars to toggle email functionality for non prod environments

### Why?

I am doing this because:

- We want to stop emails being sent to the support mailbox for non-prod functionality as it is confusing.


neilmiddleton (commented):
I'm not sure how I feel about this approach.  This enables/disables email rather than sending the email to a differently configured address.  This means that we have no way to test emails in these environments, only to make sure that we are not sending email.

I think what we want to do is have a catch-all email address we can all get to, and change the configured address in the appropriate environments?

terraform/locals.tf
```diff
@@ -137,6 +137,10 @@ locals {
       name  = "XE_API_URL"
       value = "https://xecdapi.xe.com"
     },
+    {
```
willfish, line 140: :man_cook: :kiss:

terraform/worker_uk.tf
```diff
@@ -70,6 +70,10 @@ module "worker_uk" {
       {
         name  = "ENVIRONMENT"
         value = var.environment
+      },
+      {
```
willfish, line 74: I don't think you need this chipolata since its configured in the common worker vars merged in from the locals

terraform/worker_xi.tf
```diff
@@ -74,6 +74,10 @@ module "worker_xi" {
       {
         name  = "ENVIRONMENT"
         value = var.environment
+      },
+      {
+        name  = "NOEMAIL"
```
willfish, line 79: Same here

willfish:
> I'm not sure how I feel about this approach. This enables/disables email rather than sending the email to a differently configured address. This means that we have no way to test emails in these environments, only to make sure that we are not sending email.
> 
> I think what we want to do is have a catch-all email address we can all get to, and change the configured address in the appropriate environments?

Don't we have a standard support email for the OTT team?

HWallenberg:
the support email is suppose to be for prod issues hence the PR to not send an email at all.  Consensus is that it's better to send an email to a test email address to ensure non prod environments still work.

HWallenberg:
the support email is suppose to be for prod issues hence the PR to not send an email at all.  Consensus is that it's better to send an email to a test email address to ensure non prod environments still work.

PR #2106, opened 2025-02-07
HMRC-667: Adds conditional implicit deletion on apply

willfish:
### Jira link

HMRC-667

### What?

I have added/removed/altered:

- [x] Added conditional implicit deletes apply 

### Why?

I am doing this because:

- This enables us to seamlessly apply updates without worrying about whether the implicit deletion should run


HWallenberg (approved):

willfish:
Tested this with a rollback/rollforward to 2022 March and all looks good

PR #2109, opened 2025-02-10
HMRC-657: Match the semantics of the frontend api requests

willfish:
### Jira link

HMRC-657

```mermaid
 flowchart TD
  style AFTER fill:#9f9,stroke:#333,stroke-width:2px

  %% After Flow
  subgraph AFTER [After]
    E[User Request] --> F[CDN/WAF]
    F --> G[Application Load Balancer]
    G --> H[Target Group XI]
    G --> I[Target Group UK]
  end

  %% Spacer between flows
  style BEFORE fill:#f9f,stroke:#333,stroke-width:2px
  %% Before Flow
  subgraph BEFORE [Before]
    A[User Request] --> B[CDN/WAF]
    B --> C[Application Load Balancer]
    C --> D[Frontend Request Forwarder]
    D --> X[Backend XI]
    D --> Y[Backend UK]
  end
```

### What?

I have added/removed/altered:

- [x] Added v1/v2 engines
- [x] Include v1/v2 engines under different paths to match frontend api request semantics
- [x] Consume engines as routes value in controller tests
- [x] Consume engine in rabl template rendered for v1
- [x] Fixes an issue with a misplaced faq controller route
- [x] Fixes an issue with some namespacing around admin commodities which were colliding with v2 commodity endpoints

### Why?

I am doing this because:

- This enables direct routing through the load balancer
- I picked engines since concerns wouldn't work for me and engines give us clear lines to load each of our main scopes (v1/v2) multiple times

### DIFF
--- a/app/controllers/api/admin/commodities/search_references_controller.rb
+++ b/app/controllers/api/admin/commodities/search_references_controller.rb
@@ -46,12 +46,16 @@ module Api
           end
         end
 
+        def admin_commodity_id
+          params[:admin_commodity_id].presence || params[:commodity_id]
+        end
+
         def commodity_id
-          params[:commodity_id].split('-', 2).first
+          admin_commodity_id.split('-', 2).first
         end
 
         def productline_suffix
-          params[:commodity_id].split('-', 2)[1] || '80'
+          admin_commodity_id.split('-', 2)[1] || '80'
         end
       end
     end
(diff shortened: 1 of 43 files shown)


app/controllers/api/admin/commodities/search_references_controller.rb
```diff
@@ -46,12 +46,16 @@ def subheading
           end
         end
 
+        def admin_commodity_id
+          params[:admin_commodity_id].presence || params[:commodity_id]
+        end
+
         def commodity_id
-          params[:commodity_id].split('-', 2).first
+          admin_commodity_id.split('-', 2).first
```
willfish, line 54: I had to rename the resource in the separated engine to avoid collisions on path helpers, here. It'll make no difference to the caller of this api

HWallenberg (approved):

PR #2112, opened 2025-02-11
HMRC-575 & HMRC-629 Fix VAT in Nested Set Measures Overview (Headings & Subheadings view) and Differences Report

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HMRC-575
https://transformuk.atlassian.net/browse/HMRC-629

### What?

I have:

- [x] Added new Geographical_Area (1400) for Measure_Type VAT (305) for Nested Set Measure Overview 
- [x] Removed redundant filter following change https://github.com/trade-tariff/trade-tariff-backend/pull/2017

### Why?

I am doing this because:

- CDS provide a new geographical area id (1400) for VAT and Excise Measure Types (305) which weren't showing on the Headings/Subheadings page or in the differences report.
- The differences report Missing VAT Measures Tab doesn't filter by geographical area anymore.

RESULTS

DIFFERENCES REPORT:
image

SUBHEADINGS VIEW:
image


app/models/measure.rb
```diff
@@ -424,7 +424,7 @@ def overview
         if TradeTariffBackend.uk?
           overview_types << {
             measures__measure_type_id: MeasureType::VAT_TYPES,
-            measures__geographical_area_id: GeographicalArea::ERGA_OMNES_ID,
+            measures__geographical_area_id: GeographicalArea::AREAS_SUBJECT_TO_VAT_OR_EXCISE,
```
willfish, line 427: Should this be an enumerable like an Array or do we want to change the name to be singular?

willfish (approved):
LGTM

app/models/measure.rb
HWallenberg: I just realised maybe it should have _ID at the end, if it turns into lots of geo_areas then it can become an array?

willfish:
Smoke tests are failing due to my branch in the frontend

PR #2117, opened 2025-02-13
HMRC-683 Enable admin in backend

neilmiddleton:
### Jira link

HMRC-683

### What?

Required by https://github.com/trade-tariff/trade-tariff-admin/pull/825

### DIFF
--- a/terraform/backend_uk.tf
+++ b/terraform/backend_uk.tf
@@ -44,6 +44,10 @@ module "backend_uk" {
         value = "true"
       },
       {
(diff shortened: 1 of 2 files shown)


HWallenberg (approved):

terraform/backend_uk.tf
```diff
@@ -43,6 +43,10 @@ module "backend_uk" {
         name  = "CDS"
         value = "true"
       },
```
willfish, line 45: I think we can just delete the env var and move to having all places serve these paths

PR #2124, opened 2025-02-17
BAU: Adjustments after removing legacy routes

willfish:
### Jira link

BAU

### What?

So the idea here was to completely remove all of the legacy routes, see what failures we experience and course correct on these failures to guarantee explicit compatibility with the new mounted paths (e.g. `/api/v2/`)

It's worth knowing that we route propagated exceptions to the relevant controller action _without_ making another request to the backend. This will route requests using the show_exceptions.rb middleware https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L55 and call our configured exceptions_app (see application.rb)

```ruby
    config.exceptions_app = routes
```

I was specifically interested in this because if an error is propagated in the backend api controllers it'll route to `/400`. If this was a brand new request/redirect then the user would always get a response from our frontend

I have added/removed/altered:

- [x] Fixed broken relative v1 links to the backend
- [x] Migrated errors/their handling to be under v1/v2 paths
- [x] Migrated faq mailer to be xi only
- [x] Removed all references to unused version headers in the test suite

### Why?

I am doing this because:

- These locations don't live under /trade-tariff anymore and can instead
be found under /uk/api/v1/*
- This is essentially a two part investigation into breakages with the routing


.env.development
```diff
@@ -1,3 +1,4 @@
+AWS_DEFAULT_REGION=eu-west-2
```
willfish, line 1: Our gems depend on this value being set so this creates a convention that reduces the configuration a dev needs to do

EraliaKats (approved):

PR #2129, opened 2025-02-18
BAU: Handle path rewrites on chapters/headings hitting the commodities

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Added middleware to handle heading/chapter requests on the commodities paths
- [x] Registered the new middleware
- [x] Added some unit test coverage

### Why?

I am doing this because:

- This is needed to fix some of the 404 responses we were getting in the backend sentry alerts and was discovered when I was investigating the failing paths


EraliaKats (approved):

willfish:
These errors are unrelated. I think what's happening is that the new middleware is part of the stack of methods that got invoked to hit these endpoints since its middleware

PR #2131, opened 2025-02-19
BAU: Put more specific routes first

willfish:
### Jira link

BAU

### What?

Occasionally but not consistently the legacy routes are being hit and failing because the Accept header isn't being set. We only saw this in [production](https://engine-le.sentry.io/issues/6294252303/events/?end=2025-02-17T23%3A59%3A59&environment=production&project=5557005&referrer=issue-stream&start=2025-02-17T00%3A00%3A00&stream_index=0&utc=true)

This change moves us to match the more specific routes first (e.g. /api/v1 /api/v2) to avoid this problem. It also fixes the missing Accept header error by calling to_s on the result of fetching the header before we call `include?` against nil.

When you put the most specific mounts first there's an unwanted side effect of clobbering the previously defined helpers. This mostly affected the tests (controller specs use helpers implicitly and request specs were using helpers explicitly).

Essentially this means we're now guaranteeing correct matches for v1/v2 paths but rendered paths now need some adjustments

I have added/removed/altered:

- [x] Added more robust route ordering to force v1/v2 matches over legacy routes
- [x] Added handling of nil Accept headers in legacy ApiConstraints class
- [x] Adjusts helper rendered paths in the tests and under v1 rabl views

### Why?

I am doing this because:

- This forces more specific routes to be recognised first
- This reduces the flimsiness of the ApiConstraints implementation
- This guarantees that rendered paths follow the new convention (there's a bit of a TODO to revert back to helpers once the legacy paths have been dropped)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -17,4 +17,12 @@ module ApplicationHelper
     regulation_id = regulation.regulation_id
     "#{regulation_id.first}#{regulation_id[3..6]}/#{regulation_id[1..2]}"
   end
+
+  # TODO: Remove this and use the original path helpers once we're removed legacy routes which clobber the correct helpers
+  def v1_api_path(resource, id = nil)
+    service = TradeTariffBackend.service
+    id = "/#{id}" if id
+
+    "/#{service}/api/v1/#{resource}#{id}"
+  end
 end
--- a/app/controllers/api/v2/goods_nomenclatures_controller.rb
+++ b/app/controllers/api/v2/goods_nomenclatures_controller.rb
@@ -74,21 +74,22 @@ module Api
       end
 
       def self.api_path_builder(object, check_for_subheadings: false)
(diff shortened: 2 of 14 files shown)


app/helpers/application_helper.rb
```diff
@@ -17,4 +17,12 @@ def self.regulation_code(regulation)
     regulation_id = regulation.regulation_id
     "#{regulation_id.first}#{regulation_id[3..6]}/#{regulation_id[1..2]}"
   end
+
+  # TODO: Remove this and use the original path helpers once we're removed legacy routes which clobber the correct helpers
+  def v1_api_path(resource, id = nil)
```
willfish, line 22: Because moving the legacy routes to the bottom of the routes.rb file clobbers the helpers we actually want to use I've opted to correct these by using this application helper method.

These will automatically remedy when we've removed the legacy routes

neilmiddleton (approved):

PR #2132, opened 2025-02-19
HMRC-670 Remove 'appendix 5a' jobs and infrastructure for CHIEF guidance

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HMRC-670

### What?

I have removed CHIEF Guidance from Appendix5a Populator Service

### Why?

I am doing this because:

- HMRC do not need CHIEF Guidance anymore and has been deprecated from the front end and in this [PR](https://github.com/trade-tariff/process-appendix-5a/pull/79) for the process-appendix-5a python script
- NOTE following successfuly merge to main, AWS `ci-appendix5a-persistence-readwrite-policy`  must be updated to allow `appendix-5a` user to `PUT` into S3 `config/cds_guidance.json` in Staging and Prod.

### DIFF
--- a/app/models/appendix_5a.rb
+++ b/app/models/appendix_5a.rb
@@ -1,5 +1,5 @@
 class Appendix5a < Sequel::Model(:appendix_5as)
-  CDS_GUIDANCE_OBJECT_KEY = 'config/chief_cds_guidance.json'.freeze
+  CDS_GUIDANCE_OBJECT_KEY = 'config/cds_guidance.json'.freeze
 
   set_primary_key %i[certificate_type_code certificate_code]
 
@@ -20,7 +20,7 @@ class Appendix5a < Sequel::Model(:appendix_5as)
   end
 
   def guidance_chief
-    chief_guidance
+    ''
   end
 
   def guidance_cds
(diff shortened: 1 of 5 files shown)


app/models/appendix_5a.rb
```diff
@@ -20,7 +20,7 @@ def document_code
   end
 
   def guidance_chief
-    chief_guidance
+    ''
```
EraliaKats, line 23: I'm conscious of this change as we're using the Dirty plugin. Will this change all guidance to ''?

HWallenberg: Interesting I don't think so.  
It should only return guidance_chief as empty which is no longer shown in the front end, this is tested in the specs as updated.
However there is another ticket (I think) to completely remove guidance_chief from the appendix5a model.
Models that delegate to appendix5a such as app/models/measure_condition.rb and app/models/certificate.rb will need updating along with their serializers.
Out of scope for this PR AFAIK as we only want to change the process for now.
Otherwise we end up with a large change and this can be seen as a interim placeholder for now to ensure the daily job still works until we completely remove chief_guidance from the model.

EraliaKats: Apologies, I meant all chief guidance. I agree with you. Also, I realised we're not going to need backwards compatibility so makes it a little easier to kill it anyway.

rasikasri (approved):

PR #2135, opened 2025-02-21
removed type from drop column to fix migration

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HMRC-670

### What?

- [x] I have removed type from migration file drop column 

### Why?

I am doing this because:

- It was causing an error in the migration


db/migrate/20250219092427_delete_chief_guidance_column_from_appendix5a.rb
```diff
@@ -3,7 +3,7 @@
 Sequel.migration do
   up do
     alter_table(:appendix_5as) do
-      drop_column :chief_guidance, String
+      drop_column :chief_guidance
```
EraliaKats, line 6: If specifying a type throws an error in up, wouldn't it do the same at down?

EraliaKats (approved):

PR #2138, opened 2025-02-24
BAU: Removes unnecessary references to CDS

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Removed references to CDS environment variable
- [x] Added minimum needed env vars to run sidekiq worker jobs for uk locally

### Why?

I am doing this because:

- This was actually replaced by and duplicates the SERVICE variable and is potentially needed consolidation for merging these two services into a single worker


app/workers/rollback_worker.rb
```diff
@@ -4,7 +4,9 @@ class RollbackWorker
   sidekiq_options queue: :rollbacks, retry: false
 
   def perform(date, redownload = false)
-    if TradeTariffBackend.use_cds?
+    if TradeTariffBackend.uk?
+      require 'pry'
+      binding.pry
```
EraliaKats, line 9: oops

EraliaKats (approved):

PR #2141, opened 2025-02-26
BAU: Adds rubocop-performance

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- Adds rubocop performance
- Lint with new rules

### Why?

I am doing this because:

- I noticed another government project using this and it seems like an easy win

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,7 +3,10 @@ inherit_gem:
     - config/default.yml
     - config/rails.yml
 
-require: rubocop-rspec
+require:
+  - rubocop-rspec
+  - rubocop-performance
+
 
 AllCops:
   Exclude:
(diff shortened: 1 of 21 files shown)


HWallenberg (approved):

HWallenberg:
> I'm unsure about this one, mostly down to it's pendantry. For instance, it's keen on this change:
> 
> ```
>         values.merge!(filename:)
>         values[:filename] = filename
> ```
> 
> but this going to yield barely anything. Yes it adds up at scale but I think we might see more developer frustration at trying to keep the cop gods happy vs any benefit we get in CPU use.

oops, there me thinking it is progressive and approved on that basis. 🤦

willfish:
This is exactly how I felt. I can take it or leave it @neilmiddleton

PR #2160, opened 2025-03-07
HMRC-265 Check all reports work

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HMRC-265

### What?

I have added code to replace whitespace unicode with whitespace in GN description to fix csv to excel conversion

### DIFF
--- a/app/lib/reporting/basic.rb
+++ b/app/lib/reporting/basic.rb
@@ -92,7 +92,7 @@ module Reporting
 
         row = []
         row << goods_nomenclature.goods_nomenclature_item_id
-        row << goods_nomenclature&.goods_nomenclature_description&.description_plain
+        row << goods_nomenclature&.goods_nomenclature_description&.description_plain&.tr("\u00A0", ' ')
         row << third_country_duty
         row << supplementary_measure&.supplementary_unit_duty_expression
 


willfish (approved):

app/lib/reporting/basic.rb
```diff
@@ -92,7 +92,7 @@ def build_row_for(goods_nomenclature)
 
         row = []
         row << goods_nomenclature.goods_nomenclature_item_id
-        row << goods_nomenclature&.goods_nomenclature_description&.description_plain
+        row << goods_nomenclature&.goods_nomenclature_description&.description_plain&.tr("\u00A0", ' ')
```
willfish, line 95:
```ruby
    it 'replaces \u00A0 with empty space' do
      expect(
        described_class.format(description: "Hadleigh\u00A0is awesome"),
      ).to eq 'Hadleigh is awesome'
    end

    str.gsub!("\u00A0", ' ')
```

https://github.com/trade-tariff/trade-tariff-backend/blob/main/app/formatters/description_trim_formatter.rb

HWallenberg: Hi @willfish I've tried this solution locally and it doesn't resolve the issue when opening the csv up in excel so I'm going to merge the original solution if that's okay?

willfish:
No worries buddy. Yeah it works so all good

PR #2174, opened 2025-03-18
HMRC-737: Change the daily load process to a 'batch' process (learnin…

rasikasri:
…gs from annual file importer)

### Jira link

### What?

I have added/removed/altered:

- [ ] Added foo in bar
- [ ] Removed baz

### Why?

I am doing this because:

-
-
-

### Have you? (optional)


app/lib/trade_tariff_backend.rb
```diff
@@ -266,6 +266,10 @@ def implicit_deletion_cutoff
       Date.parse(ENV.fetch('IMPLICIT_DELETION_CUTOFF', '2024-03-25'))
     end
```
willfish, line 268: It’s safer to have a default here using fetch

app/lib/trade_tariff_backend.rb
```diff
@@ -266,6 +266,10 @@ def implicit_deletion_cutoff
...
 
+    def cds_importer_batch_size
+      ENV['CDS_IMPORT_BATCH_SIZE'].to_i
```
willfish, line 270: It’s safer to just have a default here I think

willfish (dismissed):
Are you okay to run this in a live environment if you haven't already?

willfish (approved):

PR #2176, opened 2025-03-20
HMRC-840: Add new Trade Defence CAs 'regulation role' types and Measure Types

rasikasri:
### Jira link

[HMRC-840](https://transformuk.atlassian.net/browse/HMRC-840)

### What?

I have added/removed/altered:

- [ ] Added filter to Import Trade Remedies CategoryAssessments task to filter out inactive regulations
- [ ] Removed filter from Trade Remedies CategoryAssessments task that filter out regulation roles other than role 1

### Why?

I am doing this because:

- We should not control for the ‘regulation role’ at all, and any Measure of these given Measure Types should be a CA no matter what the ‘regulation role’ is

- Need to run this task in all the environemtns manually

### DIFF
--- a/lib/tasks/green_lanes.rake
+++ b/lib/tasks/green_lanes.rake
@@ -126,19 +126,9 @@ namespace :green_lanes do
     types = %w[551 552 553 554 555 561 562 564 565 566 570 690 695 696]
 
     GreenLanes::CategoryAssessment.db.transaction do
-      current_date = Time.zone.today
-      actual_condition = Sequel.lit('validity_start_date <= ? AND ' \
-                                      '(validity_end_date IS NULL OR validity_end_date > ?)')
-
       # select active measure_id, regulation_id and regulation_role of given measure type to create category_assessment
-      Measure.where(actual_condition, current_date, current_date)
-             .where(measure_type_id: types)
-             .where(
-               Sequel.|(
-                 { measure_generating_regulation_id: ModificationRegulation.where(actual_condition, current_date, current_date).select(:modification_regulation_id) },
-                 { measure_generating_regulation_id: BaseRegulation.where(actual_condition, current_date, current_date).select(:base_regulation_id) },
-               ),
-             )
+      Measure.where(measure_type_id: types)
+             .actual.with_regulation_dates_query
              .select_group(:measure_type_id, :measure_generating_regulation_id, :measure_generating_regulation_role)
              .all.each do |tr|
         key = [tr.measure_type_id, tr.measure_generating_regulation_id, tr.measure_generating_regulation_role]
--- a/lib/tasks/green_lanes.rake
+++ b/lib/tasks/green_lanes.rake
@@ -126,9 +126,8 @@ namespace :green_lanes do
     types = %w[551 552 553 554 555 561 562 564 565 566 570 690 695 696]
 
     GreenLanes::CategoryAssessment.db.transaction do
-      # select active measure_id, regulation_id and regulation_role of given measure type to create category_assessment
       Measure.where(measure_type_id: types)
-             .actual.with_regulation_dates_query
+             .with_generating_regulation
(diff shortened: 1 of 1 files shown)


lib/tasks/green_lanes.rake
```diff
@@ -121,20 +121,26 @@ namespace :green_lanes do
...
                                       '(validity_end_date IS NULL OR validity_end_date > ?)')
 
       Measure.where(actual_condition, current_date, current_date)
-             .where(measure_type_id: types, measure_generating_regulation_role: 1)
+             .where(measure_type_id: types)
+             .where(
+               Sequel.|(
+                 { measure_generating_regulation_id: ModificationRegulation.where(actual_condition, current_date, current_date).select(:modification_regulation_id) },
+                 { measure_generating_regulation_id: BaseRegulation.where(actual_condition, current_date, current_date).select(:base_regulation_id) },
+               ),
```
EraliaKats, line 139: Thinking it'd be good to add a comment about what you're trying to achieve here? Had to read the PR to completely understand the need for this

rasikasri: Done

lib/tasks/green_lanes.rake
```diff
@@ -121,20 +121,27 @@ namespace :green_lanes do
...
     end
-    types = %w[551 552 553 554 555 561 562 564 565 566 570 695]
+    types = %w[551 552 553 554 555 561 562 564 565 566 570 690 695 696]
 
     GreenLanes::CategoryAssessment.db.transaction do
       current_date = Time.zone.today
       actual_condition = Sequel.lit('validity_start_date <= ? AND ' \
                                       '(validity_end_date IS NULL OR validity_end_date > ?)')
 
+      # select active measure_id, regulation_id and regulation_role of given measure type to create category_assessment
```
EraliaKats, line 133: Sorry should've specified what I meant: a comment to explain what we're doing business-wise. Like what you said on the PR "[...] filter out inactive regulations"

lib/tasks/green_lanes.rake
```diff
@@ -121,20 +121,27 @@ namespace :green_lanes do
...
       Measure.where(actual_condition, current_date, current_date)
-             .where(measure_type_id: types, measure_generating_regulation_role: 1)
+             .where(measure_type_id: types)
+             .where(
+               Sequel.|(
+                 { measure_generating_regulation_id: ModificationRegulation.where(actual_condition, current_date, current_date).select(:modification_regulation_id) },
```
willfish, line 138: You can forego this filter since there are only two kinds of generating regulation

lib/tasks/green_lanes.rake
```diff
@@ -121,20 +121,17 @@ namespace :green_lanes do
...
 
     GreenLanes::CategoryAssessment.db.transaction do
-      current_date = Time.zone.today
-      actual_condition = Sequel.lit('validity_start_date <= ? AND ' \
-                                      '(validity_end_date IS NULL OR validity_end_date > ?)')
-
-      Measure.where(actual_condition, current_date, current_date)
-             .where(measure_type_id: types, measure_generating_regulation_role: 1)
+      # select active measure_id, regulation_id and regulation_role of given measure type to create category_assessment
+      Measure.where(measure_type_id: types)
```
willfish, line 130: I think you'll want to wrap this in `TimeMachine.now do ... end`

lib/tasks/green_lanes.rake
EraliaKats: The new changes don't require a comment now. So please remove it 😊

lib/tasks/green_lanes.rake
rasikasri: now it set with the thread current provided by active support, isn't it?
here - 
https://github.com/trade-tariff/trade-tariff-backend/blob/5f91a668f475cd2790b04ae2d21e75f46911963a/app/lib/trade_tariff_request.rb#L3

willfish (approved):

PR #2200, opened 2025-04-03
BAU: Fixes some flaky tests

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- Fixes flaky order-independent tests
- Bumps my local database version to 17 to match the CI env

### DIFF
--- a/spec/models/goods_nomenclature_spec.rb
+++ b/spec/models/goods_nomenclature_spec.rb
@@ -460,8 +460,8 @@ RSpec.describe GoodsNomenclature do
     context 'when footnote_types_and_ids is empty' do
       let(:footnote_types_and_ids) { [] }
 
-      it { expect(dataset.pluck(:footnote_id)).to eq %w[123 456 789] }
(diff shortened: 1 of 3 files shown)


spec/models/goods_nomenclature_spec.rb
```diff
@@ -460,8 +460,8 @@
     context 'when footnote_types_and_ids is empty' do
       let(:footnote_types_and_ids) { [] }
 
-      it { expect(dataset.pluck(:footnote_id)).to eq %w[123 456 789] }
-      it { expect(dataset.pluck(:footnote_type_id)).to eq %w[Y N Z] }
+      it { expect(dataset.pluck(:footnote_id)).to include('123', '456', '789') }
```
vertism, line 463: Could you use `match_array`?

vertism: I think match_array ignores order but will fail if there are extra items.

vertism (approved):

PR #2206, opened 2025-04-09
Move ROO to lib

neilmiddleton:
For some reason, all the ROO code sits in `/db` which makes no sense as it's nothing to do with the DB.

This PR moves it to `/lib`


HWallenberg (approved):
thanks @neilmiddleton! 🚀

PR #2210, opened 2025-04-11
HMRC-979: Invalidate paths that are short medium and long duration

willfish:
### Jira link

[HMRC-979](https://transformuk.atlassian.net/browse/HMRC-979)

### What?

I have added/removed/altered:

- [x] Added more specific invalidations of paths

### Why?

I am doing this because:

- This preserves caches for very very long cached things


vertism:
What are the paths that this doesn't invalidate?

willfish:
Things that don't change often (assets/packs/images)

PR #2216, opened 2025-04-16
HMRC-977 GovUK Notify Wrapper

neilmiddleton:
### Jira link

[HMRC-977](https://transformuk.atlassian.net/browse/HMRC-977)

### What?

Adds a wrapper around Gov.Uk Notify allowing us to send emails via that service as we need and track the emails that have been sent in an audit table.  e.g:

```ruby
GovNotifier.new.send_email('test@example.com', <template_id>,  personalisation: {
  foo: 'bar',
})
```

Emails are logged in `public.govuk_notifier_audits` in addition to the 7 days stored on the service itself.

Note, there is still a TODO around email unsubscription which needs to be figured out, but this can be used as is.

This code does require `GOVUK_NOTIFY_API_KEY` to be set in `ENV`
Requires https://github.com/trade-tariff/trade-tariff-platform-aws-terraform/pull/354


db/migrate/20250411114556_create_govuk_notifier_audit.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    create_table Sequel[:govuk_notifier_audits].qualify(:public), id: :uuid do
+      primary_key :idid
```
willfish, line 6:
```suggestion
      primary_key :id
```

spec/factories/notifications_client_post_email_response.rb
```diff
@@ -0,0 +1,27 @@
+require 'notifications/client'
```
willfish, line 1: I don’t think you need this require as zeitwerk is doing this for you

neilmiddleton:
Blocked by https://github.com/trade-tariff/trade-tariff-platform-aws-terraform/pull/354

PR #2220, opened 2025-04-22
HMRC-971: Clear Cache-Control header on non-frontend requests

willfish:
### Jira link

[HMRC-971](https://transformuk.atlassian.net/browse/HMRC-971)

### What?

I have added/removed/altered:

- [x] Added middleware to conditionally clear out [Cache-Control][cache-control] headers

### Why?

I am doing this because:

- This is required to prevent instructions to cache responses in the CDN for a very short lifetime.
- Without this the API requests have the same duration as the etagged requests from the frontend which are purposely short lived because verifying etags is a lightweight request

[cache-control]: https://datatracker.ietf.org/doc/html/rfc5861#:~:text=RFC%205861%20%2D%20HTTP%20Cache%2DControl%20Extensions%20for%20Stale%20Content


HWallenberg (approved):
lgtm.

PR #2227, opened 2025-04-22
Update deploy-to-development.yml

neilmiddleton:
As discussed https://trade-tariff.slack.com/archives/C05T3PAH1EJ/p1745333138908849.

Don't deploy to dev on every PR/Push


willfish (commented):
I think we need to run the other jobs but filter out builds and deploys and conditionally run notifications and e2e tests

PR #2229, opened 2025-04-24
HMRC-998 Refactor and fix bug where not all results are counted

vertism:
### Jira link

[HMRC-998](https://transformuk.atlassian.net/browse/HMRC-998)

### What?

When total results were being counted for logging purposes, some potential results were not being assigned and not being added to the total.
The code has been refactored to loop through all possible combinations.

### Why?

I am doing this because:

it fixes a bug


willfish (approved):
Love it

PR #2232, opened 2025-04-25
HMRC-274: Hide old CHIEF P and Q series Measures in /XI APIs

rasikasri:
### Jira link

[HMRC-274](https://transformuk.atlassian.net/browse/HMRC-274)

### What?

I have added/removed/altered:

- [ ] Added a condition to filter out P and Q series measures from XI api

### Why?

I am doing this because:

- Measures in Series “P” and “Q” in the XI database are no longer valid and should not be used by traders, and are switched for the UK Measures on thefront end


app/models/goods_nomenclatures/nested_set.rb
```diff
@@ -127,6 +127,7 @@ def initialize
           .dedupe_similar
           .with_regulation_dates_query
           .without_excluded_types
+          .without_excluded_series
```
willfish, line 130: Rather than permanently extending the series, I'm wondering whether we should do a migration to delete these measures?

willfish: Talked about this offline

The plan is to:

1. Turn these off
2. Observe for an arbitrary period to see if we get complaints/people depending on these
3. Remove the data in a subsequent story

willfish:
@rasikasri Have you ran the entire test suite locally?

willfish (approved):

rasikasri:
Screenshot 2025-04-28 at 14 52 46

willfish (approved):

PR #2255, opened 2025-05-14
HMRC-1025 Select Chapters

HWallenberg:
### Jira link

[HMRC-1025](https://transformuk.atlassian.net/browse/HMRC-1025)

### What?

I have added migration and model to create `change_notifications` for storing userids and selected chapters

### Why?

I am doing this because:

- MyOtt requires this data to be stored


app/models/my_ott/user_preference.rb
```diff
@@ -0,0 +1,8 @@
+module MyOtt
```
vertism, line 1: Should this be in the PublicUsers module along with User?

HWallenberg: Agree just wasn't sure what is for the best I'll update it.

app/models/my_ott/user_preference.rb
```diff
@@ -0,0 +1,8 @@
...
+  class UserPreference < Sequel::Model(Sequel[:user_preferences].qualify(:public))
+    plugin :auto_validations
+    plugin :timestamps, update_on_create: true
+
+    many_to_one :user_subscriptions, class: 'PublicUsers::Subscription', key: :user_id
```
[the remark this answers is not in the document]
HWallenberg: Probably not and overkill if you think we don't need constraints. 👍

vertism (approved):

willfish (approved):

PR #2269, opened 2025-05-20
Decrypt identity JWT

vertism:
### What?

Decrypts id token
Removes verification of token when running locally

### Why?

To make developing easier and allow removal of decryption from frontend.


app/lib/cognito_token_verifier.rb
```diff
@@ -23,4 +17,34 @@ def self.jwks_keys
...
+  def initialize(token)
+    @token = token
+  end
+
+  def verify
+    decrypt.decode.token[0]
+  end
+
+  def decrypt
+    unless Rails.env.development?
```
willfish, line 32: Could the service just handle this question and let the assignment happen regardless?

vertism: I did consider that option but decided to do it here so it's consistent with the decode method.

HWallenberg: Both I'm going to mark this as resolved so we can progress with this.
@willfish I hope this is acceptable, if we need to change where the env flag goes then we can do so in another ticket, thanks. H

HWallenberg (approved):

PR #2272, opened 2025-05-23
HMRC-1074 Subscription matcher worker

vertism:
### Jira link

[HMRC-1074](https://transformuk.atlassian.net/browse/HMRC-1074)

### What?

Triggers worker after news item is saved that identifies users whose subscription matches the news item.

Matching is based on:
- The user having an active subscription
- The news item being emailable
- The news item relating to chapters that the user set a preference for

### Why?

I am doing this because:

- Part of MyOTT subscriptions.

### DIFF
--- a/app/models/public_users/user.rb
+++ b/app/models/public_users/user.rb
@@ -10,6 +10,30 @@ module PublicUsers
 
     attr_accessor :email
 
+    dataset_module do
+      def with_active_stop_press_subscription
+        join(:public__user_subscriptions, user_id: :id)
+          .where(
+            Sequel[:user_subscriptions][:active] => true,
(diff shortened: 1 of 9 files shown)


app/models/public_users/user.rb
```diff
@@ -10,6 +10,30 @@ class User < Sequel::Model(Sequel[:users].qualify(:public))
 
     attr_accessor :email
 
+    dataset_module do
+      def with_active_stop_press_subscription
+        join(:public__user_subscriptions, user_id: :id)
```
willfish, line 15: Could this be defined by an association?

vertism: I don't think I can use the association in dataset_module, and this method needs to return the user, so I can't define the whole thing in an association. 
Unless I'm missing something this seems like the best way.

willfish (approved):

PR #2277, opened 2025-05-29
HMRC-1143:Adjust BE logic to automatically create pairs for new MType…

rasikasri:
### Jira link

[HMRC-1143](https://transformuk.atlassian.net/browse/HMRC-1143)

### What?

I have added/removed/altered:

- [ ] Updated green lane update worker
- [ ] If new measure found with mtype that is identified as SPIMM ralated, then create a new category assessment

### Why?

I am doing this because:

- To support automatic generation of Category Assessments for specific Measure Types, a new database table is needed to map Measure Type IDs to SPIMM Themes
- If the MType is in the list, it should auto-create a Category Assessment for the new MType-RegID-Role combination with the correct theme assigned


spec/models/green_lanes/identified_measure_type_category_assessment_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  describe 'validations' do
+    subject(:errors) { instance.tap(&:valid?).errors }
+
+    let(:instance) { described_class.new }
+
+    it { is_expected.to include measure_type_id: ['is not present'] }
+    it { is_expected.to include theme_id: ['is not present'] }
+  end
+
+  describe 'validations duplicates' do
```
willfish, line 20: It feels like we're using a describe here as a context (context is a thing that varies). Essentially you're describing validations and varying the context of what validations get triggered https://www.betterspecs.org/ but essentially describing the same thing. In fact from a SUT perspective the thing you're really describing is `#valid?`

rasikasri: Changed it to a context

willfish (approved):

PR #2278, opened 2025-05-29
HMRC-1074 Worker to send subscription emails

vertism:
### Jira link

### What?

I have added/removed/altered:

- [ ] Added foo in bar
- [ ] Removed baz

### Why?

I am doing this because:

-
-
-

### Have you? (optional)


app/lib/trade_tariff_backend.rb
```diff
@@ -273,5 +273,17 @@ def cupid_team_to_emails
...
+
+    def identity_api_host
+      ENV['IDENTITY_API_HOST']
+    end
+
+    def identity_api_key
+      ENV['IDENTITY_API_KEY']
+    end
+
+    def myott?
```
willfish, line 285:
```suggestion
    def myott_enabled?
```

PR #2296, opened 2025-06-05
Check group membership

vertism:
### Jira link

[HMRC-1174](https://transformuk.atlassian.net/browse/HMRC-1174)

### What?

I have added/removed/altered:

- Added check for membership of myott user group
- Added logging for failed JWT verification

### Why?

I am doing this because:

- Only myott users should be given access

### DIFF
--- a/app/lib/cognito_token_verifier.rb
+++ b/app/lib/cognito_token_verifier.rb
@@ -7,7 +7,11 @@ class CognitoTokenVerifier
     return nil if jwks_keys.nil? && !Rails.env.development?
 
     new(token).verify
-  rescue StandardError
+  rescue JWT::ExpiredSignature
+    Rails.logger.info('Cognito JWT::ExpiredSignature')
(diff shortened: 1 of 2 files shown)


app/lib/cognito_token_verifier.rb
```diff
@@ -7,7 +7,11 @@ def self.verify_id_token(token)
     return nil if jwks_keys.nil? && !Rails.env.development?
 
     new(token).verify
-  rescue StandardError
+  rescue JWT::ExpiredSignature
+    Rails.logger.info('Cognito JWT::ExpiredSignature')
+    nil
+  rescue JWT::DecodeError
+    Rails.logger.info('Cognito JWT::DecodeError')
     nil
```
HWallenberg, line 15: @vertism do you think there is a need to catch a generic error here still? Thanks.

HWallenberg (approved):

PR #2308, opened 2025-06-10
Added destroy (unsubscribe) functionality

HWallenberg:
### Jira link

[HMRC-1221](https://transformuk.atlassian.net/browse/HMRC-1221)

### What?

I have added 

- [x] Soft Delete functionality

### Why?

I am doing this because:

- Its required for unsubscribe journey


db/structure.sql
```diff
@@ -12915,6 +12916,11 @@ INSERT INTO "schema_migrations" ("filename") VALUES ('20250425145123_add_chapter
 INSERT INTO "schema_migrations" ("filename") VALUES ('20250514093320_create_user_preferences.rb');
 INSERT INTO "schema_migrations" ("filename") VALUES ('20250516154516_add_notify_subscribers_to_news_item.rb');
 INSERT INTO "schema_migrations" ("filename") VALUES ('20250522141643_add_subscribable_to_news_collection.rb');
+<<<<<<< Updated upstream
```
vertism, line 12919: Needs removing

spec/models/public_users/user_spec.rb
```diff
@@ -97,6 +97,40 @@
     end
   end
 
+  describe 'soft_delete!' do
```
willfish, line 100:
```suggestion
  describe '#soft_delete!' do
```

app/models/public_users/user.rb
```diff
@@ -58,6 +64,14 @@ def stop_press_subscription=(active)
       end
     end
 
+    def soft_delete!
+      return if deleted || stop_press_subscription
```
vertism, line 68: Something like
```suggestion
      return if deleted
      return if subscriptions_dataset.where(active: true).any?
```

HWallenberg: see committed changes thanks

db/structure.sql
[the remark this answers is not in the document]
HWallenberg: yes not sure what happened here!

vertism (approved):

PR #2310, opened 2025-06-10
HMRC-1221 Unsubscribe from stop press emails

vertism:
### Jira link

[HMRC-1221](https://transformuk.atlassian.net/browse/HMRC-1221)

### What?

I have added/removed/altered:

- Added API method for unsubscribing
- Will also try to delete user

### Why?

I am doing this because:

- API for frontend to use for when user unsubscribes


HWallenberg (approved):
lgtm! 🚀

PR #2319, opened 2025-06-12
Delete external user worker

vertism:
### Jira link

[HMRC-1224](https://transformuk.atlassian.net/browse/HMRC-1224)

### What?

- Added worker to request users are deleted from Identity service

### Why?

I am doing this because:

- To remove users personal details when they are no longer required

### DIFF
--- a/app/lib/identity_api_client.rb
+++ b/app/lib/identity_api_client.rb
@@ -2,14 +2,24 @@ module IdentityApiClient
   def self.get_email(username)
     return nil unless username
 
-    url = URI.join(TradeTariffBackend.identity_api_host, '/api/users/', username).to_s
-    response = Faraday.get(url) do |req|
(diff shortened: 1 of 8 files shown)


joao-krauchenco-mottu (approved):

app/lib/identity_api_client.rb
```diff
@@ -2,14 +2,24 @@ module IdentityApiClient
...
+    return nil unless username
+
+    response = user_request(:delete, username)
+    response.success?
+  end
+
+  def self.user_request(method, username)
     url = URI.join(TradeTariffBackend.identity_api_host, '/api/users/', username).to_s
-    response = Faraday.get(url) do |req|
+    Faraday.send(method, url) do |req|
```
willfish, line 20:
```suggestion
    Faraday.public_send(method, url) do |req|
```

willfish (approved):

PR #2331, opened 2025-06-17
HMRC-1265 Optimise search suggestions

neilmiddleton:
### Jira link

[HMRC-1265](https://transformuk.atlassian.net/browse/HMRC-1265)

### What?

Add a more efficient query for search suggestions, and add indexes that will help

### DIFF
--- a/app/models/search_suggestion.rb
+++ b/app/models/search_suggestion.rb
@@ -38,22 +38,20 @@ class SearchSuggestion < Sequel::Model
           .limit(10)
           .all
       else
-        suggestions = with_sql("
-          WITH filtered AS (
-            SELECT DISTINCT ON (value) *
-            FROM search_suggestions
-            WHERE
-              type IN (?, ?) AND
-              value ILIKE ?
-            ORDER BY value, priority
-          )
-          SELECT *,
-                ? AS query,
-                similarity(value, ?) AS score
-          FROM filtered
-          ORDER BY priority ASC, score DESC
-          LIMIT ?;
-        ", 'search_reference', 'full_chemical_name', "%#{query}%", query, query, 10).all
+        filtered_cte = select_all
+          .distinct(:value)
+          .where(type: %w[search_reference full_chemical_name])
+          .where(Sequel.ilike(:value, "%#{query}%"))
+          .order(:value, :priority)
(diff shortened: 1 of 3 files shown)


db/migrate/20250617113942_add_search_suggestion_indexes.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    run <<-SQL
+      CREATE INDEX IF NOT EXISTS idx_search_suggestions_value_trgm
```
willfish, line 6: I think we can also delete the old one which is using gist over gin?
```
CREATE INDEX search_suggestions_value_trgm_idx ON uk.search_suggestions USING gist (value public.gist_trgm_ops);
```

app/models/search_suggestion.rb
```diff
@@ -38,15 +38,22 @@ def fuzzy_search(query)
...
           )
-          .limit(10)
-          .all
+          SELECT *,
+                ? AS query,
+                similarity(value, ?) AS score
+          FROM filtered
+          ORDER BY priority ASC, score DESC
+          LIMIT ?;
+        ", 'search_reference', 'full_chemical_name', "%#{query}%", query, query, 10).all
```
willfish, line 56: I think there's a mismatch between the number of passed parameters and we're also bypassing the builtin sanitisation by going direct for sql :thinking:

willfish:
```ruby
        filtered_cte = select_all
          .distinct(:value)
          .where(type: ['search_reference', 'full_chemical_name'])
          .where(Sequel.ilike(:value, "%#{query}%"))
          .order(:value, :priority)

        suggestions = with(:filtered, filtered_cte)
          .from(:filtered)
          .select_all(:filtered)
          .with_query(query)
          .with_score(query)
          .order(:priority, Sequel.desc(:score))
          .limit(10)
          .all

        suggestions.select do |suggestion|
          # ilike filters can return suggestions with a score of 0
          suggestion[:score] > ILIKE_SIMILARITY_THRESHOLD
        end
```

willfish (dismissed):

PR #2332, opened 2025-06-17
HOS 5 | Build Live Issues v2 and admin endpoints

EraliaKats:
### Jira link

[HOS-5](https://transformuk.atlassian.net/browse/HOS-5)

### What?

I have added/removed/altered:

- [x] Added endpoints for the live issues log feature in api and admin

### Why?

I am doing this because:

- Need this for the new feature


db/structure.sql
```diff
@@ -353,20 +385,20 @@ CREATE TABLE uk.additional_code_description_periods_oplog (
 --
```
HWallenberg, line 385: Hi I think you're db/structure.sql may have too many changes in it.  I normally look at the diffs locally and add the changes I need for the feature.  It will make it easier for others to see your change too.

app/models/live_issue.rb
```diff
@@ -0,0 +1,12 @@
+class LiveIssue < Sequel::Model(Sequel[:live_issues].qualify(:public))
+  plugin :timestamps, update_on_create: true
+  plugin :auto_validations, not_null: :presence
+
+  def validate
+    super
+
+    validates_presence %i[title status commodities date_discovered]
+    validates_max_length 256, %i[description suggested_action]
+    validates_includes %w[Active Resolved], :status
```
EraliaKats, line 10: This is something I'd like to discuss with someone. I've added status like this because in the future we're looking into including 'Archived'. Now I'm thinking that maybe it should a boolean on its own so you can validate if it's resolved before you toggle Archived on. What do people think?

HWallenberg (approved):

PR #2345, opened 2025-06-23
HMRC-1317: Move views to materialized views

rasikasri:
### Jira link

### What?

I have added/removed/altered:

- [ ] Changed additional code views to materialized views

### Why?

I am doing this because:

### DIFF
--- a/app/helpers/materialize_view_helper.rb
+++ b/app/helpers/materialize_view_helper.rb
@@ -6,5 +6,11 @@ module MaterializeViewHelper
     GeographicalArea.refresh!(concurrently:)
     GeographicalAreaDescription.refresh!(concurrently:)
     GeographicalAreaDescriptionPeriod.refresh!(concurrently:)
+    AdditionalCode.refresh!(concurrently:)
+    AdditionalCodeType.refresh!(concurrently:)
+    AdditionalCodeTypeMeasureType.refresh!(concurrently:)
+    AdditionalCodeTypeDescription.refresh!(concurrently:)
+    AdditionalCodeDescription.refresh!(concurrently:)
+    AdditionalCodeDescriptionPeriod.refresh!(concurrently:)
   end
 end
--- a/app/helpers/materialize_view_helper.rb
+++ b/app/helpers/materialize_view_helper.rb
@@ -1,16 +1,11 @@
 module MaterializeViewHelper
   def refresh_materialized_view(concurrently: false)
     GoodsNomenclatures::TreeNode.refresh!(concurrently:)
-    GeographicalAreaMembership.refresh!(concurrently:)
-    MeasureExcludedGeographicalArea.refresh!(concurrently:)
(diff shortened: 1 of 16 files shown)


app/helpers/materialize_view_helper.rb
```diff
@@ -6,5 +6,11 @@ def refresh_materialized_view(concurrently: false)
     GeographicalArea.refresh!(concurrently:)
     GeographicalAreaDescription.refresh!(concurrently:)
     GeographicalAreaDescriptionPeriod.refresh!(concurrently:)
+    AdditionalCode.refresh!(concurrently:)
```
willfish, line 9: We could just register these when the configure method of the oplog plugin is called and then iterate over them

app/models/additional_code.rb
```diff
@@ -29,6 +29,12 @@ class AdditionalCode < Sequel::Model
   one_to_many :measures, key: :additional_code_sid,
                          primary_key: :additional_code_sid
 
+  class << self
+    def refresh!(concurrently: false)
```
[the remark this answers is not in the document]
rasikasri: Done

app/helpers/materialize_view_helper.rb
```diff
@@ -1,10 +1,11 @@
...
   def refresh_materialized_view(concurrently: false)
     GoodsNomenclatures::TreeNode.refresh!(concurrently:)
-    GeographicalAreaMembership.refresh!(concurrently:)
-    MeasureExcludedGeographicalArea.refresh!(concurrently:)
-    GeographicalArea.refresh!(concurrently:)
-    GeographicalAreaDescription.refresh!(concurrently:)
-    GeographicalAreaDescriptionPeriod.refresh!(concurrently:)
+
+    Sequel::Plugins::Oplog.models.each do |model|
+      if model.materialized?
```
willfish, line 6: Awesome

app/lib/sequel/plugins/oplog.rb
```diff
@@ -146,6 +153,14 @@ def insert(*_args)
         def operation_klass
           @operation_klass ||= "#{self}::Operation".constantize
         end
+
+        def refresh!(concurrently: false)
+          if materialized?
+            db.refresh_view(table_name, concurrently:)
+          else
```
willfish, line 160: I'm wondering whether its fine to just skip the raise, here?

spec/rails_helper.rb
```diff
@@ -102,6 +97,16 @@ def silence
   @original_stdout = nil
 end
 
+def refresh_materialized_view(concurrently: false)
+  GoodsNomenclatures::TreeNode.refresh!(concurrently:)
+
+  Sequel::Plugins::Oplog.models.each do |model|
+    if model.materialized?
```
willfish, line 104: You shouldn't need this if check?

willfish (approved):

PR #2361, opened 2025-06-30
HMRC-1317: Move views to materialized views - measures and regulations

rasikasri:
### Jira link

### What?

I have added/removed/altered:

- [ ] Changed measures and regulations views to materialized views

### Why?

I am doing this because:


spec/unit/changes_table_populator/measure_created_or_updated_spec.rb
```diff
@@ -4,32 +4,33 @@
   describe '#populate' do
     context 'when the database is empty' do
       before do
-        db[:measures].delete
```
willfish, line 7: How did this ever work :confused:

willfish:
So this is a big one I think. Are you okay to properly deploy it and run the e2e tests?

PR #2366, opened 2025-07-02
HMRC-1317: Move views to materialized views - add index to materialized views

rasikasri:
### Jira link

### What?

I have added/removed/altered:

- [ ] Add all the indices from corresponding tables to materialized views

### Why?

I am doing this because:

### DIFF
--- a/db/migrate/20250701123907_add_materialized_view_indexes.rb
+++ b/db/migrate/20250701123907_add_materialized_view_indexes.rb
@@ -1,0 +1,231 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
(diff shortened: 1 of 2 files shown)


db/migrate/20250701123907_add_materialized_view_indexes.rb
```diff
@@ -0,0 +1,231 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    alter_table :geographical_areas do
+      add_index :operation_date
```
willfish, line 6: I guess this is for rollbacks

willfish (approved):
Clean and smooth!

PR #2382, opened 2025-07-11
Fix xi Basic Report

vertism:
### Jira link

### What?

Basic report is failing due to the `Measure.supplementary_unit_duty_expression` method
Commodity "8708405025" has a supplementary measure without a measure component which causes an exception.


EraliaKats (approved):
Nice spot!

PR #2419, opened 2025-08-08
HMRC-1404: Basic auth to admin backends when auth disabled

rasikasri:
### Jira link

[HMRC-1404](https://transformuk.atlassian.net/browse/HMRC-1404)

### What?

I have added/removed/altered:

- [ ] Removed disable_admin_api_authentication flag from admin api request authentication for basic user

### Why?

I am doing this because:

- Admin use long living auth token for request auth in backend, it is not related to the logged in user in admin api. So the token can be used regardless of GDS SSO or Basic user auth use in prod and lower env. Token is set up in admin secrets of BEARER_TOKEN

### Have you? (optional)

### DIFF
--- a/app/controllers/api/admin/admin_controller.rb
+++ b/app/controllers/api/admin/admin_controller.rb
@@ -4,10 +4,6 @@ module Api
       include GDS::SSO::ControllerMethods
       include AdminApi.routes.url_helpers
 
-      def authenticate_user!
-        TradeTariffBackend.disable_admin_api_authentication? || super
-      end
-
       no_caching
     end
   end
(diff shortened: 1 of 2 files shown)


AigulDj (approved):

rasikasri:
> ### Jira link
> [HMRC-1404](https://transformuk.atlassian.net/browse/HMRC-1404)
> 
> ### What?
> I have added/removed/altered:
> 
> * [ ]  Removed disable_admin_api_authentication flag from admin api request authentication for basic user
> 
> ### Why?
> I am doing this because:
> 
> * Admin use long living auth token for request auth in backend, it is not related to the logged in user in admin api. So the token can be used regardless of GDS SSO or Basic user auth use in prod and lower env. Token is set up in admin secrets of BEARER_TOKEN
> 
> ### Have you? (optional)
> * [ ]  Added documentation for new apis
> * [ ]  Added new environment variables with correct values to all apps in all spaces
> 
> ### Deployment risks (optional)
> * Includes destructive/migratory actions which may effect critical data
> * Changes an api that is used in production

PR #2427, opened 2025-08-12
HMRC-1472: Generate Changes API response for 1/1/2022 (XI and UK)

rasikasri:
### Jira link

[HMRC-1472](https://transformuk.atlassian.net/browse/HMRC-1472)

### What?

I have added/removed/altered:

- [ ] Filter out measurs with null goods_nomenclature_sid from change report source set

### Why?

I am doing this because:

- Measures with goods_nomenclature_sid causing error in generating change report


HWallenberg (approved):
excellent! 🙌

PR #2439, opened 2025-08-19
MYOTT-72 Create excel file with tariff changes for a given day

vertism:
### Jira link

[MYOTT-72](https://transformuk.atlassian.net/browse/MYOTT-72)

### What?

For a given date, gets details for all tariff changes to the following models:
- GoodsNomenclature
- Measure
- MeasureCondition
- MeasureComponent
- Certificate
- GeographicalArea

Each change is related back to any declarable commodity that is affected and a record is written to a spreadsheet with the details.

### Why?

So users can be made aware of changes to the tariff.


HWallenberg (approved):
Thank you @vertism 

This is a very elegant method of storing the changes by operation date in excel.
 
I like the fact it's a service, for capturing all the deltas.

I'm unsure if the excel will actually be needed for the myott solution, are we just showing the changes on screen?   I hope we do need the spreadsheet as it's had a lot of effort put into it.

A good large step forward for what we are trying to present to the users. 🙏

PR #2444, opened 2025-08-21
HMRC-1519: Drops tradeset descriptions

willfish:
### Jira link

[HMRC-1519](https://transformuk.atlassian.net/browse/HMRC-1519)

### What?

I have added/removed/altered:

- [x] Removed all references to tradeset descriptions

### Why?

I am doing this because:

- These were used for an old prototype building FPO search initially using opensearch in the backend
- These are unused, out of date and all extra code is a liability


app/controllers/api/admin/applies_controller.rb
```diff
@@ -2,11 +2,6 @@ module Api
   module Admin
     class AppliesController < AdminController
       before_action :authenticate_user!
-      before_action :collection, only: :index
```
willfish, line 5: Unrelated but unused code so on theme. Happy to revert if preferred

rasikasri (approved):

PR #2449, opened 2025-08-22
HMRC-1443: Reduce use of long `IN` lists in queries

rasikasri:
### Jira link

[HMRC-1443](https://transformuk.atlassian.net/browse/HMRC-1443)

### What?

I have added/removed/altered:

- [ ] Customized eager loader query in measure/excluded_geographical_areas association to avoid using IN

### Why?

I am doing this because:

- To improve the performance of the query

### DIFF
--- a/app/lib/sequel/plugins/optimized_many_to_many.rb
+++ b/app/lib/sequel/plugins/optimized_many_to_many.rb
@@ -39,17 +39,17 @@ module Sequel
               FROM #{target_table}
               JOIN #{join_table} ON #{join_table}.#{right_key} = #{target_table}.#{right_pk}
               WHERE #{join_table}.#{left_key} = ?
-              #{order ? "ORDER BY #{associated_class.dataset.literal(order)}" : ""}
+              #{order ? "ORDER BY #{associated_class.dataset.literal(order)}" : ''}
             SQL
 
-            associated_class.with_sql(sql, self.send(left_pk))
+            associated_class.with_sql(sql, send(left_pk))
           end
         end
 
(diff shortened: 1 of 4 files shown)


neilmiddleton (dismissed):

app/lib/sequel/plugins/optimized_many_to_many.rb
```diff
@@ -0,0 +1,98 @@
+# lib/sequel/plugins/optimized_many_to_many.rb
+module Sequel
+  module Plugins
+    module OptimizedManyToMany
+      module ClassMethods
+        def many_to_many(name, opts = OPTS)
```
willfish, line 6: Do you mean to set the default OPTS somewhere?

app/models/measure.rb
```diff
@@ -23,6 +23,7 @@ class Measure < Sequel::Model
   plugin :time_machine
   plugin :oplog, primary_key: :measure_sid, materialized: true
   plugin :national
+  plugin :optimized_many_to_many
```
willfish, line 26: You can make this a default plugin https://github.com/trade-tariff/trade-tariff-backend/blob/HMRC-1443/config/application.rb#L52 and then you can easily roll it out everywhere

app/services/cached_commodity_service.rb
```diff
@@ -48,7 +48,6 @@ class CachedCommodityService
       ],
     },
     { quota_order_number: { quota_definition: %i[quota_balance_events quota_suspension_periods quota_blocking_periods] } },
-    { excluded_geographical_areas: :geographical_area_descriptions },
```
willfish, line 51: Is this eager load not needed?

PR #2496, opened 2025-09-17
Add deleted records to the delta report

vertism:
### What?

Add deleted records via the Operation class.
Regular objects are loaded via views that exclude deleted records. By using the Operation class we can access the oplogs directly and load these records into a model.

The models that have deleted records that users may find useful are:
- Measures
- MeasureExcludedGeographicalArea
- FootnoteAssociationMeasure
- MeasureCondition

---
Some tweaks have also been made to ignore some records when created, as they need to be linked to a measure or commodity to be of interest to a user, and this change is already dealt with.


HWallenberg (approved):
🚀

PR #2519, opened 2025-10-01
BAU: Consolidate TimeMachine in search service

willfish:
### Jira link

BAU

image

### What?

I have added/removed/altered:

- [x] Consolidate time machine usage under the search service

### Why?

I am doing this because:

- Previously we were using a custom implementation of the as_of handling that duplicates the existing implementation in the application controller
- This change also fixes a bug where the date is nil if the as_of isn't supplied as a query param since we were relying on the query parameter to be present in order to set even a default date.

### DIFF
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -22,7 +22,7 @@ class SearchService
   class EmptyQuery < StandardError
   end
 
-  attr_reader :q, :result, :as_of, :data_serializer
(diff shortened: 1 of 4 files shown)


willfish:
Download failure in the e2e test is a byproduct of the dev environment being off but having an updated database. Because its off it never pulls new exchange rates but the file rows get populated by a restore from the production database

PR #3014, opened 2026-04-13
[HMRC-2187] Adds varchar_pattern_ops index to improve speed of quota search

spencerldixon:
### Jira link

[HMRC-2187](https://transformuk.atlassian.net/browse/HMRC-2187)

### What?

The `/quotas/search` endpoint has a high average db time when searching with an order number and no other filters.

The existing index is a standard btree on `(ordernumber, validity_start_date)`. Postgres cannot make use of this index for LIKE queries, and falls back to scanning all index entries and filters them out row by row.

This PR adds a `varchar_pattern_ops` index, allowing Postgres to index by byte order, and rewrite the query to an efficient tight range seek, avoiding the full table scan, and making the query significantly more performant.

### Why?

This endpoint was identified as having a high average db time and a high number of requests. Improving this helps us reduce db load, deliver results faster and keep users happy.

### Notes

To test, run this in your rails console before applying the migration, then apply the migration and run it again to see the difference in query times:

```
Sequel::Model.db.loggers << Logger.new($stdout)
start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
QuotaSearchService.new({'order_number' => '050001'}, 1, 10, Date.today).call
elapsed = Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
puts "Time: #{(elapsed * 1000).round}ms"
```

### DIFF
--- a/db/migrate/20260413120000_add_ordernumber_pattern_ops_index_to_measures.rb
+++ b/db/migrate/20260413120000_add_ordernumber_pattern_ops_index_to_measures.rb
@@ -1,0 +1,13 @@
+Sequel.migration do
+  no_transaction
+
+  up do
+    execute "CREATE INDEX CONCURRENTLY IF NOT EXISTS measures_ordernumber_pattern_idx ON uk.measures (ordernumber varchar_pattern_ops)"
+    execute "CREATE INDEX CONCURRENTLY IF NOT EXISTS measures_ordernumber_pattern_idx ON xi.measures (ordernumber varchar_pattern_ops)"
+  end
+
+  down do
+    execute "DROP INDEX CONCURRENTLY IF EXISTS uk.measures_ordernumber_pattern_idx"
+    execute "DROP INDEX CONCURRENTLY IF EXISTS xi.measures_ordernumber_pattern_idx"
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2,15 +2,12 @@
 -- PostgreSQL database dump
 --
 
-\restrict r0VKKTCTLFgbgikL8JJLnq0TRfZfPrUo0zfc6OEwM9PO376VXeZElYPqOlYhdQN
-
--- Dumped from database version 18.3
--- Dumped by pg_dump version 18.3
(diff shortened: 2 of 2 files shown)


db/migrate/20260413120000_add_ordernumber_pattern_ops_index_to_measures.rb
```diff
@@ -0,0 +1,13 @@
+Sequel.migration do
+  no_transaction
+
+  up do
+    execute "CREATE INDEX CONCURRENTLY IF NOT EXISTS measures_ordernumber_pattern_idx ON uk.measures (ordernumber varchar_pattern_ops)"
```
willfish, line 5: If you remove the schema it'll just populate for both

Live environments deploy services and run migrations separately.

db/structure.sql
```diff
@@ -2,15 +2,12 @@
 -- PostgreSQL database dump
 --
 
-\restrict r0VKKTCTLFgbgikL8JJLnq0TRfZfPrUo0zfc6OEwM9PO376VXeZElYPqOlYhdQN
-
--- Dumped from database version 18.3
--- Dumped by pg_dump version 18.3
+-- Dumped from database version 13.23 (Debian 13.23-1.pgdg12+1)
```
willfish, line 5: Interesting version! I used to use debian like 15 years ago 🦜

db/structure.sql
```diff
@@ -515,20 +512,20 @@ CREATE TABLE uk.additional_code_description_periods_oplog (
...
-    additional_code_sid,
-    additional_code_type_id,
-    additional_code,
-    validity_start_date,
-    validity_end_date,
-    oid,
-    operation,
-    operation_date,
-    filename
+ SELECT additional_code_description_periods1.additional_code_description_period_sid,
```
willfish, line 515: I think this qualification will just get removed on the next dump so is just version-based noise

spencerldixon:
Closing this PR in favour of https://github.com/trade-tariff/trade-tariff-backend/pull/3026

PR #3016, opened 2026-04-13
AI-553: Add excluded chapters admin configuration

willfish:
### Jira link

[AI-553](https://transformuk.atlassian.net/browse/AI-553)

### What?

- [x] Add `multi_options` support to admin configuration validation and normalisation
- [x] Seed `interactive_search_excluded_chapters` with chapters 98 and 99
- [x] Apply excluded chapter filtering to guided search retrieval paths
- [x] Add focused model, service, query and rake spec coverage

### Why?

- This exposes the capability and default configurations for a multi options type in the admin configuration and enables dynamic changes as part of a requirement from HMRC

### DIFF
--- a/app/models/admin_configuration.rb
+++ b/app/models/admin_configuration.rb
@@ -31,6 +31,7 @@ class AdminConfiguration < Sequel::Model(Sequel[:admin_configurations].qualify(:
     'expand_search_enabled' => false,
     'expand_model' => NESTED_OPTION_DEFAULTS['expand_model'][:selected],
     'interactive_search_enabled' => true,
+    'interactive_search_excluded_chapters' => %w[98 99].freeze,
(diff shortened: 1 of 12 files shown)


neilmiddleton:
One thing:

The `excluded_chapter?` check in `search_service.rb` calls `AdminConfiguration.multi_options_values` on every call to `hidden?`, which itself gets called per result in `find_by_goods_nomenclature`. Depending on how that method is implemented, this could result in a database hit per result. Worth confirming there's caching in place, or memoising the result within the service call, to avoid that becoming a performance issue at scale.

neilmiddleton (approved):

willfish:
It is actually just called once per search and it is cached in production (though not in staging)

PR #3026, opened 2026-04-14
[HMRC-2187] Improve performance of quota search

spencerldixon:
### Jira link

[HMRC-2187](https://transformuk.atlassian.net/browse/HMRC-2187)

### What?

The /quotas/search endpoint has a high average db time when searching with an order number and no other filters.

The existing index is a standard btree on (ordernumber, validity_start_date). Postgres cannot make use of this index for LIKE queries, and falls back to scanning all index entries and filters them out row by row, causing a high average db time.

Since we don't need to search partial order numbers, and an order number will always be 6 digits, we can remove the offending LIKE, and refactor this for a direct search instead.

This PR:

- refactors to remove the LIKE in favour of a direct where
- adds validation to raise an error for requests where the order number is not 6 digits
- adds test

### Why?

This endpoint was identified as having a high average db time and a high number of requests. Improving this helps us reduce db load, deliver results faster and keep users happy.

### Testing

Run the following on main vs this branch to see the difference in query time

```
Sequel::Model.db.loggers << Logger.new($stdout)
start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
QuotaSearchService.new({'order_number' => '050001'}, 1, 10, Date.today).call
elapsed = Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
puts "Time: #{(elapsed * 1000).round}ms"
```

### DIFF
--- a/app/controllers/api/v2/quotas_controller.rb
+++ b/app/controllers/api/v2/quotas_controller.rb
@@ -26,7 +26,7 @@ module Api
 
       def validate_order_number
         return if params[:order_number].blank?
-        raise NotImplementedError, params[:order_number] unless params[:order_number].match?(/\A\d{6}\z/)
+        raise ActionController::BadRequest, params[:order_number] unless params[:order_number].match?(/\A\d{6}\z/)
       end
 
       def serialized_quota_definitions
--- a/app/controllers/api/v2/quotas_controller.rb
+++ b/app/controllers/api/v2/quotas_controller.rb
@@ -16,11 +16,18 @@ module Api
 
       ALLOWED_INCLUDES = %w[quota_balance_events].freeze
 
+      before_action :validate_order_number, only: :search
+
       def search
         render json: serialized_quota_definitions
       end
 
       private
+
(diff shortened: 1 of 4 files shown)


app/controllers/api/v2/quotas_controller.rb
```diff
@@ -16,12 +16,19 @@ class QuotasController < ApiController
...
+
       def search
         render json: serialized_quota_definitions
       end
 
       private
 
+      def validate_order_number
+        return if params[:order_number].blank?
+        raise NotImplementedError, params[:order_number] unless params[:order_number].match?(/\A\d{6}\z/)
```
willfish: Should this be a 429?

spencerldixon: 400 maybe? bad request? 429 is too many requests

willfish: Sorry I got my status codes mixed up I meant 422

neilmiddleton (approved):

PR #3038, opened 2026-04-16
AI-514: Import and archive Customs Tariff documents

HWallenberg:
### Jira link

[AI-514](https://transformuk.atlassian.net/browse/AI-514)

### What?

This PR lays the foundations for the Customs Tariff notes pipeline. A daily Sidekiq job fetches the UKGT `.docx` document(s) from GOV.UK, extracts chapter notes, section notes, and general rules, and stores them in four new tables ready for operator review and approval.

The pipeline logic lives under a dedicated `CustomsTariffImporter` module in `app/lib/`:

- **`CustomsTariffImporter::DocumentFetcher`** — hits the GOV.UK publication page and downloads all available UKGT `.docx` versions, parsing entry-into-force and publication dates from the attachment link text
- **`CustomsTariffImporter::NotesExtractor`** — opens the `.docx` and extracts chapter notes, section notes, and General Rules of Interpretation in a single pass
- **`CustomsTariffImporter::Importer`** — ties it together: fetch, archive to S3, parse, and persist to the DB; already-imported versions and duplicate content are skipped gracefully, failures are recorded
- **`CustomsTariffImporter::Instrumentation`** — fires `ActiveSupport::Notifications` events at each stage of the pipeline (fetch, parse, import, skip, fail)
- **`CustomsTariffImporter::Logger`** — a `LogSubscriber` attached to the `:customs_tariff_importer` namespace that writes structured JSON log lines for each event, consistent with the pattern used by `SelfTextGenerator` and `LabelGenerator`
- **`ImportCustomsTariffDocumentWorker`** — the Sidekiq worker (default queue) that drives the run and emits `import_run_started` / `import_run_completed` / `import_run_failed` instrumentation events
- Four new migrations and models: `customs_tariff_updates`, `customs_tariff_chapter_notes`, `customs_tariff_section_notes`, `customs_tariff_general_rules`
- `CustomsTariffUpdate` uses the `time_machine` plugin — `validity_start_date` is the entry-into-force date; `validity_end_date` is always null (open-ended)

### Why?

Before operators can review and approve Customs Tariff notes for use in guided search, we need a reliable, versioned store of that content. This PR builds that store.

- We fetch **all** versions from GOV.UK in one run — if a day is missed we catch up gracefully rather than losing older versions
- We use **entry-into-force dates** (not publication dates) to drive when a version becomes applicable, matching how the real-world tariff works
- Pipeline observability follows the established module pattern — `Instrumentation` fires events, `Logger` subscribes and writes structured JSON; nothing is hardcoded to a specific sink


app/models/customs_tariff_update.rb
```diff
@@ -0,0 +1,30 @@
+class CustomsTariffUpdate < Sequel::Model
+  set_primary_key :version
+  unrestrict_primary_key
+
+  plugin :time_machine
+  plugin :timestamps, update_on_create: true
+
+  AWAITING_APPROVAL = 'awaiting_approval'.freeze
```
willfish: Maybe `pending`?

app/services/gov_uk_tariff_document_fetcher.rb
```diff
@@ -0,0 +1,88 @@
+require 'digest'
+require 'net/http'
+
+class GovUkTariffDocumentFetcher
+  PUBLICATION_URL = 'https://www.gov.uk/government/publications/reference-document-for-the-customs-tariff-establishment-eu-exit-regulations-2020'.freeze
+  MAX_REDIRECTS = 5
+
+  VERSION_PATTERN = /UKGT_(\d+\.\d+)\.docx/i
+
+  Result = Struct.new(:content, :url, :version, :checksum, :published_on, :entry_into_force_on, keyword_init: true)
```
willfish: We should probably stop using struct as this is deprecated really/replaced by data classes

app/services/import_customs_tariff_document_service.rb
```diff
@@ -0,0 +1,92 @@
+class ImportCustomsTariffDocumentService
+  S3_KEY_PREFIX = 'data/customs_tariff_documents'.freeze
+
+  Result = Struct.new(:status, :version, :error, keyword_init: true)
```
willfish: Same again re data classes

willfish:
So I'd move to instrumented logging with structured events/timings so that we can create a RED dashboard around this feature/pipeline

db/migrate/20260414113626_create_customs_tariff_updates.rb
```diff
@@ -0,0 +1,25 @@
...
+        String   :status, null: false, default: 'awaiting_approval'
+        String   :source_url, text: true
+        String   :s3_path, text: true
+        String   :file_checksum
+        Date     :document_created_on
+        String   :import_error, text: true
+        DateTime :created_at, null: false
+        DateTime :updated_at, null: false
+
+        index :file_checksum
```
willfish: How comes we're indexing this chipolata?

HWallenberg: we don't need to it's not like we're going to get thousands of these I've removed it. 👍

neilmiddleton:
A few things worth checking before approval:

Will has already flagged structured/instrumented logging -- the suggestion to build a RED dashboard around this pipeline is a good one and worth doing before this ships, rather than bolting it on later.

Error handling: when ImportCustomsTariffDocumentService records a failure, is there a retry mechanism, or does the Sidekiq worker rely on its own retry? Just want to make sure a transient GOV.UK failure doesn't silently drop a version.

Has the .docx parsing been tested against edge cases in the document structure, or are we relying on GOV.UK keeping the format consistent? Worth noting any brittleness in the extractor.

Overall shape is good -- just want to be comfortable on the observability and error handling side before it goes in.

db/migrate/20260414113628_create_customs_tariff_section_notes.rb
```diff
@@ -0,0 +1,18 @@
+Sequel.migration do
+  up do
+    if TradeTariffBackend.uk?
+      create_table :customs_tariff_section_notes do
+        primary_key :id
+        String :customs_tariff_update_version, null: false
+        String :section_id, size: 10, null: false
```
willfish:
```suggestion
        Integer :section_id, size: 10, null: false
```

db/migrate/20260414113629_create_customs_tariff_general_rules.rb
```diff
@@ -0,0 +1,18 @@
+Sequel.migration do
+  up do
+    if TradeTariffBackend.uk?
```
willfish: Probably create in the XI and UK schema as we'll be doing a parser for the XI file soon

app/lib/customs_tariff_importer/document_fetcher.rb
```diff
@@ -0,0 +1,91 @@
...
+      uri = URI(url)
+      response = Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == 'https') do |http|
+        http.get(uri.request_uri)
+      end
+
+      case response
+      when Net::HTTPSuccess
+        response.body
+      when Net::HTTPRedirection
+        fetch_url(response['location'], redirect_count: redirect_count + 1)
```
willfish, line 85: This is a bit odd! I assume a quirk in their site

app/lib/customs_tariff_importer/importer.rb
```diff
@@ -0,0 +1,92 @@
+module CustomsTariffImporter
+  class Importer
+    S3_KEY_PREFIX = 'data/customs_tariff_documents'.freeze
+
+    Result = Data.define(:status, :version, :error) do
+      def initialize(status:, version: nil, error: nil)
```
willfish, line 6: Is this the default?

willfish (approved):
Wunderbar. Exquisite. Magnifique 💃

PR #3039, opened 2026-04-16
AI-554: Add description intercept search capabilities

willfish:
### Jira link

[AI-554](https://transformuk.atlassian.net/browse/AI-554)

```mermaid
flowchart TB
    A[Internal guided search request] --> B{Description intercept matched?}
    B -->|Excluded| C[Return empty results with intercept metadata]
    B -->|Filtering| D[Apply prefixes to exact match, OpenSearch and vector retrieval]
    B -->|Guidance or escalation| E[Return results with intercept metadata]
    B --> F[Emit intercept match instrumentation]
    F --> G[Search dashboard intercept widgets]
    D --> E
```

### What?

- [x] Match description intercepts during internal guided search.
- [x] Return empty results when an intercept excludes the search.
- [x] Surface guidance, location, level and escalation metadata in search responses.
- [x] Apply filter prefixes across exact match, OpenSearch, vector and hybrid retrieval paths.
- [x] Instrument description intercept matches and misses in search logs.
- [x] Extend the search dashboard with intercept match, outcome, top term and recent match widgets.
- [x] Add request, service, retrieval, instrumentation and logger coverage for excluded, filtering, guidance and escalation behaviour.

### Why?

Description intercepts are now configured by the admin APIs, but guided search still needs to act on them. Search needs to stop excluded terms before retrieval, restrict filtered terms consistently across both retrieval paths, and return enough metadata for the frontend to decide how to render guidance or HMRC support escalation.

The dashboard also needs a reliable denominator for intercept usage. Excluded intercepts short-circuit before normal search completion, so a dedicated intercept check event records both matches and misses before the search path branches.

### DIFF
--- a/app/services/api/internal/search_service.rb
+++ b/app/services/api/internal/search_service.rb
@@ -5,7 +5,7 @@ module Api
 
       RetrievalResult = Struct.new(:goods_nomenclatures, :max_score, :expanded_query, :results_type, keyword_init: true)
 
-      attr_reader :q, :as_of, :answers, :request_id
+      attr_reader :q, :as_of, :answers, :request_id, :description_intercept
 
       def initialize(params = {})
         sanitiser_result = InputSanitiser.new(params[:q]).call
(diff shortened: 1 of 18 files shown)


app/services/api/internal/search_service.rb
```diff
@@ -25,21 +25,26 @@ def initialize(params = {})
       def call
         return @sanitiser_errors if @sanitiser_errors
 
-        if q.blank? || ::SearchService::RogueSearchService.call(q)
+        return { data: [] } if q.blank?
+
+        @description_intercept = DescriptionIntercept.for_search(q, source: 'guided_search')
+        return empty_response if description_intercept&.excluded
+
+        if ::SearchService::RogueSearchService.call(q)
```
willfish: This is actually getting replaced by description intercepts which are more sophisticated so we can just remove this from guided search now.

```ruby
class SearchService
  class RogueSearchService
    ROGUE_TERMS = %w[gif gifts gift].freeze

    attr_reader :query_string

    def self.call(query_string)
      new(query_string).call
    end

    def initialize(query_string)
      @query_string = query_string&.downcase
    end

    def call
      ROGUE_TERMS.include?(query_string)
    end
  end
end
```

rasikasri (approved):

PR #3045, opened 2026-04-17
[HMRC-2190] Refactor goods nomenclature query to remove subquery

spencerldixon:
### Jira link

[HMRC-2190](https://transformuk.atlassian.net/browse/HMRC-2190)

### What?

The goods nomenclature endpoint was identified as having a high average db time.

`ancestor_node_constraints` used a correlated subquery (`previous_sibling`) to determine whether a candidate ancestor node's subtree contained the origin node. For each row evaluated in the join, PostgreSQL re-ran a subquery against `goods_nomenclature_tree_nodes` to find the maximum sibling position, scaling poorly for batch queries and for the GreenLanes service, which eager-loads ancestors for entire trees.

```
SubPlan 2
  ->  Result  (actual time=0.003..0.003 rows=1 loops=3030)
Execution Time: 1,371ms
```

This PR rebuilds the materialized view with a `valid_next_siblings` CTE that self-joins `tree_nodes` to find each node's nearest sibling using validity-overlap logic (not just current-date). Produces two new columns per row:

- `next_sibling_or_end_position` - precomputed position of the next sibling at the same depth (`1_000_000_000_000` sentinel if none).
- `next_sibling_validity_start_date`- validity start of that sibling (used for historic query correctness).
- `DISTINCT ON (goods_nomenclature_indent_sid)` ordered by sibling position picks the nearest sibling. A chapter-scoped predicate (`position / 10_000_000_000`) bounds join fanout.
- Adds index on `next_sibling_validity_start_date` for the historic query path.
- Removes `previous_sibling` and spec (no longer used)
- `ancestor_node_contraints` now compares against the precomputed column

Overall reduces query from ~300ms to ~30ms.

### Why?

The goods nomenclature endpoint was identified as having a high average db time during an investigation into service performance. Refactoring to remove the subquery helps reduce query time, deliver services faster and free up load on the database.

### Deployment risks

- This is a big refactor of this query, tests pass and it seems to be working fine, but please triple check it returns the same data!

### DIFF
--- a/app/models/goods_nomenclatures/tree_node.rb
+++ b/app/models/goods_nomenclatures/tree_node.rb
@@ -67,10 +67,10 @@ module GoodsNomenclatures
         # queries fall back to the validity-aware subquery because the view stores
         # one next-sibling row per indent and can't answer "at point_in_time".
         origin_next_sib = if historical_query?
-          next_sibling_or_end(origin.position, origin.depth)
-        else
-          Sequel.qualify(origin.table, :next_sibling_or_end_position)
-        end
+                            next_sibling_or_end(origin.position, origin.depth)
+                          else
+                            Sequel.qualify(origin.table, :next_sibling_or_end_position)
+                          end
 
         (descendants.position > origin.position) &
           validity_dates_filter(date_constraints.table) &
--- a/app/models/goods_nomenclatures/tree_node.rb
+++ b/app/models/goods_nomenclatures/tree_node.rb
@@ -21,21 +21,6 @@ module GoodsNomenclatures
         db.refresh_view(:goods_nomenclature_tree_nodes, concurrently:)
       end
 
-      def previous_sibling(origin_position, origin_depth)
-        siblings_position = Sequel.qualify(:siblings, :position)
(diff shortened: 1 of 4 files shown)


neilmiddleton:
Nice work on this -- the description is thorough and the 10x query time reduction is a meaningful improvement. The approach of precomputing the next-sibling boundary in the materialized view rather than re-running a correlated subquery per row is the right call.

One thing worth addressing before merge: the valid_next_siblings CTE joins all nodes at the same depth across the entire dataset, but the PR description mentions a chapter-scoped predicate (`position / 10_000_000_000`) to bound join fanout. Copilot has already flagged this and it looks like it hasn't been applied yet -- worth adding to keep view refresh cost in check and to prevent unexpected cross-chapter sibling matches.

The `historical_query?` path is a sensible fallback and the `DISTINCT ON` ordering logic looks correct. The removal of `previous_sibling` is clean -- no orphaned call sites that I can see.

Happy to approve once the chapter-scope predicate is in.

willfish (approved):
Approving to enable stale PR cleanup workflows.

PR #3048, opened 2026-04-20
HMRC-2118: Review and optimise the quota search loading strategy

rasikasri:
### Jira link

[HMRC-2118](https://transformuk.atlassian.net/browse/HMRC-2118)

### What?

I have added/removed/altered:

- [ ] Quota search no longer eagerly loads full quota_balance_events by default.
- [ ] QuotaDefinition#last_balance_event no longer requires scanning the full association unless it is already loaded for includes.

### Why?

I am doing this because:

- QuotasController only allows quota_balance_events as an optional include, and default response does not include them.
- last_balance_event currently scans all loaded quota_balance_events, filters by point_in_time, then max_by timestamp.

### DIFF
--- a/app/controllers/api/v2/quotas_controller.rb
+++ b/app/controllers/api/v2/quotas_controller.rb
@@ -53,7 +53,13 @@ module Api
       end
 
       def search_service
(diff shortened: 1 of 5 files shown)


willfish (approved):
💃

PR #3057, opened 2026-04-21
Add covering OID indexes to speed up goods nomenclature description deduplication and Fix copy-paste bug: oplog primary key in GoodsNomenclatureDescriptionPeriod

neilmiddleton:
## Summary

- `GoodsNomenclatureDescriptionPeriod` had `:geographical_area_description_period_sid` as the oplog primary key — a copy-paste from a neighbouring model
- The correct key is `:goods_nomenclature_description_period_sid`, matching the `set_primary_key` declaration on line 5
- This mismatch caused `GoodsNomenclatureDescriptionPeriod::Operation` to have the wrong composite primary key, meaning `previous_record` deduplication queries would reference a non-existent column on the oplog table

- Adds `(goods_nomenclature_description_period_sid, oid DESC)` to `goods_nomenclature_description_periods_oplog`
- Adds `(goods_nomenclature_sid, goods_nomenclature_description_period_sid, oid DESC)` to `goods_nomenclature_descriptions_oplog`

## Root cause

`goods_nomenclature_description_periods` and `goods_nomenclature_descriptions` are plain SQL views that inline a correlated `MAX(oid)` subquery per row to deduplicate the oplog:

```sql
WHERE oid IN (
  SELECT MAX(oid) FROM ...oplog
  WHERE primary_key_col = outer.primary_key_col
)
```

Unlike `measures` (which uses `plugin :oplog, materialized: true` and pre-computes the deduplicated rows into a stored materialized view), these views re-execute the correlated subquery at query time for every candidate row.

When `goods_nomenclature_descriptions` is eager-loaded for a heading or chapter page (~500+ `goods_nomenclature_sid` values in the IN list), the views produce ~1700 period rows and must execute ~3400 correlated subquery executions in total. This shows up as a slow query on commodity/heading detail pages.

## What the indexes do

The existing deduplication lookup indexes support the WHERE clause but force a scan over all matching rows to compute MAX:

| Index | Covers |
|-------|--------|
| `gono_desc_primary_key (goods_nomenclature_description_period_sid)` | WHERE only |
| `gono_desc_pk (goods_nomenclature_sid, goods_nomenclature_description_period_sid)` | WHERE only |

Adding `oid DESC` as a trailing column makes `SELECT MAX(oid) WHERE key = ?` an **index-only forward scan returning exactly one row** — the planner reads the top entry of the `(key, oid DESC)` B-tree and stops immediately, rather than scanning all rows with that key.

## Long-term fix

Converting both models to materialized views (`materialized: true` on `plugin :oplog`) would eliminate the correlated subqueries entirely by pre-computing and storing the deduplicated rows. That is a larger change tracked separately.

### DIFF
--- a/spec/models/goods_nomenclature_description_period_spec.rb
+++ b/spec/models/goods_nomenclature_description_period_spec.rb
@@ -1,0 +1,7 @@
+RSpec.describe GoodsNomenclatureDescriptionPeriod do
+  describe 'oplog primary key' do
+    it 'uses goods_nomenclature_description_period_sid, not the geographical_area equivalent' do
+      expect(described_class::Operation.primary_key).to include(:goods_nomenclature_description_period_sid)
+    end
+  end
+end
--- a/app/models/goods_nomenclature_description_period.rb
+++ b/app/models/goods_nomenclature_description_period.rb
@@ -1,6 +1,6 @@
 class GoodsNomenclatureDescriptionPeriod < Sequel::Model
   plugin :time_machine
-  plugin :oplog, primary_key: :geographical_area_description_period_sid
+  plugin :oplog, primary_key: :goods_nomenclature_description_period_sid
 
   set_primary_key [:goods_nomenclature_description_period_sid]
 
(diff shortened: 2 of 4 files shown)


willfish:
I actually wonder if we can completely get rid of the correlated subquery in future

spec/models/goods_nomenclature_description_period_spec.rb
```diff
@@ -0,0 +1,7 @@
+RSpec.describe GoodsNomenclatureDescriptionPeriod do
+  describe 'oplog primary key' do
+    it 'uses goods_nomenclature_description_period_sid, not the geographical_area equivalent' do
```
willfish, line 3: Not sure we need to reference geographical area, here 🤔

willfish (approved):

PR #3058, opened 2026-04-21
Eliminate N+1 queries on api/v2/commodities#show

neilmiddleton:
## Summary

Six sources of lazy queries eliminated from the commodity show serialization path. On a commodity with ~50 measures these were producing ~100-300 extra queries per cache miss.

### `app/models/measure.rb`

**`Measure#justification_regulation` was firing a raw `.find` per measure** — `ModificationRegulation.find(...)` or `BaseRegulation.find(...)` bypasses Sequel's association cache entirely. Two new `many_to_one` associations are added to mirror the existing `base_regulation` / `modification_regulation` pattern:

```ruby
many_to_one :justification_modification_regulation, class: 'ModificationRegulation', ...
many_to_one :justification_base_regulation,         class: 'BaseRegulation', ...
```

`justification_regulation` now reads from whichever cached association is appropriate.

### `app/services/cached_commodity_service.rb`

| Fix | Root cause |
|-----|-----------|
| `{ measure_type: %i[measure_type_description measure_type_series_description] }` | `MeasureTypeSerializer` renders `measure_type_series_description` but only `measure_type_description` was in the eager graph — 1 query per measure |
| `:justification_base_regulation`, `:justification_modification_regulation` in `MEASURES_PAYLOAD_EAGER_LOAD_GRAPH` | Enables batch loading for the new associations above |
| `goods_nomenclature_descriptions: {}` at the commodity level | `description`, `formatted_description`, etc. delegate through `goods_nomenclature_descriptions.first` — ancestor descriptions were eager-loaded, the commodity's own were not |
| `heading: :footnotes` | `CommodityPresenter` combines `commodity.footnotes + heading.footnotes`; heading was loaded but its footnotes were not |
| `chapter: [:chapter_note, { sections: :section_note }]` | `ChapterSerializer` renders `chapter_note`; `SectionSerializer` renders `section_note` (reached via `chapter.sections.first`) — neither was in any eager load |

## What remains

`MeasureTypeExclusion#find_geographical_areas` can still issue a `GeographicalArea.where(...)` per measure when the CSV file has an exclusion entry matching the measure's `(measure_type_id, geographical_area_id)`. Fixing that requires either pre-loading all referenced GeographicalArea rows at CSV parse time or restructuring the exclusion lookup. Tracked separately.

## Test plan

- [ ] CI passes (existing `justification_legal_act` presenter spec covers the model change)
- [ ] Verify query count in dev with `bullet` or `rack-mini-profiler` — expect significant reduction from ~500 to low double-digits on a commodity with many measures

### DIFF
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -170,6 +170,22 @@ class Measure < Sequel::Model
                                         measure_generating_regulation_role],
                                 conditions: { approved_flag: true }
 
+  # The justification regulation is a separate regulation from the generating
+  # one: it records the legal act that justifies the measure ending. Both
+  # association directions are defined so they can be batch-eager-loaded in the
+  # commodity serialization path, avoiding a raw `.find` per measure.
+  many_to_one :justification_modification_regulation,
+              class: 'ModificationRegulation',
+              primary_key: %i[modification_regulation_id modification_regulation_role],
+              key: %i[justification_regulation_id justification_regulation_role],
+              conditions: { approved_flag: true }
+
+  many_to_one :justification_base_regulation,
+              class: 'BaseRegulation',
+              primary_key: %i[base_regulation_id base_regulation_role],
+              key: %i[justification_regulation_id justification_regulation_role],
+              conditions: { approved_flag: true }
+
   def validity_start_date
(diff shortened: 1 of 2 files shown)


spencerldixon (approved):
sure why not 👍

PR #3059, opened 2026-04-21
Fix three more N+1 query sources on commodities#show

neilmiddleton:
## Summary

Three more lazy-load sources eliminated from the commodity show serialization path, on top of PR #3058.

### `modification_regulation: :base_regulation` (was `:modification_regulation`)

`Measure#legal_acts` appends `generating_regulation.base_regulation` when the generating regulation is a modification regulation — this includes the parent base regulation in the legal acts list rendered by `MeasureLegalActSerializer`. `modification_regulation` was already batch-loaded, but its `one_to_one :base_regulation` (on `ModificationRegulation`) was not. Every modification-regulation-backed measure fired one extra query. Most preference measures are backed by modification regulations.

### `certificate: %i[certificate_descriptions exempting_certificate_override]`

`MeasureCondition#is_exempting_certificate_overridden?` accesses `certificate.exempting_certificate_override`. This is called from `MeasureConditionPermutations::Calculator` for every condition in a measure. The existing eager load had `{ certificate: :certificate_descriptions }` but not `exempting_certificate_override`. Every condition with a certificate fired a separate lookup. `GreenLanes::FetchGoodsNomenclatureService` already loads this correctly — this brings the commodity path in line.

### `geographical_area: [:contained_geographical_areas, { referenced: :contained_geographical_areas }]`

`Measure::GeographicalRelevance#contained_area_ids` resolves:

```ruby
(measure.geographical_area.referenced.presence || measure.geographical_area)
  .contained_geographical_areas
  .pluck(:geographical_area_id)
```

The filter meta graph previously loaded `geographical_area: :contained_geographical_areas` but not `referenced`. For any measure whose geographical area is a reference to another area (common for UK regional preference groups like `1011`, `1013` etc.), this fired two extra queries per measure — one for the `referenced` GeographicalArea and one for its `contained_geographical_areas`.

### DIFF
--- a/app/models/measure_type_exclusion.rb
+++ b/app/models/measure_type_exclusion.rb
@@ -8,12 +8,18 @@ class MeasureTypeExclusion
 
   class << self
     def load_from_file(file = SOURCES.fetch(TradeTariffBackend.service))
-      exclusions_by_service[TradeTariffBackend.service] = parse_file(file)
+      service = TradeTariffBackend.service
+      exclusions_by_service[service] = parse_file(file)
+      # Clear the geo area cache so it is rebuilt lazily on the next
+      # find_geographical_areas call. We cannot preload here because
+      # load_from_file may run before DB records exist (e.g. in tests).
+      geographical_areas_by_service.delete(service)
       self
     end
 
     def reset_data
       @exclusions_by_service = nil
+      @geographical_areas_by_service = nil
       self
     end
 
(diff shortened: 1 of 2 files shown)


willfish (commented):
Not sure we need the comments, but otherwise looks awesome to me

willfish (approved):

neilmiddleton:
> Not sure we need the comments, but otherwise looks awesome to me

The comments are there purely for future us to not fall into the same hole again

PR #3061, opened 2026-04-22
HMRC-2110: Enable database migration step in deployments

amberstarlight:
### Jira link

[HMRC-2110](https://transformuk.atlassian.net/browse/HMRC-2110)

### What?

I have added/removed/altered:

- Enabled `migrate` in deployment actions

### Why?

I am doing this because:

- This workflow has been tested in development and runs database migrations as an explicit pre-deployment task.

### DIFF
--- a/.github/workflows/deploy-to-development-full.yml
+++ b/.github/workflows/deploy-to-development-full.yml
@@ -45,6 +45,7 @@ jobs:
           }
         ]
       test-flavour: tariff
+      migrate: true
     secrets:
       ssh-key: ${{ secrets.PRIVATE_SSH_KEY }}
       slack-webhook: ${{ secrets.SLACK_WEBHOOK }}
--- a/.github/workflows/deploy-to-development.yml
+++ b/.github/workflows/deploy-to-development.yml
@@ -24,6 +24,7 @@ jobs:
       app-name: tariff-backend
       environment: development
       test-flavour: none
+      migrate: true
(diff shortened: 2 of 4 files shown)


neilmiddleton (approved):
A few things:

Is the migration step configured to run before the new ECS service tasks are started, not concurrently? The key risk with automated migrations in a rolling deployment is new code starting on old schema, or old code running against new schema during the rollover window. The migration task should complete and exit cleanly before any new application containers receive traffic.

Does the migration step fail the deployment if it exits non-zero? The `run-task` script does propagate the exit code, but worth confirming the workflow checks it and stops the deployment rather than continuing to a potentially broken state.

Is there a mechanism to prevent concurrent migration runs if two deployments are triggered in quick succession? A migration running twice simultaneously against a non-idempotent migration could cause issues.

Given this is the first time automated migrations are running in CI for this service, worth keeping a close eye on the first few production deploys to confirm the migration task is completing within the ECS wait timeout and not causing deployment delays.

amberstarlight:
The steps for this migration are:

- deploy the new code to _only_ the `backend-job` task,
- run database migrations using `backend-job`,
- if successful, deploy backend services (where ECS may rollback if the deployment is not successful)
- if the migrate job doesn't exit cleanly, downstream actions are cancelled

We currently don't have any deployment queueing; this would be the next step to avoid concurrent deployments clashing.

PR #3063, opened 2026-04-22
HMRC-2119: Add targeted cache warming for the hottest commodity response

rasikasri:
### Jira link

[HMRC-2119](https://transformuk.atlassian.net/browse/HMRC-2119)

### What?

I have added/removed/altered:

- [ ] Added PrewarmCommoditiesWorker logic to fetch top requested goods_nomenclature_item_id values from CloudWatch (result_selected search logs)
- [ ] Added merge of preconfigured commodity IDs from PREWARM_COMMODITY_IDS with log-derived IDs, with deduplication
- [ ] Bulk-load commodities by_codes before warming cache

### Why?

I am doing this because:

- we want to prewarm commodity cache using real usage patterns from production search activity
- we also need deterministic prewarming for known high-priority commodity codes via configuration

### DIFF
--- a/app/workers/clear_cache_worker.rb
+++ b/app/workers/clear_cache_worker.rb
@@ -15,6 +15,7 @@ class ClearCacheWorker
 
     Sidekiq::Client.enqueue(PrecacheHeadingsWorker, Time.zone.today.to_formatted_s(:db))
     Sidekiq::Client.enqueue(PrewarmQuotaOrderNumbersWorker)
+    Sidekiq::Client.enqueue(PrewarmCommoditiesWorker)
     Sidekiq::Client.enqueue(ReindexModelsWorker)
 
     # NOTE: Make sure caches have been refreshed before invalidating the CDN
--- a/app/workers/prewarm_commodities_worker.rb
+++ b/app/workers/prewarm_commodities_worker.rb
@@ -1,0 +1,119 @@
+class PrewarmCommoditiesWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :sync, retry: true
+
+  DEFAULT_LOOKBACK_HOURS = 24
+  DEFAULT_LIMIT = 1000
+  QUERY_POLL_INTERVAL_SECONDS = 1
+  QUERY_MAX_POLLS = 60
+
+  def perform(log_group_name = ENV['SEARCH_LOG_GROUP_NAME'])
+    preconfigured_ids = preconfigured_goods_nomenclature_item_ids
+    most_requested_ids = if log_group_name.present?
+                           most_requested_goods_nomenclature_item_ids(
+                             log_group_name:,
+                             lookback_hours: DEFAULT_LOOKBACK_HOURS,
+                             limit: DEFAULT_LIMIT,
+                           )
+                         else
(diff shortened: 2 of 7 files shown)


app/workers/prewarm_commodities_worker.rb
```diff
@@ -0,0 +1,117 @@
+class PrewarmCommoditiesWorker
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :sync, retry: true
```
willfish, line 4: Worst case it blocks for 60s polling CloudWatch, then warms up to 1000 commodity responses before the reindex worker gets a look-in. That also makes the fixed 1 minute CDN invalidation a bit optimistic.

app/workers/prewarm_commodities_worker.rb
```diff
@@ -0,0 +1,117 @@
...
+  def preconfigured_goods_nomenclature_item_ids
+    ENV.fetch('PREWARM_COMMODITY_IDS', '')
+       .split(',')
+       .map(&:strip)
+       .reject(&:empty?)
+       .uniq
+  end
+
+  def most_requested_goods_nomenclature_item_ids(log_group_name:, lookback_hours:, limit:)
+    query_id = self.class.client.start_query(
```
willfish, line 75: **This needs the task role wiring as well.** The app role currently only has log write actions. `StartQuery`/`GetQueryResults` need `logs:StartQuery` and `logs:GetQueryResults`, otherwise prod will hit AccessDenied here and quietly fall back to the configured IDs only.

app/workers/clear_cache_worker.rb
```diff
@@ -15,6 +15,7 @@ def perform
 
     Sidekiq::Client.enqueue(PrecacheHeadingsWorker, Time.zone.today.to_formatted_s(:db))
     Sidekiq::Client.enqueue(PrewarmQuotaOrderNumbersWorker)
+    Sidekiq::Client.enqueue(PrewarmCommoditiesWorker)
```
willfish, line 18: Can we add the ClearCacheWorker expectation for this enqueue? The existing spec asserts the other warmers get queued, so this should be covered too.

spec/workers/prewarm_commodities_worker_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  let(:cached_commodity_service) { instance_double(CachedCommodityService, call: true) }
+
+  before do
+    allow(described_class).to receive(:client).and_return(client)
+    allow(client).to receive_messages(start_query: start_query_response, get_query_results: query_results_response)
+    allow(Commodity).to receive_message_chain(:actual, :by_codes).and_return(query_scope)
+    allow(CachedCommodityService).to receive(:new).and_return(cached_commodity_service)
+    allow(TimeMachine).to receive(:now).and_yield
+    allow(ENV).to receive(:fetch).and_call_original
+    allow(ENV).to receive(:fetch).with('PREWARM_COMMODITY_IDS', '').and_return('')
```
willfish, line 29: These specs mock out most of the risky bits. Could we cover CloudWatch failure/polling and the generated query? That is where this will break in prod if IAM/config is wrong.

willfish (commented):
A few things to tighten up before this is safe in prod.

rasikasri:
Tested in dev

Screenshot 2026-04-23 at 11 57 44

willfish (approved):

PR #3069, opened 2026-04-23
AI-596: Exclude approved content from review listings

willfish:
### Jira link

[AI-596](https://transformuk.atlassian.net/browse/AI-596)

### What?

- [x] Exclude approved self-texts from normal generated-content listings by default
- [x] Exclude approved labels from normal generated-content listings by default
- [x] Preserve explicit `status=approved` filters for self-texts and labels
- [x] Keep approved self-texts and labels searchable by commodity code or text

### Why?

Approved generated content should not keep returning to normal review work. Operators can still inspect approved rows deliberately using the approved filter or search when investigating a specific commodity or query.

### DIFF
--- a/app/controllers/api/admin/goods_nomenclature_labels_controller.rb
+++ b/app/controllers/api/admin/goods_nomenclature_labels_controller.rb
@@ -37,7 +37,19 @@ module Api
           .for_status(params[:status])
           .for_score_category(params[:score_category])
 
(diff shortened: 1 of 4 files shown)


neilmiddleton (approved):
Only thing I can see is the `normal_review_listing?` and `search_query?` methods are duplicated identically across both controllers. This is low-risk given there are only two, but worth extracting into a shared concern or base class if a third controller ends up needing the same logic.

willfish:
I'm thinking I'll consolidate these completely in a follow on PR

PR #3070, opened 2026-04-23
BAU: Remove init containers

amberstarlight:
### What?

I have added/removed/altered:

- Removed init containers from Terraform ECS services
- Bumped pre-commit hooks

### Why?

I am doing this because:

- Now that database migrations for the backend run in the deployment pipeline, the init containers are effectively a no-op, and can be removed. This should speed up application deployment.

### Deployment risks

- Removing the init containers should be a no-op as schema migrations are run in the pipeline, however we should continue doing best practices for safe schema changes; that is, removing code before removing tables such that there is less chance of deploying a broken application.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
         args: ["--severity=warning"]
 
   - repo: https://github.com/trufflesecurity/trufflehog
-    rev: v3.93.6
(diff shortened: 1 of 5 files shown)


willfish:
Do we want to do this for the other services with migrations:

- dev-hub
- admin

willfish (approved):

PR #3079, opened 2026-04-23
AI-556: Add V2 description intercepts endpoint

willfish:
## What
- add a new V2 `description_intercepts#index` endpoint
- support `source` and `excluded` filters for public API consumers
- add a dedicated V2 serializer that exposes the full description intercept attribute set
- add request and Swagger coverage for the new endpoint

## Why
FPO needs a V2 description intercepts endpoint with a stable JSON:API response, without coupling consumers to the admin API or guided-search flow.

### DIFF
--- a/app/models/description_intercept.rb
+++ b/app/models/description_intercept.rb
@@ -20,6 +20,12 @@ class DescriptionIntercept < Sequel::Model
       where(Sequel.lit('? = ANY(sources)', source))
     end
 
+    def matching_excluded(value)
+      return self if value.nil? || value == ''
+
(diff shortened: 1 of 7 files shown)


app/models/description_intercept.rb
```diff
@@ -20,6 +20,12 @@ def for_source(source)
       where(Sequel.lit('? = ANY(sources)', source))
     end
 
+    def matching_excluded(value)
+      return self if value.nil? || value == ''
```
willfish:
```suggestion
      return self if value.blank?
```

vertism (approved):

PR #3082, opened 2026-04-24
Materialise footnote and measure hot-path plain views

neilmiddleton:
## Summary

- Converts five high-traffic plain views to materialized views: `footnotes`, `footnote_description_periods`, `footnote_descriptions`, `measure_conditions`, `measure_components`
- As plain views, every row evaluation triggered a correlated `MAX(oid)` subquery against the backing oplog table — this happened on every commodity, heading, and subheading show request via `MEASURES_EAGER_LOAD_GRAPH`
- Materializing pre-computes the deduplicated rows so all subsequent reads use conventional index scans with no subquery overhead
- Adds targeted indexes to each view covering the primary eager-load query patterns (FK `IN`-list lookups and time-machine validity date filters)
- Handles the `simplified_procedural_code_measures` plain view which depends on `measure_components` — dropped and recreated around the conversion

Refresh is already fully wired: `MaterializeViewHelper#refresh_materialized_view` iterates all `materialized: true` oplog models and is called at the end of `TaricUpdatesSynchronizerWorker`, `CdsUpdatesSynchronizerWorker`, `ApplyWorker`, and `RollbackWorker`. No changes to the sync pipeline are needed.

### DIFF
--- a/app/models/footnote.rb
+++ b/app/models/footnote.rb
@@ -1,6 +1,6 @@
 class Footnote < Sequel::Model
   plugin :time_machine
-  plugin :oplog, primary_key: %i[footnote_type_id footnote_id]
+  plugin :oplog, primary_key: %i[footnote_type_id footnote_id], materialized: true
 
   set_primary_key %i[footnote_type_id footnote_id]
 
--- a/app/models/footnote_description.rb
+++ b/app/models/footnote_description.rb
@@ -4,7 +4,7 @@ class FootnoteDescription < Sequel::Model
   plugin :time_machine
   plugin :oplog, primary_key: %i[footnote_description_period_sid
                                  footnote_id
-                                 footnote_type_id]
+                                 footnote_type_id], materialized: true
 
   set_primary_key %i[footnote_description_period_sid footnote_id footnote_type_id]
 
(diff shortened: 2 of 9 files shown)


neilmiddleton:
The latest commit reverts the five newly-materialized views back to plain views in `structure.sql`.

**Why**: `REFRESH MATERIALIZED VIEW` only sees committed data. CI runs tests under DatabaseCleaner's `:transaction` strategy, which wraps every test in an open (never-committed) transaction. So when `_refresh_get` triggered a refresh mid-test the just-inserted record was invisible, causing `national_footnote.reload.description` to appear unchanged and `Footnote.last` to return `nil`.

Plain views read straight through the oplog tables within the current transaction, so in-progress inserts are immediately visible — which is the behaviour those tests depend on.

**Question**: does production deployment apply migrations on top of the running schema, or does it load from `structure.sql` (i.e. wipe and reload)? The answer affects whether this approach is safe:

- If migrations are applied incrementally (the common Rails approach), the five views will be properly materialized in production even though `structure.sql` still shows them as plain.
- If deployment loads from `structure.sql` directly, the materialized-view migration would never take effect in production and this PR wouldn't deliver the performance win it's designed for.

willfish (approved):

willfish (approved):

PR #3083, opened 2026-04-24
OTTIMP-530 Refactor duty expression formatters

vertism:
Splits the monolithic `DutyExpressionFormatter` and `RequirementDutyExpressionFormatter` classes into smaller, focused collaborator objects and extracts shared logic into reusable modules.

#### Structural decomposition
Each formatter is now split into three collaborators:
- **`Context`** — builds and validates input data
- **`OutputBuilder`** — assembles output fragments
- **`Strategies`** — one class per duty expression type

This replaces the original single-class implementations.

#### Shared modules extracted
- **`DutyExpressionPrettifier`** — `prettify(float)` was duplicated across all three formatters; now a single shared module
- **`FormatterOutputHelpers`** — HTML fragment helpers (`formatted_amount_fragment`, `measurement_unit_abbr_tag`, `render_measurement_unit_fragment`, `render_prefixed_measurement_unit_fragment`) were duplicated; now shared
- **`DutyExpressionFormattable`** — `duty_expression_formatter_options`, `formatted_duty_expression`, `verbose_duty_expression`, and `duty_expression_str` were duplicated between `Componentable` and `MeasureConditionComponentPresenter`; extracted into a concern included by both

#### Caller clean-up
- Removed a redundant `currency:` kwarg from `RequirementDutyExpressionFormatter.format(...)` in `MeasureCondition`

#### Collaborator specs added
- `DutyExpressionFormatter::Context`, `OutputBuilder`, and `Strategies`
- `RequirementDutyExpressionFormatter::Context` and `OutputBuilder`

## Ticket
https://transformuk.atlassian.net/browse/OTTIMP-530

## Risk: 
**Risk level:** 🟢 

**Reason for rating:**
Refactor with full test coverage and no behaviour change

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -265,6 +265,7 @@ GEM
       ast (~> 2.4.1)
       racc
     pg (1.6.3)
+    pg (1.6.3-arm64-darwin)
     pp (0.6.3)
       prettyprint
     prettyprint (0.2.0)
@@ -477,6 +478,7 @@ GEM
     zeitwerk (2.7.5)
 
 PLATFORMS
+  arm64-darwin-24
   arm64-darwin-25
   x86_64-linux
 
--- a/app/formatters/duty_expression_formatter.rb
+++ b/app/formatters/duty_expression_formatter.rb
@@ -10,122 +10,16 @@ class DutyExpressionFormatter
     end
 
     def format(opts = {})
-      duty_expression_id = opts[:duty_expression_id]
-      duty_expression_description = opts[:duty_expression_description]
-      duty_expression_abbreviation = opts[:duty_expression_abbreviation]
-      duty_amount = opts[:duty_amount]
-      monetary_unit = opts[:monetary_unit_abbreviation].presence || opts[:monetary_unit]
-      measurement_unit = opts[:measurement_unit]
-      measurement_unit_qualifier = opts[:measurement_unit_qualifier]
-      measurement_unit_abbreviation = measurement_unit.try(:abbreviation,
-                                                           measurement_unit_qualifier:)
-      resolved_meursing_component = opts[:resolved_meursing]
-      formatted = opts[:formatted]
-      verbose = opts[:verbose]
-
-      if verbose
-        measurement_unit_expansion = measurement_unit.try(
-          :expansion,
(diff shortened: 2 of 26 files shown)


neilmiddleton (dismissed):
A few observations:

The `measurement_unit_render_options` hash in `Strategies::Base` still passes both `abbreviation:` and `unformatted:` with the same value (`context.measurement_unit_abbreviation`). The PR description says the redundant `unformatted:` parameter was removed, but it's still present in the base class. Worth confirming whether `FormatterOutputHelpers` actually uses `unformatted:` downstream, and if not, it can come out.

The `Context.build` method still has a fairly long parameter list passed through `verbose_context_values`. It's not a blocker but a value object or keyword-forwarding approach would make it easier to add new context fields in future without touching the argument list.

New collaborator specs look well structured. It would be worth adding at least one integration-level test per strategy that exercises the full `DutyExpressionFormatter.format(...)` path for a representative set of duty expression IDs, to give confidence the decomposition hasn't shifted any output.

willfish (dismissed):

vertism:
> A few observations:
> 
> The `measurement_unit_render_options` hash in `Strategies::Base` still passes both `abbreviation:` and `unformatted:` with the same value (`context.measurement_unit_abbreviation`). The PR description says the redundant `unformatted:` parameter was removed, but it's still present in the base class. Worth confirming whether `FormatterOutputHelpers` actually uses `unformatted:` downstream, and if not, it can come out.

The codepath where it is used is not actually hit so it could potentially be removed, but I'm going to leave for now so that we aren't making any behaviour changes in this PR.

> 
> The `Context.build` method still has a fairly long parameter list passed through `verbose_context_values`. It's not a blocker but a value object or keyword-forwarding approach would make it easier to add new context fields in future without touching the argument list.

This method is only used in this one place and as it is unlikely to change I'm going to leave it as is.

> New collaborator specs look well structured. It would be worth adding at least one integration-level test per strategy that exercises the full `DutyExpressionFormatter.format(...)` path for a representative set of duty expression IDs, to give confidence the decomposition hasn't shifted any output.

Integration tests added

neilmiddleton (approved):

PR #3086, opened 2026-04-27
Prewarm base_regulations after materialized view refresh

neilmiddleton:
## Summary

The commodity query plan always executes a Hash Left Join over `base_regulations`, requiring a full sequential scan of ~5,600 rows on every request. After a nightly sync refreshes the materialized view, those pages are cold. `PrewarmCommoditiesWorker` then fires many concurrent commodity requests that all seq-scan `base_regulations` from disk simultaneously, causing I/O contention — observed as queries taking up to 11 seconds.

Adding a `prewarm_views` step at the end of `refresh_materialized_view` runs `count(*)` on `base_regulations` and `modification_regulations` immediately after refresh. This is a single sequential read of each view that loads all pages into PostgreSQL's `shared_buffers` before any application traffic hits them.

The cost is negligible (~5ms per view in the sync worker, which already takes seconds).

### DIFF
--- a/app/helpers/materialize_view_helper.rb
+++ b/app/helpers/materialize_view_helper.rb
@@ -1,4 +1,12 @@
 module MaterializeViewHelper
+  # Tables the commodity query plan always seq-scans via Hash Left Join.
+  # Loading them into shared_buffers here means they are warm before
+  # PrewarmCommoditiesWorker fires HTTP requests that hit them concurrently.
+  VIEWS_TO_PREWARM = %w[
+    BaseRegulation
+    ModificationRegulation
+  ].freeze
+
   def refresh_materialized_view(concurrently: false)
     eager_load_materialized_view_models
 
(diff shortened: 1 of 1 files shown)


amberstarlight:
4398758 appears to be almost a straight revert of #2339, this PR should ideally also change the text of that report email so that the body reflects the change: `app/views/action_log_mailer/daily_report.text.erb`

Edited: I was confused by the two commits but they make sense in a wider context

neilmiddleton:
> [4398758](https://github.com/trade-tariff/trade-tariff-backend/commit/439875851a8a1696e37d43d8fd2be8d93acebc53) appears to be almost a straight revert of #2339, this PR should ideally also change the text of that report email so that the body reflects the change: `app/views/action_log_mailer/daily_report.text.erb`
> 
> Edited: I was confused by the two commits but they make sense in a wider context

There's an extraneous commit I removed because I'm bad at managing branches

amberstarlight (approved):
Looks good to me :)

PR #3087, opened 2026-04-27
Replace deprecated rswag swagger_* config with openapi_* equivalents

neilmiddleton:
## What:

Rename three deprecated rswag-specs configuration keys in `spec/swagger_helper.rb`:
- `swagger_root` → `openapi_root`
- `swagger_docs` → `openapi_specs`
- `swagger_format` → `openapi_format`

Also fix a trailing whitespace in `.github/pull_request_template.md` (line 10) and guard `MaterializeViewHelper.refresh_materialized_view` against calling `refresh!` on models that are declared `materialized` but whose view has not yet been materialised in the current database.

## Why:

The three `swagger_*` settings print deprecation warnings on every test run and will be removed in rswag-specs 3.0. The `openapi_*` replacements are direct renames with no behaviour change.

The `refresh!` guard prevents a `NotImplementedError` in the `before(:suite)` hook when a model has `materialized: true` but the test database schema does not yet have the corresponding materialized view (e.g. after a schema load against an older `structure.sql`).

## Ticket:

N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:** Test-only and schema-helper changes with no production behaviour change.

### DIFF
--- a/app/helpers/materialize_view_helper.rb
+++ b/app/helpers/materialize_view_helper.rb
@@ -11,7 +11,7 @@ module MaterializeViewHelper
     eager_load_materialized_view_models
 
     Sequel::Plugins::Oplog.models.each do |model|
-      if model.materialized?
+      if model.materialized? && model.actually_materialized?
         model.refresh!(concurrently:)
       end
     end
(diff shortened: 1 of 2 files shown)


spencerldixon (approved):
LGTM 👍

PR #3096, opened 2026-04-28
Fix commodity cache fragmentation from meursing additional code

neilmiddleton:
## What:
Only include `meursing_additional_code_id` in the `CachedCommodityService` cache key when the commodity actually has meursing measures (measure types 672, 673, 674 — agricultural duty add-ons for the XI tariff).

## Why:
The cache key previously included the meursing code unconditionally. Meursing codes only change the serialized response for compound agricultural products (those with measures of type 672/673/674). For all other commodities — vehicles, electronics, clothing — the response is identical regardless of what meursing code the client sends.

This meant every distinct meursing code seen in a request produced a separate cache entry for the same commodity, even though those entries were byte-for-byte identical. With ~200 valid meursing codes in the XI tariff, a popular commodity like `8703331900` (cars) could accumulate hundreds of redundant `Cache::RedisCacheStore/generate` events — exactly the symptom seen in APM.

The fix introduces `effective_meursing_code`: if the commodity has no meursing measures, the effective code is `nil` and all requests collapse to the canonical cache entry. The `commodity_has_meursing_measures?` check is itself cached for 24 hours per commodity+date (one Redis lookup per request, one DB query ever), so the overhead is minimal.

## Ticket:
N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:** Pure cache key change. Correct meursing substitution still happens for commodities that need it (agricultural products); no data is suppressed. The worst case on deploy is a brief burst of cache misses while entries warm under the canonical key.

### DIFF
--- a/app/services/cached_commodity_service.rb
+++ b/app/services/cached_commodity_service.rb
@@ -148,7 +148,33 @@ class CachedCommodityService
     TradeTariffRequest.meursing_additional_code_id
   end
 
+  # Only fragment the cache by meursing code when the commodity actually has
+  # meursing measures (types 672, 673, 674 — agricultural duty add-ons).
+  # Most commodities (e.g. vehicles, electronics) never have these, so
+  # including the code in the key would create N identical cache entries —
+  # one per distinct meursing code seen in requests — for no benefit.
+  def effective_meursing_code
+    return nil if meursing_additional_code_id.blank?
+
+    commodity_has_meursing_measures? ? meursing_additional_code_id : nil
+  end
+
+  # Cached per commodity+date so we pay at most one DB query per commodity
+  # per day. The result is stable within a trading day; the 24-hour TTL
+  # matches the commodity response TTL for today's date.
+  def commodity_has_meursing_measures?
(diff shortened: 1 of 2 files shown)


app/services/cached_commodity_service.rb
```diff
@@ -148,7 +148,33 @@ def meursing_additional_code_id
...
+    commodity_has_meursing_measures? ? meursing_additional_code_id : nil
+  end
+
+  # Cached per commodity+date so we pay at most one DB query per commodity
+  # per day. The result is stable within a trading day; the 24-hour TTL
+  # matches the commodity response TTL for today's date.
+  def commodity_has_meursing_measures?
+    Rails.cache.fetch(
+      "commodity-has-meursing-#{@commodity_sid}-#{actual_date}",
+      expires_in: 24.hours,
```
amberstarlight:
```suggestion
      expires_in: ttl,
```

If we use the `ttl` method, this can be aware of the 3-day rolling window for caching

amberstarlight (approved):

PR #3103, opened 2026-04-30
Fix RefreshAppendix5aGuidanceWorker job death from SES v2 single-address bug

neilmiddleton:
## What:
Two targeted fixes for the `ArgumentError` that kills `RefreshAppendix5aGuidanceWorker` when Appendix 5a changes are detected:

1. **Prepend a fix onto `Aws::ActionMailer::SESV2::Mailer`** — wraps the return of `to_addresses` in `Array()` so the AWS SDK always receives an Array, regardless of how many recipients the `mail` gem produces.
2. **Fix `TradeTariffBackend.cupid_team_to_emails`** — splits the env var on commas so multi-address configuration works and the method always returns an Array.

## Why:
The `mail` gem normalises a single-recipient `To:` header to a plain `String` (not an `Array`). The `aws-actionmailer-ses` gem's `SESV2::Mailer#to_addresses` passes `message.to` directly to the AWS SDK, which validates:

```
ArgumentError: expected params[:destination][:to_addresses] to be an Array,
got class String instead
```

`CUPID_TEAM_TO_EMAILS` contains one address. Every time Appendix 5a changes are detected and the mailer fires, the worker exhausts retries and dies.

Fix (1) is the primary defence — it handles the String/Array/nil ambiguity at the delivery layer. Fix (2) is belt-and-braces: correctly splitting a comma-separated env var is the right behaviour regardless, and makes multi-recipient configuration actually work.

## Ticket:
N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:** Narrowly scoped to the Appendix 5a notification mailer. The prepend only overrides one private method on the SES v2 delivery class; all other mailers are unaffected. No data or schema changes.

### DIFF
--- a/app/lib/trade_tariff_backend/config.rb
+++ b/app/lib/trade_tariff_backend/config.rb
@@ -179,7 +179,7 @@ module TradeTariffBackend
     end
 
     def cupid_team_to_emails
-      ENV['CUPID_TEAM_TO_EMAILS']
+      ENV['CUPID_TEAM_TO_EMAILS']&.split(',')&.map(&:strip) || []
     end
 
     def myott_report_email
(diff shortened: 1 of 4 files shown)


HWallenberg (approved):
lgtm 🙌

PR #3115, opened 2026-05-05
Cache small reference tables in process memory with static_cache

neilmiddleton:
## Summary

- Adds Sequel's `plugin :static_cache, frozen: false` to 15 small reference models (duty expressions, measure types, measurement units, monetary units, measure actions, measure condition codes and their description tables)
- These tables contain < 100 rows each and change only when a daily TARIC/CDS sync applies new data
- `Model.all`, `Model.count`, and `Model[pk]` are now served from an in-memory hash instead of hitting PostgreSQL on every call
- `MaterializeViewHelper#refresh_materialized_view` calls the new `reload_static_caches` method after each MV refresh to keep the in-memory copies current
- The plugin is skipped in the test environment to avoid breaking between-example DB isolation

## Why

These reference tables are accessed repeatedly across requests (serializers, presenters, search). Before this change every `Model.all` or primary-key lookup issued a SQL query. With `static_cache` the data lives in the Ruby process and is repopulated once per daily sync.

`frozen: false` is used so that the tariff sync process can still persist new/updated records — the default `frozen: true` would cause `before_save` to cancel any write attempt on a cached instance.

## Multi-process behaviour

Each Ruby process (web server, Sidekiq worker) holds its own in-memory copy. On boot, `load_cache` populates it from the DB. After the daily sync, `reload_static_caches` refreshes the cache in whichever process runs `refresh_materialized_view` (a Sidekiq worker) — but other running web-server processes are not notified and will serve their boot-time copy until they are restarted.

This is acceptable because these reference tables change very rarely (major TARIC releases only), and commodity responses are already cached in Redis so stale in-memory data only affects uncached requests. Cross-process invalidation (e.g. Redis pub/sub) would be needed if same-day consistency becomes a requirement.

## Risk:
**Risk level:** 🟢 
Memoisation of the small, regularly used items only updated with the tariff.

### DIFF
--- a/spec/helpers/materialize_view_helper_spec.rb
+++ b/spec/helpers/materialize_view_helper_spec.rb
@@ -1,0 +1,39 @@
+RSpec.describe MaterializeViewHelper do
+  describe '.reload_static_caches' do
+    context 'when in the test environment' do
+      it 'does nothing and raises no error' do
+        # The guard `return if Rails.env.test?` short-circuits the method so
+        # no constantize calls are attempted and no load_cache is fired.
+        expect { described_class.reload_static_caches }.not_to raise_error
+      end
+    end
+
+    context 'when outside the test environment' do
+      before { allow(Rails.env).to receive(:test?).and_return(false) }
+
+      it 'calls load_cache on every model listed in STATIC_CACHE_MODEL_NAMES' do
+        MaterializeViewHelper::STATIC_CACHE_MODEL_NAMES.each do |name|
+          allow(name.constantize).to receive(:load_cache)
+        end
+
+        described_class.reload_static_caches
+
+        MaterializeViewHelper::STATIC_CACHE_MODEL_NAMES.each do |name|
+          expect(name.constantize).to have_received(:load_cache).once
+        end
+      end
+    end
+  end
+
+  describe 'STATIC_CACHE_MODEL_NAMES' do
+    it 'contains 15 entries' do
+      expect(MaterializeViewHelper::STATIC_CACHE_MODEL_NAMES.size).to eq(15)
+    end
+
+    it 'every name resolves to an existing constant' do
+      expect {
+        MaterializeViewHelper::STATIC_CACHE_MODEL_NAMES.each(&:constantize)
+      }.not_to raise_error
+    end
+  end
+end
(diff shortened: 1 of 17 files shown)


spec/helpers/materialize_view_helper_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    it 'does not error when models do not have static_cache loaded' do
+      # In the test environment plugin :static_cache is not applied, so
+      # none of the models define load_cache.  reload_static_caches must
+      # handle this gracefully via respond_to? rather than blowing up.
+      expect { described_class.reload_static_caches }.not_to raise_error
+    end
+
+    it 'calls load_cache on models that respond to it' do
+      model_name = 'FakeStaticCacheModel'
+      model_class = double('model_class', load_cache: nil) # rubocop:disable RSpec/VerifiedDoubles
```
amberstarlight: Minor nitpick, but should this be `instance_double`? Then we wouldn't have to use the RuboCop disable here

neilmiddleton: Done — switched to anonymous classes with `Class.new` rather than doubles entirely. `respond_to?(:load_cache)` works naturally on the real object so the `respond_to?` stub is gone too.

amberstarlight (approved):

willfish:
I'm sure this is fine but I'm wondering if we should extend our documentation to highlight that we now have a new caching mechanism?

Currently we cache:

- Fragments via etags in redis implicitly in the frontend requests to the backend
- Explicit fragments via redis in the frontend
- View fragments via redis in the frontend
- CDN responses for GET/HEAD requests with some specific exceptions
- API gateway responses via api. with some specific exceptions
- Backend fragments via redis
- These in memory fragments

All of this might be quite confusing to a new developer

I'm also not convinced adding this additional layer will benefit us much but probably its worth testing/proving before I make this claim

neilmiddleton:
We do have a bunch of caching documented (for instance, [here](https://transformuk.atlassian.net/wiki/spaces/HO/pages/22719463434/Backend+Technical+Reference+User+Guide#9.-Caching) and [here](https://transformuk.atlassian.net/wiki/spaces/HO/pages/22681518101/Caching) so it can probably go there somewhere.

I would say unless we see a benefit, we revert this to maintain simplicity, but I think we might.

PR #3124, opened 2026-05-07
Revert: Remove init containers from worker-uk and worker-xi

HWallenberg:
## What:
This reverts the changes from PR #3070 (\"BAU: Remove init containers\"), which removed the init containers from the `worker-uk` and `worker-xi` ECS task definitions.

## Why:
After the merge of #3070 landed on `main`, the staging deploy started failing. Both `worker-uk` and `worker-xi` got stuck in a `PENDING` state for the full 15-minute ECS stabilisation window and never became healthy — causing the Terraform apply to time out and the deploy to fail.

The init containers that were removed were responsible for running database migrations (`rails db:migrate` and `rails data:migrate`) before the worker process started. Removing them triggered ECS to roll out a new task definition revision for both services at the same time, and the new tasks couldn't start successfully.

Reverting this gets staging (and production) back to a known-good state. Once we've confirmed deploys are stable again, we can look at re-landing the init container removal more carefully — making sure any health check or startup dependencies are handled cleanly first.

## Ticket:
N/A

## Risk:
**Risk level:** 🟠

**Reason for rating:**
This touches ECS task definitions for both worker services and will trigger a rolling redeploy on staging and production. It's a revert to a previously working state, so the risk is low in practice, but it's worth a quick team check before merging given it affects live workers.

### DIFF
--- a/terraform/worker_xi.tf
+++ b/terraform/worker_xi.tf
@@ -28,6 +28,9 @@ module "worker_xi" {
   container_entrypoint = [""]
   container_command    = local.worker_command
 
+  init_container_entrypoint = [""]
+  init_container_command    = local.init_command
+
   service_environment_config = local.worker_xi_secret_env_vars
 
   has_autoscaler = local.has_autoscaler
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
         args: ["--severity=warning"]
 
   - repo: https://github.com/trufflesecurity/trufflehog
-    rev: v3.95.2
+    rev: v3.93.6
     hooks:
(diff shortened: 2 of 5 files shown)


terraform/worker_xi.tf
```diff
@@ -28,6 +28,9 @@ module "worker_xi" {
   container_entrypoint = [""]
   container_command    = local.worker_command
 
+  init_container_entrypoint = [""]
```
willfish, line 31: Possibly what we're missing is a module update?

neilmiddleton (approved):

willfish (approved):

amberstarlight (approved):

PR #3127, opened 2026-05-08
Stage CDS import writes to UNLOGGED tables, promote atomically on success

neilmiddleton:
## What:

Introduces `CdsImporter::StagingManager`, which replaces the single large Sequel transaction wrapping the entire CDS file import with a two-phase write approach:

1. **Stage** — each batch is inserted into per-table `UNLOGGED` PostgreSQL staging tables. UNLOGGED tables bypass the WAL entirely.
2. **Promote** — on success, one short transaction copies all staged rows into the real oplog tables via `INSERT … SELECT`. The data is already on disk so this transaction is fast regardless of file size.
3. **Cleanup** — an `ensure` block drops all staging tables whether the import succeeded or failed, leaving the real oplog tables untouched on any error.

The outer `Sequel::Model.db.transaction` in `BaseUpdateImporter` is removed. Atomicity for CDS is now owned by `CdsUpdate#import!`. TARIC is unaffected.

## Why:

The previous approach wrapped the entire file import in one Sequel transaction. A daily CDS extract touches tens of thousands of rows across 50+ oplog tables. PostgreSQL must hold all dirty pages in `shared_buffers` (or spill to WAL temp files) until `COMMIT` — this was the cause of the nightly memory spike and the DB slowdown seen during sync.

UNLOGGED tables generate no WAL, so there is no large transaction and no buffer pressure during the multi-minute parse-and-insert phase.

**Crash safety:** PostgreSQL truncates UNLOGGED tables on crash recovery, so a mid-import crash leaves the real oplog tables clean. The sync re-runs from the source file on S3, which is idempotent.

**Atomicity:** preserved at the application layer. If parsing fails before `promote!` is called, cleanup drops the staging tables and nothing is written to the real oplog. If `promote!` fails, the promoting transaction rolls back and the staging tables are cleaned up.

## Ticket:

N/A — performance / stability fix

## Risk:

🟠 Medium — this changes the fundamental write path for CDS imports. The staging/promote pattern is well-understood but it is a significant behaviour change. Key mitigations:
- The `StagingManager` has unit tests covering the happy path, empty promotion, and cleanup idempotency
- Crash recovery is safe by construction (UNLOGGED tables are truncated by PostgreSQL on recovery)
- TARIC sync is completely unaffected
- The `ensure staging_manager&.cleanup` guard means staging tables are never orphaned even on unexpected exceptions

### DIFF
--- a/app/lib/cds_importer/staging_manager.rb
+++ b/app/lib/cds_importer/staging_manager.rb
@@ -25,6 +25,13 @@ class CdsImporter
   # UNLOGGED tables are lost (PostgreSQL truncates them on recovery), so no
   # partial data lands in the real oplog.  The sync simply re-runs from the
   # source file on S3.
+  #
+  # Orphaned tables: if the Ruby process is killed with SIGKILL (or crashes
+  # without running ensure blocks), cleanup is never called and the staging
+  # tables remain.  This is safe — table names include a per-run SecureRandom
+  # hex ID so they never collide with a future import run.  Future imports are
+  # unaffected.  Orphans can be identified and removed with:
+  #   SELECT relname FROM pg_class WHERE relname LIKE 'stg_%' AND relkind = 'r';
   class StagingManager
     def initialize
       # Short unique ID scoped to this import run.  Used to build staging
--- a/app/lib/cds_importer/staging_manager.rb
+++ b/app/lib/cds_importer/staging_manager.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+class CdsImporter
+  # Coordinates writing CDS import data to UNLOGGED staging tables, then
+  # promoting the staged rows into the real oplog tables in a single atomic
+  # transaction.
+  #
+  # Why UNLOGGED?  PostgreSQL UNLOGGED tables bypass the write-ahead log
+  # (WAL).  A full CDS daily extract can touch tens of thousands of rows
+  # across 50+ oplog tables.  Writing all of that inside one WAL-logged
+  # transaction means PostgreSQL must hold every dirty page in shared_buffers
+  # (or spill to temp files) until COMMIT — the source of the nightly memory
+  # spike.  Writing to UNLOGGED tables generates no WAL at all, so there is
+  # no large transaction and no buffer pressure.
+  #
+  # Atomicity is preserved at the application layer:
+  #   1. All rows are written to staging (no transaction needed — UNLOGGED).
+  #   2. On success, promote! copies everything into the real oplog tables
+  #      inside one short Sequel transaction.  That transaction is tiny in
+  #      wall-clock terms because the data is already on disk.
+  #   3. On any error, cleanup drops the staging tables before they are ever
+  #      promoted, leaving the real oplog tables untouched.
+  #
+  # Crash safety: if the database crashes during the staging phase, the
+  # UNLOGGED tables are lost (PostgreSQL truncates them on recovery), so no
+  # partial data lands in the real oplog.  The sync simply re-runs from the
+  # source file on S3.
+  class StagingManager
+    def initialize
+      # Short unique ID scoped to this import run.  Used to build staging
+      # table names that are guaranteed not to collide with other runs.
+      @import_id = SecureRandom.hex(4) # 8 hex chars
+      @counter   = 0
+      # Maps real oplog table name (Symbol) → staging table name (Symbol)
+      @staged_tables = {}
+    end
+
+    # Returns a Sequel dataset pointing at the UNLOGGED staging table for the
+    # given oplog operation class.  The staging table is created on first access.
+    def dataset_for(operation_klass)
+      real_table = operation_klass.table_name
+
+      staging_table = @staged_tables[real_table] ||= begin
+        name = next_staging_name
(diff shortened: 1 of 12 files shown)


app/lib/tariff_synchronizer/base_update_importer.rb
```diff
@@ -15,18 +15,12 @@ def apply
       track_latest_sql_queries
       keep_record_of_presence_errors
 
-      # IMPORTANT: running large update files may cause out of memory exception.
-      # Run `import!` outside of this class to prevent that.
-      Sequel::Model.db.transaction(reraise: true) do
```
willfish, line 20: Isn't this shared by both the XI and UK importers? I'm guessing this basically removes atomicity from the taric files and that's accidental?

app/lib/cds_importer/staging_manager.rb
```diff
@@ -0,0 +1,105 @@
...
+      end
+
+      Sequel::Model.db[staging_table]
+    end
+
+    # Copies all staged rows into the real oplog tables inside one atomic
+    # transaction.  The individual INSERT … SELECT statements are fast because
+    # the data is already on disk in the UNLOGGED tables.
+    def promote!
+      Sequel::Model.db.transaction do
```
willfish, line 63: I guess this is faster but still creates memory and WAL pressure since we're committing the whole lot at the end again?

app/lib/cds_importer/staging_manager.rb
```diff
@@ -0,0 +1,105 @@
...
+        end
+      end
+    end
+
+    # Drops all staging tables.  Safe to call even if some tables were never
+    # created (uses IF EXISTS) and idempotent across multiple calls.
+    def cleanup
+      @staged_tables.each_value do |staging_table|
+        Sequel::Model.db.run("DROP TABLE IF EXISTS #{staging_table}")
+      rescue StandardError => e
```
willfish, line 82: I'm wondering if there are situations where we might end up orphaning some of these old tables if the process shuts down in an uncontrolled way?

We might find ourselves blocked essentially until we come and manually clear these orphaned tables out

app/lib/tariff_synchronizer/base_update_importer.rb
neilmiddleton: Good catch, this is a real regression. BaseUpdateImporter used to provide the transaction wrapper for both CDS and TARIC. Removing it left TARIC with no atomicity guarantee. Fixed in the latest commit: moved the transaction into `TaricUpdate#import!` so each update type owns its own atomicity contract. CDS uses staging-table promotion; TARIC uses a single wrapping transaction (same semantics as before, just co-located with the importer).

app/lib/cds_importer/staging_manager.rb
neilmiddleton: Good question, the total volume of WAL written is the same (you always have to WAL-log the inserts into the real tables), but the memory pressure is fundamentally different. The old single transaction ran for 20-30 minutes while parsing and inserting every row, forcing PostgreSQL to hold all dirty pages in shared_buffers (or spill to temp files) until COMMIT. The `promote!` transaction by contrast takes seconds: the data is already materialised on disk in the UNLOGGED tables, so PostgreSQL's `INSERT ... SELECT` executor can pipeline — read a batch from UNLOGGED, write to WAL-logged, flush, repeat — with a small working set rather than holding the entire import in memory. Short transaction = no opportunity for buffer pressure to accumulate.

app/lib/cds_importer/staging_manager.rb
neilmiddleton: Valid concern, a SIGKILL leaves the staging tables behind. Two things mitigate the risk though:

1. **No blocking**: table names include a per-run `SecureRandom.hex(4)` ID (e.g. `stg_a1b2c3d4_01`), so a new import always creates a fresh set of tables under a different prefix. Orphans are completely invisible to future `StagingManager` instances and don't interfere with any subsequent run.

2. **PostgreSQL crash = auto-cleanup**: UNLOGGED tables are truncated by PostgreSQL on crash recovery, so the scenario most likely to kill the process without running `ensure` also auto-cleans the tables.

3. The only orphaning case that sticks is SIGKILL with the DB staying up. Those tables accumulate as dead weight but cause no functional problems. I've added a pg_class query to the class comment so ops can find and drop them if needed:

4. Happy to add a rake task or a startup cleanup if you'd prefer a more automated approach.

willfish: Or can we use something that's temporary/dies at session close https://www.postgresql.org/docs/current/sql-createtable.html?

app/lib/cds_importer/staging_manager.rb
willfish: Oh wow that's a massive difference 🐑 :it:

willfish (dismissed):
Excellent!

willfish (approved):

willfish:
I've validated this by doing a full rollback of the UK/XI tariffs in development after a fresh deploy of this branch. I think we're safe to merge

neilmiddleton:
Thanks for validating willfish — merging now.

PR #3130, opened 2026-05-11
Route GET requests to Aurora reader instance

neilmiddleton:
## What:
Routes GET and HEAD requests to the Aurora reader instance instead of sending all traffic to the writer.

- New `UseReaderForReads` Rack middleware wraps GET/HEAD in `Sequel::Model.db.with_server(:read_only)` — Sequel's `server_block` extension (already loaded) transparently routes those queries to the reader connection pool
- Adds `servers.read_only` to `config/database.yml` production stanza, pointing at a new `READER_DATABASE_URL` env var (falls back to `DATABASE_URL` so dev/test are unaffected)
- Backend Terraform: fetches `aurora-postgres-ro-connection-string` from Secrets Manager and injects it as `READER_DATABASE_URL` into the UK and XI web task definitions; updates IAM policy to allow access
- Sidekiq worker task definitions are intentionally excluded — they use the writer only, as they read data immediately after writing it

**Depends on https://github.com/trade-tariff/trade-tariff-platform-aws-terraform/pull/591 being applied first (creates the reader secret in Secrets Manager).**

## Why:
All traffic currently hits the Aurora writer even though the cluster has a dedicated reader instance. This is inefficient and increases load on the writer unnecessarily. The app is overwhelmingly read-heavy — GET requests are the vast majority of traffic and do no Sequel writes.

## Ticket:
N/A

## Risk:
**Risk level:** 🟠

**Reason for rating:**
Changes query routing for all GET requests. If the reader falls behind or is unhealthy, Aurora removes it from the reader endpoint automatically, but there is a brief window during failover where reads may get connection errors. The `connection_validator` extension (already configured with a 60s timeout) handles stale connections. The `READER_DATABASE_URL` fallback to `DATABASE_URL` in database.yml means no production impact until the env var is set.

### DIFF
--- a/app/middleware/use_reader_for_reads.rb
+++ b/app/middleware/use_reader_for_reads.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# Routes GET and HEAD requests to the Aurora reader instance.  All other HTTP
+# methods (POST, PUT, PATCH, DELETE) use the default (writer) connection pool.
+#
+# Sequel's server_block extension (already loaded in application.rb) sets a
+# thread-local so every Sequel::Model.db[...] call within the block is
+# transparently routed to the :read_only server defined in database.yml.
+#
+# Sidekiq workers do not go through Rack middleware and therefore continue
+# using the writer, which is correct — they often read data immediately after
+# writing it.
+class UseReaderForReads
+  READ_METHODS = %w[GET HEAD].freeze
+
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if READ_METHODS.include?(env['REQUEST_METHOD'])
+      Sequel::Model.db.with_server(:read_only) { @app.call(env) }
+    else
+      @app.call(env)
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


app/middleware/use_reader_for_reads.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+# Sequel's server_block extension (already loaded in application.rb) sets a
+# thread-local so every Sequel::Model.db[...] call within the block is
+# transparently routed to the :read_only server defined in database.yml.
+#
+# Sidekiq workers do not go through Rack middleware and therefore continue
+# using the writer, which is correct — they often read data immediately after
+# writing it.
+class UseReaderForReads
+  READ_METHODS = %w[GET HEAD].freeze
```
willfish, line 14: This is a nice approach

willfish (dismissed):
Very tidy

willfish:
I'm assuming we have the secrets populated and in place, etc

neilmiddleton:
> I'm assuming we have the secrets populated and in place, etc

see the other PR

willfish (dismissed):

willfish (approved):

AigulDj (approved):

PR #16, opened 2021-01-25
Add newrelic gem for monitoring

tavy88:
### Jira link

HOTT-<TODO>

### What?

Add newrelic gem for APM monitoring.

### Why?

Ability to debug easier and track performance spikes.

PR #3131, opened 2026-05-11
AI-696: Validate tariff update filenames

willfish:
## What:

- [x] Add an explicit filename whitelist for CDS and TARIC tariff update downloads
- [x] Reject path-like or unexpected filenames before deriving a storage path
- [x] Add regression coverage for traversal-style filenames
- [x] Update downloader specs to use real CDS/TARIC filename shapes

## Why:

GitHub CodeQL alert #16 flags external tariff update data flowing into file path operations.

The downloader now accepts only the known CDS/TARIC filename formats before calling `FileService`, preventing unexpected local or S3 paths from being read or written. Unexpected filenames fail closed and are recorded as failed updates for operator review.

Security alert: https://github.com/trade-tariff/trade-tariff-backend/security/code-scanning/16

## Ticket:

[AI-696](https://transformuk.atlassian.net/browse/AI-696)

## Risk:

**Risk level:** 🔴

**Reason for rating:**
This touches the tariff update synchronisation path from upstream CDS/TARIC feeds. The code footprint is small, but an upstream filename format change would now fail closed instead of downloading the file, so this should get explicit approval from Thor or Neil before merging.

### DIFF
--- a/app/lib/tariff_synchronizer/tariff_downloader.rb
+++ b/app/lib/tariff_synchronizer/tariff_downloader.rb
@@ -2,8 +2,13 @@ module TariffSynchronizer
   # Download pending updates for TARIC and CDS data
   class TariffDownloader
     TariffDownloaderZipError = Class.new(StandardError)
+    InvalidFilenameError = Class.new(StandardError)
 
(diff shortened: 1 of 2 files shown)


neilmiddleton:
might need a description update as per the template

PR #3145, opened 2026-05-12
BAU: Remove secret-pulled value for read only connection

willfish:
## What:

- Remove borked secret reference being passed in as configuration

## Why:

- The secrets are different (lower cased) from the actual configured value in each environment. We can unpick that separately and set manually for now

## Risk:
**Risk level:** 🟠 

**Reason for rating:**

This requires a manual change to a core part of the system across all environments but won't actually break anything since we fallback to the writer connection


spencerldixon (approved):
LGTM 👍

rasikasri (approved):

PR #3146, opened 2026-05-12
[HMRC-2127] Improve DifferenceReportCheck logic readability

spencerldixon:
## What:

Small QOL to the readability of the differences report logic

## Why:

Looking into some errors around this worker and this line annoyed me. Since the reports are generated on a monday, and this job runs on tuesday, we can send an alert if the last report date was before the start of the week, meaning a report hasn't been generated this week.

Instead of "is the last log less than 7 days ago at the time of running the job"
We do "did a report get logged this week?"

## Ticket:
[HMRC-2127](https://transformuk.atlassian.net/browse/HMRC-2127)

## Risk:
**Risk level:** 🟢 

**Reason for rating:**
Should do the same thing but nicer to read and understand, no awkward 7 day window, just "did it happen this week?"

### DIFF
--- a/app/workers/differences_report_check_worker.rb
+++ b/app/workers/differences_report_check_worker.rb
@@ -11,7 +11,8 @@ class DifferencesReportCheckWorker
 
     return notify if last_log.blank?
 
-    notify if last_log < 7.days.ago
+    # Notify if the report hasn't run this week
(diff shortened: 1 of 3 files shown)


amberstarlight (approved):
Much nicer!

willfish (approved):

PR #3151, opened 2026-05-14
AI-558: Update generic intercept template copy

willfish:
### Jira link

[AI-558](https://transformuk.atlassian.net/browse/AI-558)

### What?

- [x] Update the generic description intercept template wording to match the interstitial content.
- [x] Add seed coverage for the updated generic message copy.

### Why?

The generic intercept template is the seed used when a search term is too broad to suggest a tariff code. The seeded copy had one line that did not match the approved page content: "The information might be found on:" instead of "This information might be found on:".

### Risk

**Risk level:** Low

**Reason for rating:**
Seed copy only. Existing configured admin values are not migrated by this change, and runtime behaviour is unchanged apart from the default template text used when the seed/default is applied.

### DIFF
--- a/app/models/admin_configuration.rb
+++ b/app/models/admin_configuration.rb
@@ -42,7 +42,7 @@ class AdminConfiguration < Sequel::Model(Sequel[:admin_configurations].qualify(:
 
     ## Where can I find this information?
 
-    The information might be found on:
+    This information might be found on:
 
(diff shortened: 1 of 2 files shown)


spencerldixon (approved):
👍 👍 👍 👍 👍 👍 👍 👍

PR #3152, opened 2026-05-14
Require jwt

vertism:
## What:
Add JWT to gemfile

## Why:
So it will be autoloaded on boot

## Ticket:
N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:**
It is an existing dependency

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,7 @@ gem 'slack-notifier'
 
 # API related
 gem 'jsonapi-serializer'
+gem 'jwt'
 gem 'rabl'
 gem 'responders'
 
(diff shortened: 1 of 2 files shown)


willfish:
Won't we just use zeitwerk to load it as we need it? How comes it needs eager loading?

willfish (approved):

vertism:
> Won't we just use zeitwerk to load it as we need it? How comes it needs eager loading?

Previously it had been loading but I'm not exactly sure how. It wasn't defined in the gemfile so it wasn't being loaded via  `Bundler.require` at boot, and wasn't required in any app code so wasn't being autoloaded by zeitwerk.

PR #3160, opened 2026-05-15
BAU: Replace Struct with Data.define in formatters, search services and token verifier

willfish:
### What?

- [x] Replace `Struct.new(..., keyword_init: true)` with `Data.define(...)` across 9 files
- [x] Update one call site in `SearchService#vector_short_list` for the new field
- [x] Zero test or documentation changes (pure refactor)

### Why?

`Struct` is mutable and carries legacy `keyword_init` noise. `Data` (Ruby 3.2+) gives us immutable value objects with `==`, `hash`, and `#with` for free. These nine Results and Contexts are pure data carriers — exactly what `Data` was designed for. Aligns them with the existing `Data.define` usage elsewhere (`TemplateImporter::Result`, `GoodsNomenclatureResult`, etc.).

This change was extracted from the large controller-specs-to-request-specs migration (#3158) so that PR can stay strictly test-only.

### Have you?

- [x] Rebased on latest `main` from the dedicated worktree
- [x] Verified relevant formatter + search service specs
- [x] No V2 API or migration impact

### Risk

**Risk level:** 🟢 Green

**Reason for rating:** Pure refactor from `Struct` to `Data.define` with no behaviour change. All affected code paths have existing test coverage. Matches the green criterion "Refactors with full test coverage and no behaviour change" exactly. No API, data, or infrastructure impact.

### DIFF
--- a/app/formatters/duty_expression_formatter/context.rb
+++ b/app/formatters/duty_expression_formatter/context.rb
@@ -1,4 +1,4 @@
-DutyExpressionFormatter::Context = Struct.new(
+DutyExpressionFormatter::Context = Data.define(
   :duty_expression_id,
   :duty_expression_description,
   :duty_expression_abbreviation,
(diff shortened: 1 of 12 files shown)


spencerldixon (dismissed):
LGTM 👍

rasikasri (approved):

PR #3162, opened 2026-05-15
BAU: Add pre-commit to flake devShell

willfish:
## What

- Add `pkgs.pre-commit` to the devShell `buildInputs`.
- Run `pre-commit install --install-hooks` automatically in the shellHook so the git hooks are wired up and actually execute on commit/push.

## Why

Having the `pre-commit` binary available is not enough. The hooks defined in `.pre-commit-config.yaml` (terraform, shellcheck, trailing whitespace, etc.) only run if `pre-commit install` has been executed for the repo.

This was the missing piece on main. Now when you `direnv allow` (or enter a worktree), the hooks are automatically installed.

## Verification

- `nix develop` now both provides the binary and runs `pre-commit install --install-hooks`.
- Hooks are active without manual steps.

## Risk

🟢 Very low — only affects local development environment setup. The command is run with `|| true` so it never breaks entry into the shell.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -255,6 +255,9 @@ OPENSEARCH_CLI
               WT_ID=$(${worktree.id})
               export GEM_HOME="$HOME/.local/share/gem/worktrees/$WT_ID"
               export BUNDLE_PATH=".bundle"
+              export BUNDLE_APP_CONFIG=".bundle"
+              export BUNDLE_IGNORE_CONFIG=1
+              export BUNDLE_FORCE_RUBY_PLATFORM=1
               mkdir -p "$GEM_HOME" ".bundle"
               echo "Worktree Bundler isolation enabled (ID: $WT_ID)"
             else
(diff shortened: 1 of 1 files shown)


amberstarlight (dismissed):

willfish:
Updated with the hardened worktree setup pattern from the cross-repo verification pass.

Key findings applied:
- setup steps now fail the shell hook instead of being masked by `|| true`
- setup output is compact by default and writes full logs to `/tmp/worktree-<id>-*.log`
- Postgres startup uses an explicit external pid file and `worktree-clean` stops the daemon before removing state
- Nix Ruby is placed before system Ruby in `PATH`
- backend uses its actual Sequel-backed DB tasks (`db:create` + `db:structure:load` for dev/test), because `db:prepare` is not available in this app

Verification run from a fresh long-path worktree:
- first-entry `direnv exec` completed bundle install, dev/test DB setup, and pre-commit hook install
- re-entry confirmed short `PGHOST`, pid file, marker file, Nix Ruby/Bundler, and pre-commit
- `bundle exec rspec`: 8130 examples, 0 failures

amberstarlight (approved):

PR #3173, opened 2026-05-19
Add tariff change query indexes

anupamc:
## What:
Add targeted PostgreSQL indexes to tariff_changes to support high-traffic query patterns used by MyOTT change endpoints and subscription email selection.

The migration adds indexes for:

- `operation_date` + `goods_nomenclature_sid`
- commodity changes filtered by `type = 'Commodity'`
- goods nomenclature description changes filtered by `type = 'GoodsNomenclatureDescription'`
- measure changes filtered by `type = 'Measure'`

These support queries that filter tariff changes by date, commodity SID, type, and sometimes action.

## Why:
Current queries against `tariff_changes` can require broader scans because the table only had indexes on `goods_nomenclature_item_id`, `metadata`, and a partial metadata GIN index.

The affected paths include:

- `GroupedMeasureChangesService`
- `CommodityChangesService`
- `MyCommoditiesSubscriptionWorker`
- MyOTT commodity change and measure grouping endpoints

Local `EXPLAIN` `ANALYZE` comparison was run against a development database with `131,052 tariff_changes` rows. The “before” result was simulated by dropping the new indexes inside a rollback-only transaction.

Query path | Before | After | Result
-- | -- | -- | --
Grouped measure changes | 20.091ms | 0.082ms | Uses `tariff_changes_measure_date_sid_index`
Commodity ending changes | 11.530ms | 0.026ms | Uses `tariff_changes_commodity_date_sid_action_index`
Commodity description updates | 7.250ms | 0.014ms | Uses `tariff_changes_commodity_description_date_sid_action_index`
Operation date lookup | 10.167ms | 0.017ms | Uses `tariff_changes_operation_date_goods_nomenclature_sid_index`
My commodities email selection | 20.371ms | 17.390ms | Small improvement; still dominated by joins outside `tariff_changes`

The biggest improvements are on the direct date/SID/type lookups, where Postgres can now use narrow btree/partial indexes instead of scanning more of `tariff_changes`.

## Ticket:
N/A

## Risk:
**Risk level:** 🟢 (Low to medium)

**Reason for rating:**

- Additional indexes increase write overhead when inserting/updating tariff_changes.
- Index creation may take time on larger environments, depending on table size.
- Query planner may not always choose the new indexes if statistics are stale or data distribution differs from local/dev data.

**Mitigations:**

- Indexes are scoped to existing query filters.
- Partial indexes limit size for type-specific lookups.
- Migration has a down path to remove the indexes if needed.
- Verified with focused service/worker specs and full backend test suite previously.


db/structure.sql
```diff
@@ -311,6 +311,15 @@ ALTER TABLE public.live_issues ALTER COLUMN id ADD GENERATED BY DEFAULT AS IDENT
 );
 
 
+--
+-- Name: schema_migrations; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.schema_migrations (
```
willfish: This doesn't feel right. Shouldn't this already exist?

db/structure.sql
```diff
@@ -8831,6 +8841,14 @@ ALTER TABLE ONLY public.live_issues
     ADD CONSTRAINT live_issues_pkey PRIMARY KEY (id);
 
 
+--
+-- Name: schema_migrations schema_migrations_pkey; Type: CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.schema_migrations
```
willfish: Yeah we shouldn't be creating this in public. We do migrations in uk or xi but never public schema

db/structure.sql
```diff
@@ -9167,14 +9185,6 @@ ALTER TABLE ONLY uk.description_intercepts
     ADD CONSTRAINT description_intercepts_pkey PRIMARY KEY (id);
 
 
---
--- Name: description_intercepts description_intercepts_term_not_alias; Type: CHECK CONSTRAINT; Schema: uk; Owner: -
---
-
-ALTER TABLE ONLY uk.description_intercepts
```
willfish, line 9174: This should remain not sure why it's being removed

neilmiddleton:
The `description_intercepts_term_not_alias` check constraint on `uk.description_intercepts` is being removed.  it enforces that a term cannot appear in its own aliases array, which is an important data integrity constraint.

Several other unrelated changes have been picked up: the `customs_tariff_updates` default status changing from `awaiting_approval` to `pending`, constraint name changes on `customs_tariff_chapter_notes` and `customs_tariff_section_notes`, index renames on `measure_components` and `description_intercepts`, and the `chapter_note_experimentals` index being removed from the index list. These are all drift between local schema and main, not changes this PR should be making.

db/structure.sql
anupamc: Noted - removed the public.schema_migrations table creation and its primary key constraint

db/structure.sql
anupamc: The PR now only adds:
- `tariff_changes_commodity_date_sid_action_index`
- `tariff_changes_commodity_description_date_sid_action_index`
- `tariff_changes_measure_date_sid_index`
- `tariff_changes_operation_date_goods_nomenclature_sid_index`
- the migration row for `20260519120000_add_tariff_changes_query_indexes.rb`

anupamc: Observed Performance Improvements:

Query | Before | After | Improvement | Index used
-- | -- | -- | -- | --
grouped_measure_changes | 7.248ms | 0.092ms | 7.156ms | tariff_changes_measure_date_sid_index
commodity_ending_changes | 6.828ms | 0.045ms | 6.783ms | tariff_changes_commodity_date_sid_action_index
commodity_description_updates | 6.443ms | 0.006ms | 6.437ms | tariff_changes_commodity_description_date_sid_action_index
operation_date_lookup | 5.821ms | 0.022ms | 5.799ms | tariff_changes_operation_date_goods_nomenclature_sid_index
my_commodities_email_selection | 2.585ms | 3.032ms | no improvement | tariff_changes_operation_date_goods_nomenclature_sid_index

willfish (approved):

PR #3185, opened 2026-05-21
AI-334: Mark exact matches as high confidence

willfish:
### Jira link

[AI-334](https://transformuk.atlassian.net/browse/AI-334)

### What?

- [x] Mark internal guided-search exact matches as `high` confidence
- [x] Update exact-match request specs for search-reference and padded-code matches

### Why?

Exact matches currently return `confidence: null`, which makes a deterministic exact match look like an unknown-confidence answer to the frontend. Returning `high` makes the backend confidence signal match the certainty of the exact-match path.

### Risk

**Risk level:** 🟠

**Reason for rating:** This is a narrow response-shape change for exact matches only, but it is still consumed by the frontend. Non-exact retrieval results keep their existing confidence behaviour, and request specs cover both exact-match paths changed here.


HWallenberg (approved):
lgtm

PR #3186, opened 2026-05-21
Fix for New Relic deployment marker failing

anupamc:
## What:
Add New Relic deployment markers for all backend runtime entities: backend UK, backend XI, worker UK, and worker XI. 

## Why:
When backend is deployed a marker should be added in New Relic as part of the deployment process. 

This job has started failing with the error - `Valid entityGuids are required to create a deployment marker`.

## Ticket:
[HMRC-1289](https://transformuk.atlassian.net/browse/HMRC-1289)

## Risk:
🟢 

**Reason for rating:**
The change is limited to GitHub Actions deployment workflows. The main risk is missing or incorrect GitHub secrets for one of the four New Relic GUIDs, which would fail the marker step after deployment. Application runtime behaviour is unaffected.

### DIFF
--- a/.github/workflows/deploy-to-development.yml
+++ b/.github/workflows/deploy-to-development.yml
@@ -10,6 +10,10 @@ on:
       - labeled
       - unlabeled
 
+concurrency:
+  group: deploy-development
+  cancel-in-progress: true
+
 permissions:
   contents: read
   id-token: write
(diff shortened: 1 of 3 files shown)


willfish:
I'm guessing we only want markers for production

anupamc:
As per discussion with @AigulDj the workflows were deleted as per the confirmation from @neilmiddleton - this bug is not valid anymore. Closing the PR

PR #3205, opened 2026-06-01
OTTIMP-629 Use Data objects

vertism:
## What:
Replaces usage of OpenStruct with [Data](https://docs.ruby-lang.org/en/3.2/Data.html)
- ostruct dependency has been removed, and instances of OpenStruct converted to use Data.define
- 2 instances of using OpenStruct to define a class in specs have been converted to use Class.new
- Admin::GoodsNomenclatureLabels::StatsService.call now returns a data object which can be passed straight to the serializer

## Why:
Data objects are immutable and more performant that OpenStructs. 

## Ticket:
https://transformuk.atlassian.net/browse/OTTIMP-629

## Risk:
**Risk level:** 🟢

**Reason for rating:**
Like for like change with improved performance

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,6 @@ gem 'logstash-event'
 gem 'newrelic_rpm'
 gem 'nokogiri'
 gem 'notifications-ruby-client'
-gem 'ostruct'
 gem 'slack-notifier'
 
 # API related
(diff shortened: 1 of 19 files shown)


Gemfile
```diff
@@ -50,7 +50,6 @@ gem 'logstash-event'
 gem 'newrelic_rpm'
 gem 'nokogiri'
 gem 'notifications-ruby-client'
-gem 'ostruct'
```
willfish, line 53: 🔥

willfish (approved):

PR #3208, opened 2026-06-01
Cache the latest updates endpoint

neilmiddleton:
## What:
Cache the `/uk/api/updates/latest` endpoint response in `Rails.cache`, keyed by service.

## Why:
The Aurora read replica has no statistics for `uk.tariff_updates` (replicas can't run `ANALYZE`), so the query planner occasionally produces a bad plan causing ~800ms response times. The data changes at most once per day after a sync, making it a good candidate for caching. No explicit invalidation wiring is needed — `ClearCacheWorker` already calls `Rails.cache.clear` after every sync.

## Ticket:
N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:**
Purely additive caching with a safe invalidation path already in place. No behaviour change for callers — the response is identical, just served from cache after the first request.

### DIFF
--- a/spec/requests/api/v2/updates_controller_spec.rb
+++ b/spec/requests/api/v2/updates_controller_spec.rb
@@ -36,5 +36,19 @@ RSpec.describe Api::V2::UpdatesController do
         ).to eq []
       end
     end
+
+    context 'when the result is cached' do
+      it 'only queries the database once across multiple requests' do
+        create :taric_update, :applied
+
+        allow(Rails).to receive(:cache).and_return(ActiveSupport::Cache::MemoryStore.new)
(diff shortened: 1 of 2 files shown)


spec/requests/api/v2/updates_controller_spec.rb
```diff
@@ -36,5 +36,19 @@
         ).to eq []
       end
     end
+
+    context 'when the result is cached' do
+      it 'only queries the database once across multiple requests' do
+        create :taric_update, :applied
+
+        allow(Rails).to receive(:cache).and_return(ActiveSupport::Cache::MemoryStore.new)
```
willfish, line 44: 👍

willfish (dismissed):

willfish (approved):

PR #3212, opened 2026-06-02
OTTIMP-634 Handle measure end date being removed

vertism:
## What:
Currently changes to measures where the end date is being removed are not handled well. The data_of_effect value will now be a string instead of a date to show that the end data has been removed.

## Why:
We will retain the same action for end date changing but the new text will make the details of the change clearer to users

## Ticket:
https://transformuk.atlassian.net/browse/OTTIMP-634

## Risk:
**Risk level:** 🟢

**Reason for rating:**
This change requires a front end change to handle the change of format of some date fields.
https://github.com/trade-tariff/trade-tariff-frontend/pull/3061
It is a fairly narrow change and has been tested in dev.

### DIFF
--- a/spec/models/tariff_change_spec.rb
+++ b/spec/models/tariff_change_spec.rb
@@ -469,7 +469,7 @@ RSpec.describe TariffChange do
   end
 
   describe '#description' do
-    let(:tariff_change) { described_class.new(type: 'Commodity', action: action) }
+    let(:tariff_change) { described_class.new(type: 'Commodity', action: action, validity_end_date: Date.parse('2024-08-15')) }
 
     context 'when action is CREATION' do
       let(:action) { TariffChangesService::BaseChanges::CREATION }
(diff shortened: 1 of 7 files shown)


spec/models/tariff_change_spec.rb
```diff
@@ -485,6 +485,22 @@
...
+            type: 'Commodity',
+            action: action,
+            operation_date: Date.parse('2024-08-14'),
+            date_of_effect: Date.parse('2024-08-15'),
+            validity_end_date: nil,
+          )
+        end
+
+        it 'returns end description' do
+          expect(tariff_change.description).to eq('Commodity will end')
```
willfish, line 501: Isn't removing an end date effectively saying:
"Commodity has no end"?

willfish (approved):

montychristo (approved):

PR #3217, opened 2026-06-03
AI-863: Match guided search suggestions case-insensitively

willfish:
## What changed

Downcases the guided-search exact suggestion query before matching stored suggestions, and adds service coverage for a capitalized query matching a lowercase stored suggestion.

## Why

AI-863 reports different guided search behaviour for `Disposable camera` and `disposable camera`. Stored text suggestions are already normalized lower-case, but the exact lookup used the raw query, so capitalized queries could miss an existing exact suggestion and fall through into the refine-question flow.

## Risk

Low. This keeps `SearchSuggestion.by_value` as an exact lookup and only normalizes guided-search input before calling it, matching how stored text suggestions are populated.

### DIFF
--- a/app/services/api/internal/search_service.rb
+++ b/app/services/api/internal/search_service.rb
@@ -234,7 +234,7 @@ module Api
       def find_by_suggestion(query)
         ::SearchSuggestion
           .declarable
-          .by_value(singular_and_plural(query))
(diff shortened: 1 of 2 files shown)


HWallenberg (approved):
🚀

PR #3232, opened 2026-06-04
AI-864: Add tariff knowledge graph ADR

willfish:
## What:

Adds a proposed ADR for a tariff knowledge graph in Postgres.

## Why:

- nested set is necessary but insufficient for classification knowledge
- we need a clear justification/traceability for this decision

## Ticket:

https://transformuk.atlassian.net/browse/AI-864

## Risk:

**Risk level:** 🟢 Green

**Reason for rating:**
Documentation-only proposed ADR. It does not change runtime behaviour, database schema, API responses, sync logic, infrastructure, or trader-facing regulatory content; the architectural approach is being socialised before implementation work starts.

### DIFF
--- a/docs/adr/2026-06-04_tariff-knowledge-graph.md
+++ b/docs/adr/2026-06-04_tariff-knowledge-graph.md
@@ -1,0 +1,184 @@
+# Tariff knowledge graph
+
+Date: 4 June 2026
+Status: Proposed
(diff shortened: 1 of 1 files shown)


neilmiddleton (approved):
Approved from a PR point of view, but still proposed

PR #3245, opened 2026-06-08
BAU: Log search type for selected results

willfish:
## What

Adds `search_type: classic` to the `result_selected` search instrumentation path for public tariff entity selections, and includes it in the structured search log output.

## Why

Search diagnostics in admin showed `-` in the Type column for selected-result rows. The admin UI was reading the top-level `search_type` field correctly, but backend logs were not emitting that field for `result_selected` events.

## Impact

New selected-result logs will include the classic search type, so diagnostics can show a complete Type value for clicked results. Existing historical logs will remain unchanged.

## Verification

- `PGHOST=/home/william/Repositories/hmrc/trade-tariff-backend/.nix/postgres DB_USER= bundle exec rspec spec/requests/concerns/search_result_tracking_spec.rb spec/lib/search/instrumentation_spec.rb spec/lib/search/logger_spec.rb`
- `PGHOST=/home/william/Repositories/hmrc/trade-tariff-backend/.nix/postgres DB_USER= bundle exec rubocop app/controllers/concerns/search_result_tracking.rb app/lib/search/instrumentation.rb app/lib/search/logger.rb spec/requests/concerns/search_result_tracking_spec.rb spec/lib/search/instrumentation_spec.rb spec/lib/search/logger_spec.rb`

### DIFF
--- a/app/controllers/concerns/search_result_tracking.rb
+++ b/app/controllers/concerns/search_result_tracking.rb
@@ -11,6 +11,7 @@ module SearchResultTracking
 
     ::Search::Instrumentation.result_selected(
       request_id: rid,
+      search_type: 'classic',
       goods_nomenclature_item_id: params[:id],
       goods_nomenclature_class: controller_name.classify,
     )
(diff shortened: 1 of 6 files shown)


willfish:
Closing this draft: selected-result search type should be derived in admin from the original search context rather than emitted as a fixed backend value.

PR #3250, opened 2026-06-09
OTTIMP-640 Improve commodity end date removal tariff change information

vertism:
## What:
If a commodity end date is removed then we should tell users with CCWL subscriptions this information more accurately.
This matches the change in https://github.com/trade-tariff/trade-tariff-backend/pull/3212 that was made for measures.

The `measure_end_date_removed?` method has been changed to `end_date_removed?` and now also applies to commodities.

This change will show up in the JSON API, and when a spreadsheet with tariff changes is requested.

## Why:
Currently when an end date is removed, this is not made obvious to the user.

## Ticket:
https://transformuk.atlassian.net/browse/OTTIMP-640

## Risk:
**Risk level:** 🟢

**Reason for rating:**
Small API change

### DIFF
--- a/app/models/tariff_change.rb
+++ b/app/models/tariff_change.rb
@@ -119,9 +119,8 @@ class TariffChange < Sequel::Model
     measure_metadata['quota_order_number']
   end
 
-  def measure_end_date_removed?
-    type == 'Measure' &&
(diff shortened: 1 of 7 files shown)


HWallenberg (approved):
lgtm 🙌

PR #3276, opened 2026-06-11
HMRC-1866: Remove caxlsx dependency

willfish:
## What:
- Removes the unused `caxlsx` dependency.
- Migrates commodity watch list XLSX generation from Axlsx/caxlsx to FastExcel.
- Uses the FastExcel fork branch for rich string support needed by active commodity reports.
- Updates request/service specs to validate generated XLSX content via ZIP/XML inspection.

## Why:
This unblocks HMRC-1859 by completing the remaining migration path to FastExcel while preserving the existing downloadable spreadsheet behaviour for tariff changes and active commodity reports.

## Ticket:
https://transformuk.atlassian.net/browse/HMRC-1866

## Risk:
**Risk level:** 🟠

**Reason for rating:**
This changes XLSX generation for live user downloads. The public API shape is unchanged, and the affected report/controller paths have focused request and service coverage.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ gem 'sidekiq-scheduler'
 gem 'opensearch-ruby'
 
 # Reporting
-gem 'fast_excel', git: 'git@github.com:willfish/fast_excel.git', branch: 'HMRC-1866-write-rich-string'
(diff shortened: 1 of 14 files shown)


vertism (approved):
Lovely stuff

PR #3283, opened 2026-06-12
Add Link header pointing to API docs llms.txt on all responses

neilmiddleton:
## Summary

- Adds an `after_action` to `ApplicationController` that sets a `Link: <https://api-docs.trade-tariff.service.gov.uk/llms.txt>; rel="describedby"` header on every API response
- Covers all controllers (public API, admin, internal, user) since they all inherit from `ApplicationController`
- Adds a request spec asserting the header is present

## Why

LLM consumers hitting the API have no way to discover the API documentation without already knowing the docs site exists. The `Link` header with `rel="describedby"` (RFC 8288) is a machine-readable pointer that LLM tooling checks at the point of first contact — surfacing the `llms.txt` index immediately.

## Risk

low-risk — additive response header, no behaviour change, no data change, read-only observability.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,6 +13,7 @@ class ApplicationController < ActionController::API
   before_action :maintenance_mode_if_active
   around_action :configure_time_machine
   after_action  :check_query_count, if: -> { TradeTariffBackend.check_query_count? }
+  after_action  :set_api_docs_link_header
 
   def nothing
     head :ok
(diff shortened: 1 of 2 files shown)


neilmiddleton:
Note: the angle brackets around the URL in the header value are required by RFC 8288 — the URI-reference in a `Link` header must be enclosed in `<>`. Without them the header is malformed and conformant parsers will reject it.

rasikasri (approved):

PR #3296, opened 2026-06-17
HMRC-2390: Add JSON:API sparse fields and includes to V2 APIs

willfish:
### Jira link

[HMRC-2390](https://transformuk.atlassian.net/browse/HMRC-2390)

### What?

- [x] Add JSON:API `include` query option support across V2 serializers
- [x] Add sparse `fields[...]` query option support across V2 serializers
- [x] Vary cache keys for shaped responses so variants stay isolated
- [x] Guard expensive commodity, heading, subheading, chapter, and section paths with eager-loading checks
- [x] Keep the new query parameters out of public V2 Swagger documentation for now
- [x] Add coverage for request paths, serializers, Swagger behaviour, cache-key variation, and SQL notifications

### Why?

The MCP app needs smaller tariff responses shaped around the data it is actually using. Full V2 responses make each interaction heavier than necessary when MCP only needs a narrow set of related data or fields.

The rollout is deliberately MCP-first. That gives us one controlled consumer for the behaviour before we use the same response-shaping capability in frontend, admin, and other apps later.

### Risk

**Risk level:** 🟠

**Reason for rating:** New public V2 API behaviour consumed by other services. The default no-query-param response path is covered by the existing V2 request suite, and the new response-shaping paths have focused request, serializer, swagger, cache-key, and SQL-notification coverage.

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -19,6 +19,11 @@ module Api
       render json: serializer.serialized_errors({ error: exception.message, url: request.url }), status: :bad_request
     end
 
+    rescue_from JSONAPI::Serializer::UnsupportedIncludeError do |exception|
+      serializer = TradeTariffBackend.error_serializer(request)
+      render json: serializer.serialized_errors({ error: exception.message, url: request.url }), status: :bad_request
+    end
+
     protected
 
     def current_page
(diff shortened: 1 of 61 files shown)


neilmiddleton (approved):

neilmiddleton:
IMO this is low risk

PR #3305, opened 2026-06-18
[HMRC-2296] Task to check for database backup freshness

spencerldixon:
## What:

Takes `ENVIRONMENT` and `S3_BUCKET` and checks the latest db dumps `LastModified` timestamp, compares it to `now` with a freshness threshold (default 24 hours, but adjustable). Also checks the archive copy and expects an archive to exist matching that latest updated dump.

We then publish it as a cloudwatch metric with the `--publish` flag. 1 for healthy, 0 for unhealthy.

Unhealthy cases are:
- Latest dump is missing
- Latest dump exists but is older than 24 hours
- Daily archive is missing
- Daily archive exists but is older than the latest dump

Also handles the case where BST time kicks in and our backup naming conventions shift by 1 day. 

Run with:

```
ENVIRONMENT=production S3_BUCKET=trade-tariff-database-backups-1234 bin/check-database-backup-freshness --publish
```

Also produces a human readable output:

```
Database backup freshness check
Status: FAIL
Environment: development
Bucket: trade-tariff-database-backups-1234
Backup job: backend-database-backup-development

Last backup
  Key: tariff-merged-development.sql.gz
  Last modified: 2026-06-16T23:01:34+00:00
  Age: 1d 15h 13m 58s (141238s)
  Threshold: 1d 0h 0m 0s (86400s)

Archive backup
  Key: 2026/06/17/tariff-merged-development.sql.gz
  Last modified: 2026-06-16T23:26:16+00:00
  Archive date timezone: Europe/London

CloudWatch metric: not published; value would be 0
Elapsed: 1s (1s)

Reasons:
  - latest object is 141238s old, over threshold 86400s
```

## Why:
Operators rely on the production database dump for local restores and downstream environment refreshes. A recent outage meant the scheduled backup path stopped producing an updated dump, but the team only found out after the dump was needed.

This PR lays the ground work for an automated task to check this, publish as a cloudwatch metric, and trigger a slack notification.

## Ticket:
[HMRC-2296](https://transformuk.atlassian.net/browse/HMRC-2296)

## Risk:
 🟢

**Reason for rating:**
Doesn't change anything existing, just adds another task to run

### DIFF
--- a/bin/check-database-backup-freshness
+++ b/bin/check-database-backup-freshness
@@ -199,17 +199,21 @@ epoch_for_timestamp() {
 
 archive_key_for_epoch() {
   local epoch=$1
+  local offset_seconds=${2:-0}
+  local archive_epoch
+
+  archive_epoch=$((epoch + offset_seconds))
 
   # bin/backup-database uses plain date +"%Y/%m/%d" in a Europe/London
-  # container. EventBridge runs cron(0 23 * * ? *) in UTC, so during BST the
-  # archive prefix is the next London calendar day even though S3 LastModified
-  # is still on the previous UTC date.
-  if TZ="$BACKUP_ARCHIVE_TIME_ZONE" date -d "@$epoch" +"%Y/%m/%d" >/dev/null 2>&1; then
-    echo "$(TZ="$BACKUP_ARCHIVE_TIME_ZONE" date -d "@$epoch" +"%Y/%m/%d")/tariff-merged-${ENVIRONMENT}.sql.gz"
+  # container. The latest object's LastModified is upload completion time, so a
+  # backup that starts before midnight and finishes after midnight may archive
+  # under the previous date.
+  if TZ="$BACKUP_ARCHIVE_TIME_ZONE" date -d "@$archive_epoch" +"%Y/%m/%d" >/dev/null 2>&1; then
+    echo "$(TZ="$BACKUP_ARCHIVE_TIME_ZONE" date -d "@$archive_epoch" +"%Y/%m/%d")/tariff-merged-${ENVIRONMENT}.sql.gz"
     return
   fi
 
-  echo "$(TZ="$BACKUP_ARCHIVE_TIME_ZONE" date -r "$epoch" +"%Y/%m/%d")/tariff-merged-${ENVIRONMENT}.sql.gz"
+  echo "$(TZ="$BACKUP_ARCHIVE_TIME_ZONE" date -r "$archive_epoch" +"%Y/%m/%d")/tariff-merged-${ENVIRONMENT}.sql.gz"
 }
 
 reasons=()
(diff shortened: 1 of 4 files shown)


shjohnson (dismissed):
I think looks good; I did notice backup starts at 23:00 UTC and this check runs at 01:00, so there’s only a ~2 hour window. If pg_dump + upload takes longer than that, we’ll still be looking at yesterday’s tariff-merged-*.sql.gz and could get a false alert. Worth scheduling the check a bit later? or building in some slack for slow backups.

bin/check-database-backup-freshness
```diff
@@ -0,0 +1,267 @@
...
+if latest_last_modified=$(head_object_last_modified "$LATEST_KEY"); then
+  latest_epoch=$(epoch_for_timestamp "$latest_last_modified")
+  latest_age_seconds=$((now_epoch - latest_epoch))
+
+  if [ "$latest_age_seconds" -gt "$BACKUP_FRESHNESS_THRESHOLD_SECONDS" ]; then
+    freshness_ok=0
+    reasons+=("latest object is ${latest_age_seconds}s old, over threshold ${BACKUP_FRESHNESS_THRESHOLD_SECONDS}s")
+  fi
+
+  archive_key=$(archive_key_for_epoch "$latest_epoch")
```
shjohnson, line 232: Backup sets the archive path at start; which we derive from upload time, could this mismatch across midnight and false-fail on a missing archive.

terraform/backend_job.tf
```diff
@@ -68,6 +68,74 @@ resource "aws_cloudwatch_event_target" "database_backup" {
...
+}
+
+resource "aws_cloudwatch_event_target" "database_backup_freshness_check" {
+  count = var.enable_alarms ? 1 : 0
+
+  rule     = aws_cloudwatch_event_rule.database_backup_freshness_check[0].name
+  arn      = data.aws_ecs_cluster.this.arn
+  role_arn = aws_iam_role.eventbridge_ecs.arn
+
+  input = jsonencode({
```
willfish, line 87: Do we need these overrides? I think they're part of the backend-job-configuration secret already

spencerldixon: The containerOverrides block is needed because we default to `bin/null-service` on line 3 via `local.job_command`.

We probably don't need `ENVIRONMENT` as we set this, but the rest we don't set as far as i can tell

bin/check-database-backup-freshness
spencerldixon: Good spot. I've added a fallback to check the previous date which should work for crossing the midnight threshold for archive name and last modified dates. I've also changed this to run at 6am which should give the backup enough time to complete

terraform/backend_job.tf
willfish: I believe we set the region

bin/check-database-backup-freshness
```diff
@@ -0,0 +1,289 @@
...
+usage() {
+  echo "Usage: $0 [--publish]"
+  echo
+  echo "Checks the database backup objects in S3."
+  echo
+  echo "Options:"
+  echo "  --publish  Publish the freshness result as a CloudWatch metric."
+}
+
+for arg in "$@"; do
```
morganism, line 26: No change requested .. just a compliment
respect for not using getopts

morganism (commented):
Just a comment. no change

terraform/backend_job.tf
spencerldixon: Sweet, I've removed those, can we test this workflow on dev/staging somehow?

willfish (approved):
Looks good to me. Thanks for addressing the region overrides.

PR #3308, opened 2026-06-19
OTTIMP-665 Add tariff change where declarable has children added

vertism:
## What:
If a declarable commodity has children added to it, it will no longer be declarable. There is no oplog event on the new parent, and so this change is not currently created as a TariffChange object.
Now when a new commodity is added, if its parent was previously declarable, a TariffChange record is created with the date of effect set to the last day the parent was declarable.

## Why:
Adds events for missing tariff change data.

## Ticket:
https://transformuk.atlassian.net/browse/OTTIMP-665

## Risk:
**Risk level:** 🟢

**Reason for rating:**
Enhances data for CCWL without changing existing functionality.

### DIFF
--- a/app/services/tariff_changes_service.rb
+++ b/app/services/tariff_changes_service.rb
@@ -147,6 +147,8 @@ class TariffChangesService
     # Multiple child creations can map to same parent on same run; keep earliest child date_of_effect.
     parent_date_of_effect_by_sid = created_children.each_with_object({}) do |child, dates_by_parent_sid|
       parent = child.parent
+      next if parent.nil?
+
       child_change = created_child_changes_by_sid[child.goods_nomenclature_sid]
       next if child_change.nil?
 
--- a/app/services/tariff_changes_service.rb
+++ b/app/services/tariff_changes_service.rb
@@ -87,6 +87,8 @@ class TariffChangesService
       add_change_record(change, change[:goods_nomenclature_item_id], change[:object_sid])
     end
 
+    add_parent_declarability_loss_records
+
     @changes[:commodity_descriptions].each do |change|
       next if matching_commodity_change?(change[:goods_nomenclature_sid], change[:action])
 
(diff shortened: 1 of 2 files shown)


app/services/tariff_changes_service.rb
```diff
@@ -124,6 +126,68 @@ def measure_declarables_by_sid
...
+      add_change_record(change, change[:goods_nomenclature_item_id], change[:goods_nomenclature_sid])
+    end
+  end
+
+  def parent_declarability_loss_changes
+    created_child_changes = @changes[:commodities]
+      .select { |change| change[:action] == BaseChanges::CREATION }
+    created_child_changes_by_sid = created_child_changes.index_by { |change| change[:object_sid] }
+    created_child_sids = created_child_changes_by_sid.keys.compact.uniq
+
```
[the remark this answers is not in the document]
vertism: This is outside the scope of this ticket

app/services/tariff_changes_service.rb
```diff
@@ -124,6 +126,70 @@ def measure_declarables_by_sid
...
+      dates_by_parent_sid[parent.goods_nomenclature_sid] =
+        if existing_date.nil? || (candidate_date && candidate_date < existing_date)
+          candidate_date
+        else
+          existing_date
+        end
+    end
+
+    return [] if parents.empty?
```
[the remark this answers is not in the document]
vertism: This handles changes that happen on the applied date.

app/services/tariff_changes_service.rb
```diff
@@ -124,6 +126,70 @@ def measure_declarables_by_sid
...
+
+  def parent_declarability_loss_changes
+    created_child_changes = @changes[:commodities]
+      .select { |change| change[:action] == BaseChanges::CREATION }
+    created_child_changes_by_sid = created_child_changes.index_by { |change| change[:object_sid] }
+    created_child_sids = created_child_changes_by_sid.keys.compact.uniq
+
+    return [] if created_child_sids.empty?
+
+    created_children = GoodsNomenclature.where(goods_nomenclature_sid: created_child_sids).eager(:parent).all
```
willfish, line 145: This lookup runs inside the outer `TimeMachine.at(date)` from `all_changes`, where `date` is the operation date. For created commodities, though, `date_of_effect` can come from a later `validity_start_date` (`BaseChanges#date_of_effect`). In that case the child is present in `@changes`, but its tree-node parent is not visible in the operation-date snapshot, so `filter_map(&:parent)` drops it and no parent ENDING change is emitted.

I reproduced this locally with a temporary spec: parent declarable on 2025-01-14, child created with `operation_date: 2025-01-15` and `validity_start_date: 2025-01-17`. At `TimeMachine.at(2025-01-15)`, `child.parent` was `nil`; at `TimeMachine.at(2025-01-17)`, the parent was present; `parent_declarability_loss_changes` returned `[]`. The parent lookup needs to be resolved at the child change's effective date, not the operation date.

app/services/tariff_changes_service.rb
```diff
@@ -124,6 +126,70 @@ def measure_declarables_by_sid
...
+  def add_parent_declarability_loss_records
+    parent_declarability_loss_changes.each do |change|
+      next if matching_commodity_change?(change[:goods_nomenclature_sid], change[:action])
+
+      add_change_record(change, change[:goods_nomenclature_item_id], change[:goods_nomenclature_sid])
+    end
+  end
+
+  def parent_declarability_loss_changes
+    created_child_changes = @changes[:commodities]
```
willfish, line 138: Driving this from `@changes[:commodities]` means the new logic only sees creations that `CommodityChanges.collect` kept, and that collector filters to `op_record.record&.declarable?`. A previously declarable parent can lose declarability when the immediate child that is added is itself non-declarable, for example a grouping/intermediate node with declarable grandchildren. In that shape the grouping child is excluded here, and the grandchildren's parent is the new grouping node rather than the old declarable parent, so the old parent is never considered for the synthetic ENDING record.

If this PR is intended to cover "a declarable commodity has children added to it", the source set needs to include relevant created children regardless of whether the newly-created child is itself declarable.

app/services/tariff_changes_service.rb
```diff
@@ -124,6 +126,70 @@ def measure_declarables_by_sid
...
+          candidate_date
+        else
+          existing_date
+        end
+    end
+
+    return [] if parents.empty?
+
+    parent_sids = parents.map(&:goods_nomenclature_sid)
+    previous_parents_by_sid = TimeMachine.at(date - 1.day) do
```
willfish, line 169: This previous-parent check is anchored to `operation_date - 1.day`, but the synthetic ending date is derived from each child's `date_of_effect`. When the child is imported before it becomes valid, this can ask "was the parent declarable before import?" rather than "was it declarable immediately before the child became effective?". It should be evaluated per parent using the effective end date (`child date_of_effect - 1.day`).

Also, the query inside the TimeMachine block does not call `.actual`; plain `GoodsNomenclature.where(...).all` is not validity-filtered by the TimeMachine plugin. Without `.actual`, a parent row that was not valid on that date can still be loaded and pass `declarable?`, producing a false ENDING change.

spec/services/tariff_changes_service_spec.rb
```diff
@@ -625,12 +625,88 @@
...
+          type: 'Commodity',
+          object_sid: child_sid,
+          action: TariffChangesService::BaseChanges::CREATION,
+          date_of_effect: child_date_of_effect,
+        }],
+        commodity_descriptions: [],
+        measures: [],
+      })
+
+      allow(GoodsNomenclature).to receive(:where).with(goods_nomenclature_sid: [child_sid]).and_return(children_scope)
```
willfish, line 661: These examples stub the `GoodsNomenclature.where(...).eager(:parent)` path and provide `child_current.parent` directly, so they do not exercise the real nested-set association under TimeMachine. That misses the failure mode where a child created on the operation date has a future `validity_start_date`: the real `parent` association is nil at the operation-date snapshot, but present at the child's effective date.

Please add an integration-style example around `generate_commodity_change_records` or `parent_declarability_loss_changes` using real goods nomenclatures/tree nodes with `operation_date < validity_start_date`, so the test proves the parent is resolved at the effective date.

willfish (changes requested):
Thanks for iterating on this. I found a couple of behavioural gaps that mean the synthetic parent ENDING record can still be missed, especially when operation date and validity start date differ, and when the added child is not itself declarable. I also left a test comment because the current stubs bypass the TimeMachine/nested-set path where the main issue reproduces.

Evidence checked locally:
- `direnv exec /home/william/Repositories/hmrc/trade-tariff-backend/.worktrees/review-pr-3308 bundle exec rspec spec/services/tariff_changes_service_spec.rb` -> 46 examples, 0 failures.
- Temporary local reproduction, removed before review: a child with `operation_date: 2025-01-15` and `validity_start_date: 2025-01-17` had no `parent` at the operation-date TimeMachine snapshot, did have the parent at its effective date, and `parent_declarability_loss_changes` returned no parent loss change.

trade-tariff-infrastructure-continuity (approved):

trade-tariff-infrastructure-continuity (approved):

PR #3358, opened 2026-07-08
Enrich openapi spec

anupamc:
## What:
Enriched the V2 OpenAPI contract generated from `trade-tariff-backend` rswag specs. This adds clearer operation, parameter, response-schema, OAuth, GB/XI dataset, and standard error-response documentation for the priority API endpoints, including commodity lookup, search, certificate/footnote/reference code endpoints, measure condition codes, and additional code search.

## Why:
The existing OpenAPI spec was structurally valid but too sparse for humans and AI-assisted tooling to use reliably. Better descriptions help consumers choose the right endpoint, understand required parameters, handle GB/XI dataset differences, configure OAuth client credentials, and process `400`, `404`, and `422` failures consistently.

## Ticket:
[HMRC-2320](https://transformuk.atlassian.net/browse/HMRC-2320)

## Risk:
**Risk level:** 🟢 

**Reason for rating:**
This is documentation-only: no API runtime behaviour, routes, serializers, or response payloads were changed.

### DIFF
--- a/app/controllers/api/admin/live_issues_controller.rb
+++ b/app/controllers/api/admin/live_issues_controller.rb
@@ -22,12 +22,10 @@ module Api
       def update
         live_issue = LiveIssue.with_pk!(params[:id])
 
-        begin
-          live_issue.update(live_issue_params)
-          render json: serialize(live_issue), status: :ok
-        rescue StandardError
-          render json: serialize_errors(live_issue), status: :unprocessable_content
-        end
+        live_issue.update(live_issue_params)
+        render json: serialize(live_issue), status: :ok
+      rescue Sequel::ValidationFailed
+        render json: serialize_errors(live_issue), status: :unprocessable_content
       end
 
       def destroy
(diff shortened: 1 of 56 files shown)


neilmiddleton (approved):
A few observations on the non-documentation changes that are also in this branch:

The `normalized_commodities` method in `live_issues_controller.rb` is more complex than it needs to be. The `return [nil]` paths inside the block and at the top level feel like an attempt to propagate invalid input through to the model for validation to catch, but it's unusual to return `[nil]` rather than an empty array or just letting the raw value through. The `invalid_commodity_codes?` check in `LiveIssue` will correctly flag a `nil` element as invalid, so this works, but the intent isn't immediately obvious. A comment would help.

The `CustomsTariffGeneralRule.latest_rules` method makes two queries - one to find the latest update and one to eager load the rules. The `latest` scope on `CustomsTariffUpdate` uses `.actual.exclude(status: FAILED).order(Sequel.desc(:validity_start_date))` but returns a dataset, so `.first` will hit the database. Then the eager-loaded rules are accessed via association. This is fine for the low-frequency classification prompt path, but worth confirming it doesn't get called per-request on high-traffic paths.

The bulk refresh work in `public_atar_search_refresh.rb` looks like a performance improvement over per-ruling refreshes. Worth confirming the batch size is bounded so a large import doesn't produce a single very large OpenSearch bulk request.

This PR is quite wide for a single review. For future work in this vein, it would be easier to review if the OpenAPI spec enrichment and the runtime behaviour changes were in separate PRs.

PR #3361, opened 2026-07-08
[HMRC - 1240] - Add security HTTP headers

punitcse:
## What:
Adds two response headers via default_headers:

`Content-Security-Policy`: default-src 'none'; frame-ancestors 'none'
`X-Frame-Options:` DENY (previously `SAMEORIGIN`, the Rails default)

Remove following unnecessary files (all commented so should not affect)
- config/initializers/content_security_policy.rb
- config/initializers/new_framework_defaults_7_0.rb
- config/initializers/new_framework_defaults_7_1.rb

## Why:
The API currently sends no **CSP**. While CSP has no effect on JSON consumed programmatically, it acts as a safety net if a response is ever rendered directly in a browser — e.g. an error page returning HTML, a missing/incorrect `Content-Type`, or a future endpoint reflecting user input. With default-src 'none', nothing can load or execute in that scenario, turning a potential XSS on the API origin into a non-issue. This is the policy [OWASP's REST Security Cheat Sheet ](https://cheatsheetseries.owasp.org/cheatsheets/REST_Security_Cheat_Sheet.html#security-headers) recommends for API responses.
`X-Frame-Options `is tightened from `SAMEORIGIN` to `DENY` because API responses have no legitimate framing use case, including same-origin. frame-ancestors 'none' in the CSP covers this in modern browsers; the explicit X-Frame-Options header is kept for older tooling.

## Ticket:
https://transformuk.atlassian.net/browse/HMRC-1240

## Risk:
**Risk level:** 🟠 Amber

**Reason for rating:**
The changes are very small but it touches every response so blast-radius is high. I also dont know if any API consumer affect by these changes so on safe side put it to amber.

## Not changed: X-XSS-Protection stays at 0

The pen test also recommended X-XSS-Protection: 1; mode=block. We're deliberately not making that change. The header is a no-op in all modern browsers (the XSS Auditor it controlled was removed from Chrome 78+, Chromium Edge, and Safari; Firefox never had it), and in older browsers the auditor was itself exploitable (script-neutralisation and XS-Leak attacks) — which is why Rails changed its own default from 1; mode=block to 0 in Rails 6.1, and why OWASP now recommends 0. The protection it was meant to provide is delivered properly by the CSP added in this PR. Refs: [OWASP HTTP Headers Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#x-xss-protection), [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection).


willfish (approved):
LGTM — CSP + `X-Frame-Options: DENY` via `default_headers.merge!` is the right approach, matches OWASP REST guidance for an API, and the request specs cover both headers. Leaving `X-XSS-Protection` at 0 is correct. Nice clean pen-test follow-up.

willfish (approved):

PR #3374, opened 2026-07-09
fix(rspec): replace indexed let names

willfish:
## What:
- [x] Rename numbered `let` helpers (for example `item_1`) to descriptive names across affected specs
- [x] Re-enable `RSpec/IndexedLet` in `.rubocop.yml`

## Why:
Indexed let names hide intent and were only allowed because the cop was disabled. Renaming the helpers is a behaviour-preserving spec cleanup that lets the cop stay on for new code.

## Ticket:
N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:**
Spec-only renames and a RuboCop re-enable. No production code changes.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -123,8 +123,6 @@ Style/StringLiterals:
 RSpec/MessageChain:
   Enabled: false
 
-RSpec/IndexedLet:
-  Enabled: false
 
 RSpec/MultipleExpectations:
   Enabled: false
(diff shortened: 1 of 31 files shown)


willfish:
Closing to avoid a multi-PR rebase quagmire. Process is now **one PR at a time** (simplest first): land/merge the current PR, cleanup the branch/worktree, rebase the next kind onto updated main, open only that one. This PR will be reopened later in sequence if still needed.

Kept open as the current single ready PR: #3372 (docs/direnv) — unless this is not that PR, in which case it will return after #3372 merges.

willfish:
Closed for sequential one-PR-at-a-time process. Will reopen after prior kinds merge.

PR #3424, opened 2026-07-09
AI-983: Adds request and serializer regression coverage for duty calculator metadata on declarable heading responses

anupamc:
## What:
Adds regression coverage for duty calculator metadata on V2 declarable heading responses, including serializer and request specs. The change also confirms non-declarable heading responses continue to omit this metadata, and keeps the field out of public swagger documentation.

## Why:
This protects the calculator decision contract for heading-level declarables without broadening the documented public API surface. It gives consumers a stable behaviour to rely on while keeping the externally documented contract focused on currently supported public fields.

## Ticket:
[AI-983](https://transformuk.atlassian.net/browse/AI-983)

## Risk:
**Risk level:** 🟢

**Reason for rating:**
The change is limited to regression coverage and swagger scope adjustment. It does not alter production serialization logic, commodity metadata behaviour, or non-declarable heading responses.

### DIFF
--- a/lib/tasks/exchange_rates.rake
+++ b/lib/tasks/exchange_rates.rake
@@ -1,7 +1,7 @@
 module ExchangeRatesRakeTasks
   FILE_TYPES = %w[monthly_csv monthly_xml monthly_csv_hmrc].freeze
 
-  module_function
+module_function
 
   def rebuild_old_monthly_rates
     require_monthly_env!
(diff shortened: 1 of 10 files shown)


willfish:
We don't want this documented in swagger since this is for us only. Probably at some point we'll move this over to the internal scope

willfish (approved):

PR #3456, opened 2026-07-10
HMRC-2437: Drop TARIC presence error app plumbing

willfish:
### Jira link

[HMRC-2437](https://transformuk.atlassian.net/browse/HMRC-2437)

### What?

- [x] Remove presence-error model, admin serializer association, and importer subscription
- [x] Remove applied-mailer presence-error section
- [x] Remove sync tooling presence-error output
- [x] Remove `TARIFF_IGNORE_PRESENCE_ERRORS` / `tariff_ignore_presence_errors`
- [x] Update related specs
- [ ] ~~Drop `tariff_update_presence_errors` table~~ deferred (table left in place)

### Why?

After [HMRC-2435](https://transformuk.atlassian.net/browse/HMRC-2435) TARIC destroy/update no longer depends on a current materialized projection, so presence errors are mostly expected sequencing noise rather than actionable import failures.

Value was extracted from the up-to-date XI database first for rollback/replay planning. App plumbing can go; schema drop can follow as a separate low-ceremony change once we are happy.

### Risk

**Risk level:** 🟢

**Reason for rating:** App/code cleanup only. No migration, no table drop, no TARIC import behaviour change beyond stopping writes to a diagnostic table that is no longer useful after HMRC-2435. Reversible by restoring the removed plumbing if needed.


willfish:
Leaving this pending an investigation by Punit

PR #3460, opened 2026-07-13
Fix search analytics selection query request_source grouping

willfish:
## What:

- [x] Stop CloudWatch selection stats from regrouping by an aggregate alias named `request_source`
- [x] Alias the per-request source as `source`, then rename it back to `request_source` after the final stats
- [x] Keep selection trend query substitutions aligned with the new query shape
- [x] Cover the fixed query fragments in the snapshot query specs

## Why:

Search analytics snapshot refresh was failing with `MalformedQueryException: Unknown symbols found in or after stats: request_source`. CloudWatch Logs Insights rejects a second `stats ... by request_source` when `request_source` only exists as an aggregate output from `earliest(request_source) as request_source`. That broke selection-rate queries introduced with request-source classification.

## Ticket:

N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:** Narrow CloudWatch Insights query rewrite for admin search analytics only. No public API, tariff data, or sync path changes; covered by existing unit specs.

### DIFF
--- a/app/services/search_analytics/cloudwatch_snapshot_query.rb
+++ b/app/services/search_analytics/cloudwatch_snapshot_query.rb
@@ -165,9 +165,10 @@ module SearchAnalytics
         | fields if(ispresent(request_source), request_source, "unknown") as request_source
         | stats sum(result_selection_marker) as result_selections,
             sum(selectable_search_marker) as selectable_searches,
-            earliest(request_source) as request_source by request_id
+            earliest(request_source) as source by request_id
         | filter selectable_searches > 0
-        | stats sum(result_selections) as selected, sum(selectable_searches) as selectable by request_source
+        | stats sum(result_selections) as selected, sum(selectable_searches) as selectable by source
+        | fields selected, selectable, source as request_source
       QUERY
     end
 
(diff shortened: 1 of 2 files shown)


trade-tariff-infrastructure-continuity (approved):

app/services/search_analytics/cloudwatch_snapshot_query.rb
```diff
@@ -165,21 +165,22 @@ def selection_query(selectable_condition)
         | fields if(ispresent(request_source), request_source, "unknown") as request_source
         | stats sum(result_selection_marker) as result_selections,
             sum(selectable_search_marker) as selectable_searches,
-            earliest(request_source) as request_source by request_id
+            earliest(request_source) as source by request_id
         | filter selectable_searches > 0
-        | stats sum(result_selections) as selected, sum(selectable_searches) as selectable by request_source
+        | stats sum(result_selections) as selected, sum(selectable_searches) as selectable by source
+        | fields selected, selectable, source as request_source
```
amberstarlight, line 171: `selected` as an ephemeral field has already been used, this will result in a `MalformedQueryException`; we don't need the last `fields` line here.

```suggestion
```

Results from this modified query will look like this:

| source   | selected | selectable |
| -------- | -------: | ---------: |
| frontend | 1        | 6          |

PR #3462, opened 2026-07-13
AI-895: Add XI Combined Nomenclature ETL pipeline

HWallenberg:
## What:

This PR adds a brand new ETL pipeline for the XI (Northern Ireland) tariff — the **XI Combined Nomenclature (CN) importer**. It also tidies up a few things on the UK side.

Here's what's been added or changed:

**New XI CN pipeline:**
- **`XiCnImporter::DocumentFetcher`** — queries the EU Publications Office SPARQL endpoint to discover new annual CN regulations, then fetches both the XHTML and PDF from the EU Cellar CDN
- **`XiCnImporter::NotesExtractor`** — state machine that walks the EU OJ XHTML body and extracts section notes, chapter notes, and General Rules of Interpretation (GRIs)
- **`XiCnImporter::NotesExtractor::Formatter`** — converts raw extracted lines into govspeak-compatible markdown, handling EU-specific formatting quirks (parenthetical list markers, em-dash bullets, deep table indentation)
- **`XiCnImporter::Importer`** — orchestrates fetch → extract → persist, saving the PDF to S3 and all notes to the database as `pending`
- **`XiCnImporter::Reimporter`** — re-fetches the stored Cellar HTML URL and re-extracts notes for one or all versions; used by the rake task and admin UI
- **`ImportXiCnDocumentWorker`** — Sidekiq worker (default queue) scheduled daily at 9am with Slack notifications
- **`XiCnReimportWorker`** — background worker for on-demand reimport triggered from the admin UI
- **`XiCnImporter::Instrumentation`** and **`XiCnImporter::Logger`** — structured logging and StatsD metrics throughout
- **`lib/tasks/xi_cn_importer.rake`** — rake task `importer:xi_cn:reimport[<version>]` for manual reimport

**UK side fixes:**
- **`ImportCustomsTariffDocumentWorker`** — added `return unless TradeTariffBackend.uk?` guard (mirrors the XI worker) so it's a no-op if accidentally enqueued on an XI instance
- **`Api::Admin::CustomsTariffUpdates::ReimportController`** — now dispatches to the correct reimporter based on `SERVICE` (UK → `CustomsTariffImporter::Reimporter`, XI → `XiCnImporter::Reimporter`)

**Spec fixes:**
- `sections_summary_controller_spec.rb` — unpended the chapter counts test by creating a proper `chapter :with_section` in the test rather than relying on real tariff data

## Why:

Northern Ireland uses EU tariff classification under the Windsor Framework, so the XI service needs its own source of section and chapter notes — the EU Combined Nomenclature — rather than the UK GOV.UK document. Without this pipeline, XI notes would need to be maintained manually, which isn't sustainable as the CN updates annually.

The EU Cellar CDN serves structured XHTML (unlike the UK's Word `.docx`), so the XI extractor is a separate implementation — a state machine walking XML nodes rather than using a MarkerTrie for indentation inference.

A couple of things worth noting for reviewers:
- The XI importer stores the Cellar HTML URL on `CustomsTariffUpdate.source_url` at import time so the reimporter can re-fetch without re-running the SPARQL query
- The UK importer already had S3 storage and a service guard on the reimport controller, but was missing a service guard on the worker itself — that's been added

## Ticket:

[AI-895](https://transformuk.atlassian.net/browse/AI-895)

## Risk:

**Risk level:** 🟢

**Reason for rating:** Entirely additive — new tables, new workers, new rake task. Nothing existing is removed or modified in a breaking way. The UK reimport controller change is covered by request specs. The new XI workers are guarded by `TradeTariffBackend.xi?` so they're a no-op on UK instances.

> ⚠️ **Follow-up ticket needed:** Existing UK customs tariff documents in the dev/staging/production S3 buckets may need migrating to the new `data/customs_tariff_documents/uk/` prefix path. This has been scoped as a separate ticket and does not block this PR.

## Have you?

### DIFF
--- a/app/lib/customs_tariff_importer/importer.rb
+++ b/app/lib/customs_tariff_importer/importer.rb
@@ -1,6 +1,6 @@
 module CustomsTariffImporter
   class Importer
-    S3_KEY_PREFIX = 'data/customs_tariff_documents'.freeze
+    S3_KEY_PREFIX = 'data/customs_tariff_documents/uk'.freeze
 
     Result = Data.define(:status, :version, :error) do
       def initialize(status:, version: nil, error: nil)
--- a/app/lib/xi_cn_importer/document_fetcher.rb
+++ b/app/lib/xi_cn_importer/document_fetcher.rb
@@ -1,0 +1,123 @@
+require 'digest'
+require 'json'
+require 'net/http'
+
+module XiCnImporter
+  class DocumentFetcher
+    SPARQL_ENDPOINT      = 'https://publications.europa.eu/webapi/rdf/sparql'.freeze
+    CELLAR_HTML_TEMPLATE = 'http://publications.europa.eu/resource/cellar/%s.0006.03/DOC_1'.freeze
+    CELLAR_PDF_TEMPLATE  = 'http://publications.europa.eu/resource/cellar/%s.0006.01/DOC_1'.freeze
+    MAX_REDIRECTS = 5
+
+    SPARQL_QUERY = <<~SPARQL.freeze
+      PREFIX cdm: <http://publications.europa.eu/ontology/cdm#>
+      PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+
+      SELECT DISTINCT ?work ?celex ?force_date ?pub_date
+      WHERE {
+        ?work cdm:resource_legal_amends_resource_legal
+              <http://publications.europa.eu/resource/cellar/c5a368fa-fd18-4efb-ae18-e4ac4e83055c> .
+        ?work cdm:resource_legal_id_celex ?celex .
+        ?work cdm:resource_legal_date_entry-into-force ?force_date .
+        OPTIONAL { ?work cdm:resource_legal_date_document ?pub_date . }
+        FILTER(MONTH(?force_date) = 1 && DAY(?force_date) = 1)
+        FILTER(?force_date >= "2024-01-01"^^xsd:date)
+      }
+      ORDER BY DESC(?force_date)
+    SPARQL
+
+    Result = Data.define(
+      :celex,
+      :force_date,
+      :publication_date,
+      :cellar_url,
+      :html_content,
+      :pdf_content,
+      :pdf_checksum,
+    )
+
+    def call
+      bindings = sparql_results
+
+      bindings.filter_map do |binding|
+        celex      = binding.dig('celex', 'value')
+        work_uri   = binding.dig('work', 'value')
+        force_date = Date.parse(binding.dig('force_date', 'value'))
+        pub_date   = binding.dig('pub_date', 'value')&.then { |d| Date.parse(d) }
+
+        next if CustomsTariffUpdate
+                  .exclude(status: CustomsTariffUpdate::FAILED)
+                  .where(version: celex)
+                  .any?
+
+        guid        = work_uri.split('/').last
+        cellar_url  = sprintf(CELLAR_HTML_TEMPLATE, guid)
+        pdf_url     = sprintf(CELLAR_PDF_TEMPLATE, guid)
+
+        start_time   = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        html_content = fetch_html(cellar_url)
+        pdf_content  = fetch_binary(pdf_url)
+        duration_ms  = ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - start_time) * 1000).round(2)
+
+        Instrumentation.document_fetched(celex:, duration_ms:)
+
+        Result.new(
+          celex:,
+          force_date:,
+          publication_date: pub_date,
+          cellar_url:,
+          html_content:,
+          pdf_content:,
+          pdf_checksum: Digest::SHA256.hexdigest(pdf_content),
+        )
+      end
+    rescue StandardError => e
+      Instrumentation.fetch_failed(
+        url: SPARQL_ENDPOINT,
+        error_class: e.class.name,
+        error_message: e.message,
+      )
+      raise
+    end
+
+    private
+
+    def sparql_results
+      uri      = URI(SPARQL_ENDPOINT)
+      response = Net::HTTP.post_form(uri, 'query' => SPARQL_QUERY, 'format' => 'application/sparql-results+json')
+      raise "SPARQL request failed: HTTP #{response.code}" unless response.is_a?(Net::HTTPSuccess)
+
+      JSON.parse(response.body).dig('results', 'bindings') || []
+    end
+
+    def fetch_html(url, redirect_count: 0)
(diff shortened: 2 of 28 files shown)


app/lib/xi_cn_importer/reimporter.rb
```diff
@@ -0,0 +1,80 @@
...
+    private
+
+    def reimport(update)
+      html_content = fetch_html(update.source_url)
+      extracted    = NotesExtractor.new(update.version, html_content).call
+
+      CustomsTariffUpdate.db.transaction do
+        CustomsTariffSectionNote.where(customs_tariff_update_version: update.version).delete
+        CustomsTariffChapterNote.where(customs_tariff_update_version: update.version).delete
+        CustomsTariffGeneralRule.where(customs_tariff_update_version: update.version).delete
```
willfish: Empty extract wipes all notes. If Cellar gives us something we can't parse (error page, structure change), we still delete section/chapter/GRI notes and write nothing. Bail out before the deletes when extract comes back empty.

app/lib/xi_cn_importer/document_fetcher.rb
```diff
@@ -0,0 +1,123 @@
...
+      raise "Too many redirects fetching #{url}" if redirect_count > MAX_REDIRECTS
+
+      uri      = URI(url)
+      response = Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == 'https') do |http|
+        http.get(uri.request_uri, 'Accept' => 'application/xhtml+xml')
+      end
+
+      case response
+      when Net::HTTPSuccess     then response.body
+      when Net::HTTPRedirection then fetch_html(response['location'], redirect_count: redirect_count + 1)
```
willfish: HTTP client is a bit thin. Cellar URLs are plain `http://`, redirects use `Location` as-is (relative ones break), and there are no open/read timeouts. Prefer HTTPS, resolve redirects with `URI.join`, and set timeouts so the daily job can't hang. Same pattern in the reimporter.

app/workers/import_xi_cn_document_worker.rb
```diff
@@ -0,0 +1,63 @@
...
+    return unless TradeTariffBackend.xi?
+
+    start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    XiCnImporter::Instrumentation.import_run_started
+
+    results = XiCnImporter::Importer.new.call
+
+    duration_ms    = ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - start_time) * 1000).round(2)
+    imported_count = results.count { |r| r.status == :imported }
+    skipped_count  = results.count { |r| r.status == :skipped }
```
willfish: `skipped` will always be 0. Skips happen in the fetcher (`next`), so the importer never returns `:skipped`. Either return/instrument skips like UK, or drop this count so Slack isn't lying.

app/lib/xi_cn_importer/importer.rb
```diff
@@ -0,0 +1,101 @@
...
+    def call
+      documents = DocumentFetcher.new.call
+      documents.map { |doc| import_document(doc) }
+    end
+
+    private
+
+    def import_document(fetched)
+      s3_path = "#{S3_KEY_PREFIX}/CN_#{fetched.celex}.pdf"
+      TariffSynchronizer::FileService.write_file(s3_path, fetched.pdf_content)
```
willfish, line 23: Only the PDF goes to S3. UK reimports from the stored file; here reimport re-fetches live Cellar HTML. If that URL dies or changes, recovery is gone. Store the HTML too (or reimport from something we own) so reimport is reproducible.

app/lib/xi_cn_importer/logger.rb
```diff
@@ -0,0 +1,10 @@
+module XiCnImporter
+  module Logger
+    module_function
+
+    def log(level, message, **context)
+      formatted = context.map { |k, v| "#{k}=#{v}" }.join(' ')
+      Rails.logger.public_send(level, "[XiCnImporter] #{message} #{formatted}".strip)
+    end
```
willfish, line 74: Instrumentation has no listener. UK attaches a LogSubscriber so import events become structured logs. Here we fire notifications but nothing is wired up, so the daily run is mostly silent beyond Slack counts. Mirror the UK logger, or drop the dead instrumentation surface.

willfish (commented):
Extractor/formatter work looks solid. Main concerns are reimport recovery/safety and ops visibility - inline.

app/lib/customs_tariff_importer/importer.rb
```diff
@@ -1,6 +1,6 @@
 module CustomsTariffImporter
   class Importer
-    S3_KEY_PREFIX = 'data/customs_tariff_documents'.freeze
+    S3_KEY_PREFIX = 'data/customs_tariff_documents/uk'.freeze
```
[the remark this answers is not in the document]
HWallenberg: yes there is a Jira AI-1019 for this

app/lib/xi_cn_importer/document_fetcher.rb
```diff
@@ -0,0 +1,123 @@
+require 'digest'
+require 'json'
+require 'net/http'
+
+module XiCnImporter
+  class DocumentFetcher
+    SPARQL_ENDPOINT      = 'https://publications.europa.eu/webapi/rdf/sparql'.freeze
+    CELLAR_HTML_TEMPLATE = 'http://publications.europa.eu/resource/cellar/%s.0006.03/DOC_1'.freeze
+    CELLAR_PDF_TEMPLATE  = 'http://publications.europa.eu/resource/cellar/%s.0006.01/DOC_1'.freeze
+    MAX_REDIRECTS = 5
```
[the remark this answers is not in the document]
HWallenberg: Resolved in next commit.

app/lib/xi_cn_importer/document_fetcher.rb
HWallenberg: Updates to be applied:
Secure Cellar URLs
Open and Read Timeouts

Updates already applied by co-pilot:
Resolve Redirects with URI.join

app/lib/xi_cn_importer/reimporter.rb
HWallenberg: agreed fix will be in next commit.

app/workers/import_xi_cn_document_worker.rb
HWallenberg: agree fix will be in next commit

app/lib/xi_cn_importer/logger.rb
HWallenberg: yes my bad, I'll get this sorted.

app/lib/xi_cn_importer/importer.rb
HWallenberg: working on it, will try it out locally and commit when ready thanks.

willfish (approved):

PR #3465, opened 2026-07-13
fix: remove request_source field shadowing in CloudWatch Logs Insights queries

neilmiddleton:
## What:
Remove `request_source` from the initial `fields` declarations in five CloudWatch Logs Insights queries (`volume_query`, `zero_result_query`, `source_all_latency_query`, `source_view_latency_query`, `selection_query`).

## Why:
CloudWatch Logs Insights raises a `MalformedQueryException` — "Unknown symbols found in or after stats: request_source" — when a `fields` command creates an alias that shadows a field name already projected by an earlier `fields` command in the same query. The affected queries were declaring `request_source` up front, then immediately redefining it via `fields if(ispresent(request_source), request_source, "unknown") as request_source`. The `stats` clause downstream saw an ambiguous symbol and rejected the query.

Removing `request_source` from the initial `fields` projection means the `fields if(...)` line reads the raw log field and assigns the alias cleanly — no shadowing, no ambiguity.

## Ticket:
N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:** Read-only observability change — corrects malformed CloudWatch Logs Insights query strings that were causing the `SearchAnalytics::SnapshotRefresh` Sidekiq worker to fail. No API, data, or infrastructure changes.

### DIFF
--- a/app/services/search_analytics/cloudwatch_snapshot_query.rb
+++ b/app/services/search_analytics/cloudwatch_snapshot_query.rb
@@ -92,7 +92,7 @@ module SearchAnalytics
 
     def volume_query
       <<~QUERY
-        fields @timestamp, event, search_type, request_source
+        fields @timestamp, event, search_type
         | #{log_stream_filter}
         | #{base_search_filter}
         | fields if(ispresent(request_source), request_source, "unknown") as request_source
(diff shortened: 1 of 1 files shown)


willfish (approved):
LGTM

PR #3466, opened 2026-07-13
fix(rubocop): re-enable BlockLength and ModuleLength

willfish:
## What:
Re-enable `Metrics/BlockLength` and `Metrics/ModuleLength` after clearing remaining hotspots:

- Split label/self-text instrumentation into generation + scoring event modules
- Split optimized many-to-many into SQL helpers, eager loader, and ClassMethods
- Split measure dataset filters into association / scope / overview modules

Documented excludes remain for deferred work: engines route DSL, nested-set association blocks, rake namespaces, CDS synchronizer, and migrations/bin.

## Why:
Closes out the length-cop part of the RuboCop warehouse programme so these cops stay green on normal app code instead of staying permanently disabled.

## Ticket:
N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:**
Behaviour-preserving structural splits with targeted specs green (optimized M2M 14/14, measure 167/167). No query-semantics or runtime setting changes intended; Sequel plugin load path preserved via explicit requires.


willfish:
Closing to split under check-pr-lines (800). Sequential slices next: instrumentation → measure filters → M2M + enable cops.

PR #3556, opened 2026-07-22
[HMRC 2497] -  Align batch operations for TARIC and CDS importers

punitcse:
## What:
TARIC import currently saves individual record. These changes update TARIC to use batch operations same as in CDS to save records and align the two loaders. 

- [x] Update TARIC to use Batch processing.
- [x] Remove code that was used to process individual records.

**NOTE:** It does not include staging manager changes which will be in next PR.
 
## Why:
The goal is to make this two loaders consistent and make it easy to understand for new people and avoid any redundant efforts to change them separately. This ticket will focus on aligning the Database operations which is the core part of how this two strategies drifted. It will also make TARIC operations faster. 

## Ticket:
https://transformuk.atlassian.net/jira/software/c/projects/HMRC/boards/155?selectedIssue=HMRC-2497

## Risk:
**Risk level:**  🟠 

**Reason for rating:**
Risk is set to medium as it is only affecting TARIC and only deployed after testing in Dev environment thoroughly. We also have sufficient coverage in unit/integration tests so it is not a high risk.


app/lib/taric_importer/entity_mapper/measure_condition_mapper.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class TaricImporter
+  class EntityMapper
+    class MeasureConditionMapper < BaseMapper
+      self.entity_class = 'MeasureCondition'
+
+      def before_build(attributes, operation)
```
punitcse, line 8: It is done to replace the before_create callback on `MeasureCondition` model.

app/lib/taric_importer/record_inserter.rb
```diff
@@ -0,0 +1,32 @@
+require 'tariff_synchronizer/import/batch_record_inserter'
+
+class TaricImporter
+  class RecordInserter < TariffSynchronizer::Import::BatchRecordInserter
```
punitcse, line 4: The goal is to reuse the BatchRecordInserter in CDS in next PR. But it is safe to deploy one importer change at a time.

app/lib/taric_importer.rb
```diff
@@ -32,32 +38,48 @@ def initialize(taric_update)
...
+    end
+    handler = XmlProcessor.new(@taric_update.issue_date, handlers:)
     file = TariffSynchronizer::FileService.file_as_stringio(@taric_update)
     TaricImporter::XmlParser::Reader.new(file, 'record', handler).parse
+    handler.after_parse
     post_import(file_path: @taric_update.file_path, filename:)
 
     @oplog_inserts
+  ensure
+    ActiveSupport::Notifications.unsubscribe(subscription) if subscription
```
punitcse, line 58: This ensures that we are not leaking subscription elsewhere.

config/debride.whitelist
```diff
@@ -514,3 +515,5 @@ TradeTariffBackend::TariffUpdateEventListener#subscribe!
 VersionedAcceptHeader#matches?
 main#SidekiqDeathHandler
 TariffKnowledge::Node#note_blocks
+ChangeOld#operation_date=
+TariffChangesService::BaseChanges#record=
```
punitcse: These are false indicators. These methods belongs to `Record` class which is now removed.

config/debride.whitelist
```diff
@@ -514,3 +515,4 @@ TradeTariffBackend::TariffUpdateEventListener#subscribe!
 VersionedAcceptHeader#matches?
 main#SidekiqDeathHandler
 TariffKnowledge::Node#note_blocks
+ChangeOld#operation_date=
```
punitcse, line 518: Both of these are false indicators. These methods belongs to Record class which is now removed.

app/lib/taric_importer/entity_mapper/measure_condition_mapper.rb
```diff
@@ -0,0 +1,15 @@
...
+        return attributes unless operation == :create
+
+        MeasureCondition.assign_national_sid_if_missing(attributes)
```
willfish: Two SID-less conditions in the same batch get the same negative SID, so one can replace the other in the materialized view. Could we make SID allocation batch-safe and add a two-record regression test?

app/models/measure_condition.rb
```diff
@@ -72,10 +72,13 @@ class MeasureCondition < Sequel::Model
...
-  def before_create
-    self.measure_condition_sid ||= self.class.next_national_sid
-
-    super
+  # Called explicitly by TaricImporter::EntityMapper before a create-operation
+  # instance is built, since writes now go through a raw batched insert
+  # (TariffSynchronizer::Import::BatchRecordInserter) rather than Sequel's
+  # model lifecycle - there is no before_create hook to rely on anymore.
+  def self.assign_national_sid_if_missing(attributes)
+    attributes['measure_condition_sid'] = next_national_sid if attributes['measure_condition_sid'].blank?
```
willfish: Could we keep the `before_create` fallback too? Moving this rule into the TARIC mapper means other creation paths can now save a condition without a SID.

app/lib/taric_importer/record_inserter.rb
```diff
@@ -0,0 +1,32 @@
...
+    def batch_size
+      TradeTariffBackend.taric_importer_batch_size
+    end
+
+    def event_name
+      'taric_importer.import.operations'
+    end
+
+    def grouped_records(records)
+      records.chunk_while do |left, right|
```
willfish, line 16: CDS groups the whole batch by oplog table, while this only groups neighbouring records. Is there a reason TARIC can’t use the CDS approach here?

punitcse: It is done because, before batching, each transaction was applied and instrumented individually one operation, one event, one accurate count  and the split preserves that guarantee under batching, ensuring every instrumented group represents exactly one operation type so TARIC and CDS feed an `oplog_inserts` summary keyed by operation then entity_class, which is only accurate if every event's operation label matches its count. CDS groups only on `operation_klass`, so if a batch contains both _create_ and _updated_ measures they can land in the same bucket the event then reports operation `create`  with `count: group.size `covering both 8 creates + 4 updates gets reported as "_12 creates, 0 updates_" corrupting the breakdown even though the actual database writes are still correct.

app/models/measure_condition.rb
punitcse: My understanding was we always create MeasureCondition from Operations tables. But this is good point as removing this could have side effect even if it is just created from importers.

app/lib/taric_importer/entity_mapper/measure_condition_mapper.rb
punitcse: Thanks, this was a great point and I missed it completely. Changes are done in [9e3c1c8](https://github.com/trade-tariff/trade-tariff-backend/pull/3556/commits/9e3c1c881d2435a79f4d55d6b65a9f3e702bf08a)

app/lib/taric_importer.rb
```diff
@@ -32,32 +39,49 @@ def initialize(taric_update)
...
-
-    handler = XmlProcessor.new(@taric_update.issue_date)
+    subscription = subscribe_to_oplog_inserts
+    handlers = @handler_classes.map do |handler_class|
+      handler_class.new(filename, staging_manager: @staging_manager)
+    end
+    handler = XmlProcessor.new(@taric_update.issue_date, handlers:, national_sid_counter: NationalSidCounter.new)
     file = TariffSynchronizer::FileService.file_as_stringio(@taric_update)
     TaricImporter::XmlParser::Reader.new(file, 'record', handler).parse
+    handler.after_parse
```
willfish, line 53: If the final batch fails here, the error skips the rescue in `process_xml_node`. It is not wrapped or logged like failures during parsing. Can we handle `after_parse` failures the same way and include the batch IDs in the log?

app/lib/trade_tariff_backend/config/tariff_sync.rb
```diff
@@ -29,6 +29,10 @@ def cds_importer_batch_size
         ENV.fetch('CDS_IMPORT_BATCH_SIZE', '100').to_i
       end
 
+      def taric_importer_batch_size
+        ENV.fetch('TARIC_IMPORT_BATCH_SIZE', '100').to_i
```
willfish: `"bad".to_i` is `0`, so a typo in `TARIC_IMPORT_BATCH_SIZE` flushes every record and turns batching off. Can we reject values below `1` and fail fast?

app/lib/taric_importer.rb
punitcse: Done in [9ef7c0a](https://github.com/trade-tariff/trade-tariff-backend/pull/3556/commits/9ef7c0a1fe5c64c632ea59efbcc0d97f3e8a67f4)

willfish (dismissed):

willfish (dismissed):

willfish (dismissed):

willfish (approved):

PR #3586, opened 2026-07-27
AI-1082: Send Appendix 5a guidance-change emails via GOV.UK Notify

HWallenberg:
## What:

This PR moves the Appendix 5a guidance-change email (the one that tells the CUPID team about new/changed/removed CDS guidance documents) off the old Rails ActionMailer/SES setup and onto GOV.UK Notify instead.

Here's what changed:

- **`Appendix5aEmailWorker`** — a new small Sidekiq worker that sends the email through Notify. This follows the same pattern already used by `StopPressEmailWorker` and `MyCommoditiesEmailWorker`, so it should look familiar.
- **`Appendix5aPopulatorService#notify`** — instead of building and delivering an ActionMailer email directly, it now loops over the configured CUPID team addresses and enqueues one `Appendix5aEmailWorker` job per address. Nothing about *when* an email gets sent has changed — it still only fires when there's an actual change (new/changed/removed guidance), never on a no-op day.
- **`config/initializers/notify.rb`** — added a `notifications` group to hold Notify template IDs, with `appendix5a` as the first entry inside it. This mirrors the existing `myott` group shape, and is set up this way so future notification emails (see below) can live alongside it without needing another restructure.
- Removed the old **`Appendix5aMailer`** and its spec, since nothing calls it anymore.
- Removed a spec that only existed to test some manual logging we'd added around the old mailer to debug an SES delivery bug. That logging isn't needed any more — see the Why section below.

## Why:

We're moving all our transactional emails over to GOV.UK Notify, and Appendix 5a was still on the older ActionMailer/SES path. Notify gives us a proper dashboard for managing templates and a built-in delivery audit trail, so it's a better fit going forward, and it gets Appendix 5a onto the same pattern as our other notification emails.

A couple of things worth calling out:

- We dropped the custom delivery logging (the `log_mail_delivery_state` method and its spec) that was added a while back to chase down an SES bug. We don't need it any more: every Notify send is now automatically recorded in `GovukNotifierAudit` (so we can see what was sent and when), and if a send fails, Sidekiq's normal retry behaviour kicks in — if it still fails after retries, our existing Slack death-handler alert fires, same as it would for any other job. So we're not losing visibility, just removing something bespoke in favour of infrastructure we already rely on everywhere else.
- The `notifications` config group is deliberately named generically rather than being called `appendix5a` directly. This is in preparation for an upcoming Notify email for the Customs Tariff Notes work, which will live in the same Notify "Notifications" folder — this shape means we can just add another key alongside `appendix5a` when that lands, rather than restructuring the config again.
- The staging and production Notify template UUIDs are still placeholders (`PENDING_APPENDIX5A_STAGING_TEMPLATE_UUID` / `PENDING_APPENDIX5A_PRODUCTION_TEMPLATE_UUID`) — these need creating in the Notify dashboard for those environments and swapping in before this is deployed past dev.

## Ticket:

[AI-1082](https://transformuk.atlassian.net/browse/AI-1082)

## Risk:
**Risk level:** 🟢

**Reason for rating:** This is a covered refactor with no change to when or to whom emails are sent — only how the email is delivered. Full test coverage exists for the new worker and the updated service, and it follows an established pattern already used elsewhere in the codebase.


config/initializers/notify.rb
```diff
@@ -9,6 +9,9 @@
           stop_press: '3295f0bf-c75f-4202-8dcf-703e4564b932',
           tariff_change: '5db33f13-7235-4ed8-b704-e3fddc01ee09',
         },
+        notifications: {
+          appendix5a: 'PENDING_APPENDIX5A_PRODUCTION_TEMPLATE_UUID', # TODO: replace before deploying to production
```
willfish: We can't merge with placeholder template IDs. Staging and production emails will fail until these are real UUIDs.

app/services/appendix_5a_populator_service.rb
```diff
@@ -87,62 +87,18 @@ def notify
...
-      removed_guidance.count,
-    )
-
-    log_mail_delivery_state('before_deliver', message_delivery)
-    message_delivery.deliver_now
-  rescue StandardError => e
-    log_mail_delivery_state('deliver_failed', message_delivery, exception: e)
-    raise
+
+    TradeTariffBackend.cupid_team_to_emails.each do |email|
```
willfish: Can we fail if this list is empty? Atm the guidance is updated, no email is queued and the notification is permanently lost.

app/services/appendix_5a_populator_service.rb
```diff
@@ -87,62 +87,18 @@ def notify
...
+      Appendix5aEmailWorker.perform_async(
```
willfish: Can we avoid putting the email address in the Sidekiq args? We log all job args and post them to Slack when retries are exhausted.

app/services/appendix_5a_populator_service.rb
```diff
@@ -87,62 +87,18 @@ def notify
...
+        email,
+        added_guidance.count,
+        changed_guidance.count,
+        removed_guidance.count,
+      )
+    end
```
willfish, line 117: If one of these enqueues fails, earlier recipients get a job and later ones are permanently missed. This needs a retryable notification record or outbox.

app/workers/appendix5a_email_worker.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def perform(email, new_count, changed_count, removed_count)
+    personalisation = {
+      new_count:,
+      changed_count:,
+      removed_count:,
+      support_email: TradeTariffBackend.support_email,
+    }
+
+    client.send_email(email, TEMPLATE_ID, personalisation, nil, nil)
```
willfish: This only proves Notify accepted the email. Can we check the final delivery status so permanent failures are alerted?

app/services/appendix_5a_populator_service.rb
HWallenberg: Added a guard: if `cupid_team_to_emails` is empty we log an error and return rather than raising. The DB transaction already commits before `notify` runs, so raising here wouldn't protect any data — it would just turn a config gap into an unhandled exception bubbling out of the daily `RefreshAppendix5aGuidanceWorker` run. Logging makes the miss visible without failing the import.

app/services/appendix_5a_populator_service.rb
HWallenberg: Fixed — the loop now enqueues the recipient's index into `cupid_team_to_emails` instead of the address itself, and `Appendix5aEmailWorker#perform` resolves the email from config inside the job. Matches the existing convention in `StopPressEmailWorker`/`MyCommoditiesEmailWorker` (resolve from `user_id`) and `EnquiryForm::SendSubmissionEmailWorker` (resolve from a cache reference) — none of them put PII directly in Sidekiq args either.

app/services/appendix_5a_populator_service.rb
HWallenberg: added code to retry enqueue failures for missed recepients

app/workers/appendix5a_email_worker.rb
HWallenberg: code added to check status of notify email and alert if there are any failures to slack and instrumentation logging

config/initializers/notify.rb
HWallenberg: @willfish waiting for re-review and if we're happy I'll create the templates and update this file.  thanks 👍

config/initializers/notify.rb
```diff
@@ -9,6 +9,9 @@
...
+        },
```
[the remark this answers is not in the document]
HWallenberg: awaiting final review and then I'll add templates and commit to file to merge.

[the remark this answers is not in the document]
willfish: Looks good to me buddy :)

willfish (dismissed):

config/initializers/notify.rb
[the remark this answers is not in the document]
HWallenberg: Thanks, adding templates

willfish (approved):

PR #3593, opened 2026-07-29
[HMRC-2296] Refactor check database backup freshness script

spencerldixon:
## What:

Refactors timestamp conversion to use ruby instead of `date`

## Why:

Date has variations in flags and naming between OSs. What works fine on mac doesn't always work on alpine where this is being deployed. Farming this out to ruby ensures consistency in date conversion on all OSs.

## Ticket:
[HMRC-2296](https://transformuk.atlassian.net/browse/HMRC-2296)

## Risk:
🟢

**Reason for rating:**
Only changes a utility script, no changes to daily ops of service

### DIFF
--- a/bin/check-database-backup-freshness
+++ b/bin/check-database-backup-freshness
@@ -218,6 +218,7 @@ check_archive_for_epoch() {
   local candidate
   local offset
   local status
+  local archive_epoch
 
   for offset in 0 -86400; do
     candidate=$(archive_key_for_epoch "$latest_epoch" "$offset")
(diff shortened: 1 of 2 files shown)


bin/check-database-backup-freshness
```diff
@@ -216,6 +207,43 @@ archive_key_for_epoch() {
...
+  local offset
+  local status
+
+  for offset in 0 -86400; do
+    candidate=$(archive_key_for_epoch "$latest_epoch" "$offset")
+    candidate_keys+=("$candidate")
+    archive_key=$candidate
+
+    if archive_last_modified=$(head_object_last_modified "$candidate"); then
+      if [ "$(epoch_for_timestamp "$archive_last_modified")" -lt "$latest_epoch" ]; then
```
willfish: Could we capture this first and return if parsing fails? As written, a Ruby parse error gives `[` an empty value, then this function returns `0` and publishes a green metric.

bin/check-database-backup-freshness
```diff
@@ -216,6 +207,43 @@ archive_key_for_epoch() {
   echo "$(TZ="$BACKUP_ARCHIVE_TIME_ZONE" date -r "$archive_epoch" +"%Y/%m/%d")/tariff-merged-${ENVIRONMENT}.sql.gz"
 }
 
+# Looks for the archive copy of the latest backup, trying the upload completion
+# date first and then the day before, because a backup that crosses midnight
+# archives under the earlier date. Sets archive_key, archive_last_modified,
+# freshness_ok and reasons. Returns non-zero only when S3 fails for a reason
+# other than the object being absent.
+check_archive_for_epoch() {
```
willfish, line 215: Can we add a small script spec for this? Same-day archive, previous-day fallback, both missing and a parse/AWS failure should cover the useful paths.

HWallenberg (dismissed):

willfish (approved):

PR #3594, opened 2026-07-29
perf: eliminate 30-min daily DB pressure from compressed note refresh

neilmiddleton:
## What:
Eliminate the 30-minute daily DB pressure spike caused by `RefreshTariffKnowledgeCompressedNotesWorker`. Three fixes to the compressed note refresh pipeline, each with a failing test written before implementation.

**Fix 1 — skip guard on unchanged state**
`CompressedNoteRefresh#call` now computes a fingerprint of the current tariff source version and sorted declarable SID set before running any loaders. If it matches the value stored in `Rails.cache` from the previous run, the worker returns immediately. On a normal day with no tariff changes this reduces the worker from ~30 minutes to milliseconds. The fingerprint expires after 8 days so a cache eviction triggers at most one redundant full run.

**Fix 2 — RETURNING in `SourceGraphLoader#upsert_node`**
Every node insert previously did an `INSERT ... ON CONFLICT DO UPDATE` followed by a separate `Node.by_key(key).first` SELECT to retrieve the row. Changed to `INSERT ... ON CONFLICT ... RETURNING *` with `Node.call(row)`, halving DB round-trips for every source, fragment, block, and range node upserted.

**Fix 3 — batch prefetch in `CompressedNoteGenerator`**
`upsert_note` and `mark_existing_note_stale` each called `CompressedNote[sid]` — a PK lookup per commodity — inside the per-batch loop. Now a single `by_sids` query prefetches all existing notes for the batch at the start of `call`; the pre-fetched note is threaded through `generate_for`, `upsert_note`, and `mark_existing_note_stale`.

## Why:
New Relic showed `RefreshTariffKnowledgeCompressedNotesWorker` running for 30–37 minutes daily at 03:00 UTC, keeping the backend DB under sustained write load and causing the frontend to hit its 15s Faraday timeout on random endpoints (`Timeout::ExitException`, first seen Feb 2026, still firing daily). A `SearchAnalyticsSnapshotWorker` retry cascade (19–20 runs/day on Jul 22–23 after a bug fix) caused the recent spike, but the underlying baseline timeouts trace back to this worker growing slower as the tariff knowledge dataset grows.

## Ticket:
BAU

## Risk:
**Risk level:** 🟢

### DIFF
--- a/app/services/tariff_knowledge/compressed_note_generator.rb
+++ b/app/services/tariff_knowledge/compressed_note_generator.rb
@@ -10,6 +10,11 @@ module TariffKnowledge
 
     def call
       nodes = declarable_nodes
+      existing_notes = CompressedNote
+        .by_sids(nodes.map(&:goods_nomenclature_sid))
+        .all
+        .index_by(&:goods_nomenclature_sid)
+
       apply_edges = apply_edges_for_declarable_nodes(nodes)
       evidence_by_node_id = evidence_by_declarable_node_id(nodes, apply_edges)
       block_evidence_by_node_id = block_evidence_by_declarable_node_id(nodes, apply_edges)
@@ -21,6 +26,7 @@ module TariffKnowledge
           evidence_by_node_id.fetch(declarable_node.id, []),
           block_evidence_by_node_id.fetch(declarable_node.id, []),
           contained_fragment_key_lookup,
+          existing_notes[declarable_node.goods_nomenclature_sid],
         )
       end
     end
@@ -35,8 +41,8 @@ module TariffKnowledge
           .all
     end
 
-    def generate_for(declarable_node, evidence, block_evidence, contained_fragment_keys_by_block_node_id)
-      return mark_existing_note_stale(declarable_node) if evidence.empty? && block_evidence.empty?
+    def generate_for(declarable_node, evidence, block_evidence, contained_fragment_keys_by_block_node_id, existing_note)
+      return mark_existing_note_stale(existing_note) if evidence.empty? && block_evidence.empty?
 
       content = evidence.any? ? content_for(declarable_node, evidence) : block_content_for(block_evidence)
       attributes = {
@@ -56,7 +62,7 @@ module TariffKnowledge
         expired: false,
       }
 
-      upsert_note(declarable_node, attributes)
+      upsert_note(declarable_node, attributes, existing_note)
     end
 
     def apply_edges_for_declarable_nodes(declarable_nodes)
@@ -335,19 +341,18 @@ module TariffKnowledge
       end
     end
 
-    def upsert_note(declarable_node, attributes)
-      note = CompressedNote[declarable_node.goods_nomenclature_sid]
-      return mark_note_stale_if_context_changed(note, attributes[:context_hash]) if note&.manually_edited
+    def upsert_note(declarable_node, attributes, existing_note)
+      return mark_note_stale_if_context_changed(existing_note, attributes[:context_hash]) if existing_note&.manually_edited
 
-      if note
-        note.update(attributes)
+      if existing_note
+        existing_note.update(attributes)
       else
         CompressedNote.create(attributes.merge(goods_nomenclature_sid: declarable_node.goods_nomenclature_sid))
       end
     end
 
-    def mark_existing_note_stale(declarable_node)
-      CompressedNote[declarable_node.goods_nomenclature_sid]&.mark_stale!
+    def mark_existing_note_stale(existing_note)
+      existing_note&.mark_stale!
       nil
     end
 
--- a/app/services/tariff_knowledge/compressed_note_generator.rb
+++ b/app/services/tariff_knowledge/compressed_note_generator.rb
@@ -345,7 +345,7 @@ module TariffKnowledge
       return mark_note_stale_if_context_changed(existing_note, attributes[:context_hash]) if existing_note&.manually_edited
 
       if existing_note
-        existing_note.update(attributes)
+        existing_note.this.where(manually_edited: false).update(attributes)
       else
         CompressedNote.create(attributes.merge(goods_nomenclature_sid: declarable_node.goods_nomenclature_sid))
       end
--- a/app/services/tariff_knowledge/compressed_note_refresh.rb
+++ b/app/services/tariff_knowledge/compressed_note_refresh.rb
@@ -3,7 +3,7 @@ module TariffKnowledge
     BATCH_SIZE = 500
     FINGERPRINT_CACHE_KEY = 'tariff_knowledge:compressed_note_refresh:fingerprint'.freeze
 
-    Result = Data.define(:goods_nomenclature_count, :expired_note_count)
+    Result = Data.define(:goods_nomenclature_count, :expired_note_count, :skipped)
 
     def self.call
       new.call
@@ -13,7 +13,7 @@ module TariffKnowledge
       current_version = current_source_version
       current_sids    = current_goods_nomenclature_sids
 
-      return Result.new(goods_nomenclature_count: 0, expired_note_count: 0) \
+      return Result.new(goods_nomenclature_count: 0, expired_note_count: 0, skipped: true) \
         if fingerprint_unchanged?(current_version, current_sids)
 
       DeclarableNodeLoader.call
@@ -29,6 +29,7 @@ module TariffKnowledge
       Result.new(
         goods_nomenclature_count: current_sids.size,
         expired_note_count: expired_note_count,
+        skipped: false,
       )
     end
 
--- a/app/services/tariff_knowledge/compressed_note_refresh.rb
+++ b/app/services/tariff_knowledge/compressed_note_refresh.rb
@@ -1,6 +1,7 @@
 module TariffKnowledge
   class CompressedNoteRefresh
     BATCH_SIZE = 500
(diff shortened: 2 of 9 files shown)


willfish:
> causing the frontend to hit its 15s Faraday timeout on random endpoints 

Not sure how this bit is possible since we're using a separate reader for api requests?

> The fingerprint expires after 8 days

So AFAIK we clear redis after each file load so it might be that we need to get more targeted with that first

app/services/tariff_knowledge/compressed_note_generator.rb
```diff
@@ -335,19 +341,18 @@ def context_type(context)
...
-    def upsert_note(declarable_node, attributes)
-      note = CompressedNote[declarable_node.goods_nomenclature_sid]
-      return mark_note_stale_if_context_changed(note, attributes[:context_hash]) if note&.manually_edited
+    def upsert_note(declarable_node, attributes, existing_note)
+      return mark_note_stale_if_context_changed(existing_note, attributes[:context_hash]) if existing_note&.manually_edited
 
-      if note
-        note.update(attributes)
+      if existing_note
+        existing_note.update(attributes)
```
willfish: Can we make this update conditional on `manually_edited` still being false? We load the note before the graph queries, so an admin edit during the batch can be overwritten here and the flag reset.

app/services/tariff_knowledge/compressed_note_refresh.rb
```diff
@@ -9,14 +10,21 @@ def self.call
     end
 
     def call
+      current_version = current_source_version
+      current_sids    = current_goods_nomenclature_sids
+
+      return Result.new(goods_nomenclature_count: 0, expired_note_count: 0) \
+        if fingerprint_unchanged?(current_version, current_sids)
```
willfish, line 17: Could we lock this refresh? Two workers can both miss this cache key and run the full rebuild, doubling the DB load.

app/services/tariff_knowledge/compressed_note_refresh.rb
```diff
@@ -9,14 +10,21 @@ def self.call
     end
 
     def call
+      current_version = current_source_version
+      current_sids    = current_goods_nomenclature_sids
+
+      return Result.new(goods_nomenclature_count: 0, expired_note_count: 0) \
```
willfish: Can we return a `skipped` flag and log it? `0` makes a skipped run look like there are no current goods nomenclatures, and the worker gives us no way to see whether the fast path is working.

spec/services/tariff_knowledge/compressed_note_generator_spec.rb
```diff
@@ -548,6 +548,20 @@ def node(name) = note_graph.fetch(name)
...
+    it 'fetches existing notes for the batch in one query rather than per-node lookups' do
+      create(
+        :tariff_knowledge_compressed_note,
+        goods_nomenclature_sid: 123,
+        goods_nomenclature_item_id: '0101210000',
+        content: 'Old content',
+        stale: false,
+      )
+
+      expect(TariffKnowledge::CompressedNote).not_to receive(:[])
```
willfish: Could we assert the SQL count with a few notes here? This only proves we stopped using `[]`; another per-note query would still pass.

spec/services/tariff_knowledge/source_graph_loader_spec.rb
```diff
@@ -791,6 +791,20 @@
...
+    it 'upserts nodes without a follow-up select query' do
+      create(
+        :customs_tariff_chapter_note,
+        :approved,
+        customs_tariff_update: update,
+        chapter_id: '01',
+        content: 'Chapter content.',
+      )
+
+      expect(TariffKnowledge::Node).not_to receive(:by_key)
```
willfish: Could we assert the SQL count here? This proves we stopped calling `by_key`, but another follow-up query would still pass.

spec/services/tariff_knowledge/compressed_note_refresh_spec.rb
```diff
@@ -77,5 +77,51 @@
...
+      end
+
+      it 'returns zero counts' do
+        result = described_class.call
+
+        expect(result).to have_attributes(goods_nomenclature_count: 0, expired_note_count: 0)
+      end
+    end
+
+    context 'when the declarable sids change between runs' do
```
willfish, line 146: Can we add a case where the source version changes but the SIDs stay the same? Without that, this does not prove both parts of the fingerprint invalidate the cache.

app/services/tariff_knowledge/source_graph_loader.rb
```diff
@@ -296,19 +296,20 @@ def prune_non_current_source_nodes(source_version)
...
-          .insert_conflict(target: :key, update: node_update_values)
-          .insert(values)
+      row = Node.dataset
+                .returning
+                .insert_conflict(target: :key, update: node_update_values)
+                .insert(values)
+                .first
 
-      Node.by_key(key).first
+      Node.call(row)
```
willfish: Can we use `Node.load(row)` here? Sequel documents `.call` as internal and exposes `.load` for this.

willfish (commented):
A few things to tighten up.

neilmiddleton:
> Not sure how this bit is possible since we're using a separate reader for api requests?

You're right that `UseReaderForReads` routes GETs to the reader — I should have been more precise in the PR description. The mechanism is indirect: 30 minutes of sustained upserting on the primary causes the reader replica to fall behind on WAL replay, slowing reads as it catches up. On Aurora (shared storage) there's also shared I/O contention. The New Relic correlation of the worker window with frontend timeouts is what originally pointed here, but the exact path isn't "reader blocked by writer" — it's more "reader degraded while primary is under load."

> So AFAIK we clear redis after each file load so it might be that we need to get more targeted with that first

That's a fair catch — if Redis is flushed on every file load the 8-day TTL is mostly academic. The guard still helps on days with no tariff sync (weekends etc.), but it won't survive a file-load day. Happy to scope the key more tightly or skip using a TTL-based approach altogether if you know the flush scope

willfish:
> the reader replica to fall behind on WAL replay

Okay that's really interesting :thinking:

app/services/tariff_knowledge/compressed_note_refresh.rb
```diff
@@ -9,14 +10,21 @@ def self.call
...
-      current_sids = current_goods_nomenclature_sids
       current_sids.each_slice(BATCH_SIZE) do |sids|
         CompressedNoteGenerator.call(goods_nomenclature_sids: sids)
       end
 
+      store_fingerprint(current_version, current_sids)
+
       Result.new(
         goods_nomenclature_count: current_sids.size,
         expired_note_count: expire_non_current_notes(current_sids),
```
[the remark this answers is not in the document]
neilmiddleton: Fixed in 4475988 —  is now called after  returns, and covered by a new test that stubs the private method via a named instance to avoid .

app/services/tariff_knowledge/compressed_note_generator.rb
neilmiddleton: Fixed in 4475988 — the update now uses `existing_note.this.where(manually_edited: false).update(attributes)`, so the UPDATE is a no-op at the SQL level if an admin flips the flag between the batch prefetch and the write. Covered by a new test that materialises the prefetch, updates the DB, then verifies the generator leaves the note intact.

app/services/tariff_knowledge/compressed_note_refresh.rb
neilmiddleton: I'd rather leave a lock out for now. The worker runs once a day via cron, so two instances racing is unlikely in practice. And when it does happen the outcome is benign — redundant work, not data corruption. A Redis lock adds failure modes (lock not released on crash, TTL tuning) that feel like more risk than they solve for a daily job. Happy to revisit if you feel strongly, but my preference is to keep it simple.

app/services/tariff_knowledge/compressed_note_refresh.rb
neilmiddleton: Good catch — will fix. Plan: add `skipped` to the `Result` struct and return `skipped: true` on early return, then log a tagged message in the worker so the fast path is visible in New Relic.

spec/services/tariff_knowledge/compressed_note_generator_spec.rb
neilmiddleton: Fair point — will tighten the test to assert query count using the `sql_queries` helper already in the spec file, so a per-note SELECT would actually fail it.

spec/services/tariff_knowledge/source_graph_loader_spec.rb
neilmiddleton: Same — will add a query count assertion here too.

spec/services/tariff_knowledge/compressed_note_refresh_spec.rb
neilmiddleton: Good call — will add a second case where the version increments but the SID set stays the same, to prove each half of the fingerprint independently invalidates the cache.

app/services/tariff_knowledge/source_graph_loader.rb
neilmiddleton: You're right — will switch to `Node.load(row)`.

trade-tariff-infrastructure-continuity (approved):

PR #3608, opened 2026-07-31
AI-1082: Notify operators of Customs Tariff document updates

HWallenberg:
## What:

This PR adds email notifications for operators whenever a new UK Customs
Tariff document or XI Combined Nomenclature document is imported, so the
team finds out about the update without having to watch Slack or the
database. It generalises the `Notifications::` pattern already used for
the Appendix 5a guidance-change emails so future notification pipelines
can reuse the same building blocks.

Here's what's been added:

- **`CustomsTariffUpdateNotifierService`** — looks up the imported
  `CustomsTariffUpdate`, works out which chapters/sections changed via
  `CustomsTariffUpdateChangeSummary`, builds the email personalisation,
  and hands off to the shared enqueue service
- **`Notifications::EnqueueService`**, **`Notifications::EmailWorker`**
  (Sidekiq `default` queue), **`Notifications::Instrumentation`** and
  **`Notifications::Logger`** — a shared, pipeline-agnostic set of
  building blocks for sending a Notify email with bounded retry
  (3 attempts, 30s apart) and a Slack alert if all retries are exhausted
- **`Notifications::DeliveryStatusChecker`** and
  **`CustomsTariffUpdateNotificationStatusCheckWorker`** (`default`
  queue) — follow up with GOV.UK Notify after sending to confirm the
  email actually delivered, alerting to Slack on permanent, temporary,
  or technical failure
- **`CustomsTariffUpdateChangeSummary`** — compares a newly imported
  update against the previous one to report which chapters and
  sections actually changed
- `ImportCustomsTariffDocumentWorker` and `ImportXiCnDocumentWorker`
  now call the notifier for each successfully imported version, with
  notifier failures rescued and alerted separately from import
  failures so the two don't get confused with each other in Slack
- The notification's `document_created_on` and `document_url`
  personalisation fields fall back to a readable placeholder instead
  of sending `null`, since GOV.UK Notify rejects missing personalisation
  outright and the source document's publication date is genuinely
  optional for XI Combined Nomenclature documents (it comes from an
  `OPTIONAL` SPARQL binding on the EU's own publication metadata)
- Added the real GOV.UK Notify template UUID for all three environments
  (production, staging, development), replacing the placeholders

## Why:

Right now, operators only find out about a new Customs Tariff or XI CN
import via Slack or by checking the database directly. This closes that
gap with a proper email notification, reusing the Appendix 5a email
infrastructure rather than building a one-off.

- Notifier failures are deliberately isolated per-item inside the
  import loop, so a broken notification never fails the underlying
  document import — the two are alerted to Slack separately so an
  on-call engineer isn't left guessing which one actually broke
- We chose readable fallback text over `null`/blank for optional
  personalisation fields because GOV.UK Notify's API hard-rejects a
  `null` value for any placeholder the template declares — this was
  caught via a real `Missing personalisation: document_created_on`
  error while manually testing the XI CN path locally

## Ticket:

[AI-1082](https://transformuk.atlassian.net/browse/AI-1082)

## Risk:
**Risk level:** 🟢

**Reason for rating:** No changes to measure, duty, quota, or sync
logic — this only adds an extra step at the end of an already-working
import, wrapped in its own rescue so a notification failure can't break
the import itself. Fully covered by tests (75 examples) and clean on
rubocop. Worth knowing regardless: this is the first time this pipeline
will send a real outbound email in production, so flagging it here even
though it doesn't hit any of the Amber/Red trigger categories.

## Have you? (optional)

## Deployment risks (optional)

- None beyond the new outbound email noted in the Risk section above —
  no destructive migrations, no changes to an API used in production.

### DIFF
--- a/app/services/customs_tariff_update_change_summary.rb
+++ b/app/services/customs_tariff_update_change_summary.rb
@@ -1,0 +1,56 @@
+class CustomsTariffUpdateChangeSummary
+  def initialize(update)
+    @update = update
+  end
+
+  def call
+    {
+      chapter_ids: changed_ids(CustomsTariffChapterNote, :chapter_id, previous_chapter_notes),
+      section_ids: changed_ids(CustomsTariffSectionNote, :section_id, previous_section_notes),
+    }
+  end
+
+private
+
+  def changed_ids(note_class, id_attribute, baseline_by_id)
+    note_class
+      .where(customs_tariff_update_version: @update.version)
+      .all
+      .select { |note| changed?(note, id_attribute, baseline_by_id) }
+      .map(&id_attribute)
+      .sort
+  end
+
+  def changed?(note, id_attribute, baseline_by_id)
+    baseline = baseline_by_id[note.public_send(id_attribute)]
+    return true if baseline.nil?
+
+    VersionDiffService.new(
+      note.class.name,
+      { 'content' => baseline.content },
+      { 'content' => note.content },
+    ).call.present?
+  end
+
+  def previous_update
+    start_date = @update.validity_start_date
+
+    @previous_update ||= CustomsTariffUpdate
+      .imported
+      .where { validity_start_date < start_date }
+      .order(Sequel.desc(:validity_start_date))
+      .first
+  end
+
+  def previous_chapter_notes
+    return {} unless previous_update
+
+    CustomsTariffChapterNote.where(customs_tariff_update_version: previous_update.version).all.index_by(&:chapter_id)
+  end
+
+  def previous_section_notes
+    return {} unless previous_update
+
+    CustomsTariffSectionNote.where(customs_tariff_update_version: previous_update.version).all.index_by(&:section_id)
+  end
+end
--- a/app/services/notifications/delivery_status_checker.rb
+++ b/app/services/notifications/delivery_status_checker.rb
@@ -1,0 +1,37 @@
+require_relative '../../lib/notifications/instrumentation'
+require_relative '../../lib/notifications/logger'
+
+module Notifications
+  class DeliveryStatusChecker
+    FAILURE_STATUSES = [
+      GovukNotifier::PERMANENT_FAILURE,
+      GovukNotifier::TEMPORARY_FAILURE,
+      GovukNotifier::TECHNICAL_FAILURE,
+    ].freeze
+
+    def initialize(notification_uuid, pipeline:, identifier:)
+      @notification_uuid = notification_uuid
+      @pipeline = pipeline
+      @identifier = identifier
+    end
+
+    def call
+      return if @notification_uuid.blank?
+
+      status = GovukNotifier.new.get_email_status(@notification_uuid)
+      return unless FAILURE_STATUSES.include?(status)
+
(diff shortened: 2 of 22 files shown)


app/services/notifications/delivery_status_checker.rb
```diff
@@ -0,0 +1,37 @@
+require_relative '../../lib/notifications/instrumentation'
+require_relative '../../lib/notifications/logger'
+
+module Notifications
```
willfish, line 4: Curious if we have this already for the appendix 5a stuff?

HWallenberg: Plan is to Refactor Appendix5a into this.

app/workers/notifications/email_worker.rb
```diff
@@ -0,0 +1,28 @@
+module Notifications
+  class EmailWorker
+    include Sidekiq::Worker
+
+    sidekiq_options queue: :default
```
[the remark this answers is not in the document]
HWallenberg: it was uncapped as we envisage one recipient but accepting advice to reduce noise of failing emails.  three tries is enough to flag an error.

app/services/customs_tariff_update_change_summary.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def changed?(note, id_attribute, baseline_by_id)
+    baseline = baseline_by_id[note.public_send(id_attribute)]
+    return true if baseline.nil?
+
+    VersionDiffService.new(
+      note.class.name,
+      { 'content' => baseline.content },
+      { 'content' => note.content },
+    ).call.present?
```
willfish, line 32: Only need a yes/no on content change. Plain equality is enough:

```suggestion
    baseline.content != note.content
```

app/services/notifications/enqueue_service.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def call
+      pending = @items
+      attempt = 0
+
+      until pending.empty? || attempt >= @max_attempts
+        attempt += 1
+        pending = attempt_enqueue(pending, attempt)
+        sleep(@retry_delay) if pending.any? && attempt < @max_attempts
```
willfish, line 26: This can hold the import worker for ~60s if enqueue fails. Prefer a dedicated notify worker so the import path stays fire-and-forget.

app/services/customs_tariff_update_change_summary.rb
```diff
@@ -0,0 +1,56 @@
...
+
+private
+
+  def changed_ids(note_class, id_attribute, baseline_by_id)
+    note_class
+      .where(customs_tariff_update_version: @update.version)
+      .all
+      .select { |note| changed?(note, id_attribute, baseline_by_id) }
+      .map(&id_attribute)
+      .sort
```
willfish, line 21: This only walks notes on the current version, so removed chapters/sections never appear in the email. Intentional? Worth a test or a one-liner in the PR if so.

willfish (commented):
Clean isolation of notify failures from import, and the shared Notifications pieces look like a good base for the Appendix 5a follow-up.

Thanks for the retry caps on the two workers. A few notes below.

willfish (approved):
Nothing blocking IMO

PR #3615, opened 2026-08-03
[HMRC 2498] Refactor import errors and logging

punitcse:
## What:

- [x] Add a shared importer error class between CDS and Taric importer.
- [x] Refactor the logging and move it in the error class itself to avoid duplication.

## Why:

Remove duplicate code to make it easy for any future changes. 

## Ticket:

https://transformuk.atlassian.net/jira/software/c/projects/HMRC/boards/155?selectedIssue=HMRC-2498

## Risk:
**Risk level:** 🟢
**Reason for rating:**
It is a refactoring change and does not directly affect the functionality,

### DIFF
--- a/app/lib/tariff_synchronizer/import/error.rb
+++ b/app/lib/tariff_synchronizer/import/error.rb
@@ -18,6 +18,10 @@ module TariffSynchronizer
         @source = source
         @original = original
         @context = context
+        # self.backtrace is nil until this exception is actually raised, which hasn't
+        # happened yet at construction time - caller is the best available stand-in,
+        # and matches what self.backtrace will show once `raise` does populate it.
+        set_backtrace(original&.backtrace || caller)
         log!
       end
 
@@ -30,8 +34,6 @@ module TariffSynchronizer
       def log_message
         lines = ["#{source.to_s.capitalize} import failed: #{original || message}"]
         context.each { |key, value| lines << "#{key.to_s.tr('_', ' ').capitalize}: #{value}" }
-
-        backtrace = (original || self).backtrace
         lines << "Backtrace:\n#{backtrace.join("\n")}" if backtrace
 
         lines.join("\n")
--- a/app/lib/tariff_synchronizer/import/error.rb
+++ b/app/lib/tariff_synchronizer/import/error.rb
@@ -5,6 +5,14 @@ module TariffSynchronizer
     class Error < StandardError
       attr_reader :original, :source, :context
 
+      # Follows a chain of wrapped exceptions (however many layers deep) down to the
+      # innermost one, so callers never surface a generic wrapper's message by mistake.
+      def self.unwrap(exception)
+        exception = exception.original while exception.respond_to?(:original) && exception.original
+
+        exception
+      end
+
       def initialize(message:, source:, original: nil, context: {})
         super(message)
         @source = source
--- a/app/lib/tariff_synchronizer/import/error.rb
+++ b/app/lib/tariff_synchronizer/import/error.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module TariffSynchronizer
+  module Import
+    class Error < StandardError
+      attr_reader :original, :source, :context
+
+      def initialize(message:, source:, original: nil, context: {})
+        super(message)
+        @source = source
+        @original = original
+        @context = context
+        log!
+      end
+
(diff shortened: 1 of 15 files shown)


app/lib/tariff_synchronizer/import/error.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module TariffSynchronizer
+  module Import
+    class Error < StandardError
+      attr_reader :original, :source, :context
+
+      # Follows a chain of wrapped exceptions (however many layers deep) down to the
+      # innermost one, so callers never surface a generic wrapper's message by mistake.
+      def self.unwrap(exception)
```
punitcse, line 10: ` BaseUpdateImporter#apply `and `Mailer#exception both ` unwrapped exception.original a single level (`e = e.original if e.respond_to?(:original) && e.original)`. That only surfaced the real root cause because there was exactly one place re-wrapping an already-typed error (`TaricImporter re-wraps UnknownOperationError` in `process_xml_node`). If a second layer of wrapping is ever introduced, a single-level unwrap would stop at a generic wrapper's message instead of the actual failure.

app/lib/tariff_synchronizer/import/error.rb
```diff
@@ -0,0 +1,41 @@
...
+
+        exception
+      end
+
+      def initialize(message:, source:, original: nil, context: {})
+        super(message)
+        @source = source
+        @original = original
+        @context = context
+        log!
```
HWallenberg, line 25: `log!` runs inside `initialize`, which fires when `.new` is called — but for every `raise Foo.new(...)` call site in this codebase, that happens *before* `raise` actually runs. Ruby only populates `Exception#backtrace` once `raise` unwinds the stack, so anything logged here without an `original` (e.g. `TaricImporter::UnknownOperationError` in `entity_mapper.rb`, or the `error` methods in both `xml_parser.rb` files) logs with no backtrace at all, even though a real one exists a moment later.

Fix: set the backtrace explicitly before `log!` runs, using the original exception's backtrace if there is one, otherwise `caller` (the call stack at construction time — effectively the same stack `raise` would have captured, since `.new` and `raise` are on the same line everywhere this is used). Verified with a small repro that `raise` does not overwrite a backtrace that's already been set via `set_backtrace`, so this survives the subsequent raise intact.

```suggestion
        # `raise` only fills in #backtrace when it's nil, so set it here from the caller
        # (or the original exception's own backtrace) before `log!` reads it — otherwise
        # any error raised as `raise Foo.new(...)` would log with no backtrace at all,
        # since `initialize` runs before `raise` gets a chance to unwind the stack.
        set_backtrace(original&.backtrace || caller)
        log!
```

spec/unit/tariff_synchronizer/import/error_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      described_class.new(message: 'wrapper', source: :taric, original:)
+
+      expect(Rails.logger).to have_received(:error).with(include('Backtrace:'))
+    end
+
+    it 'omits the backtrace section when neither the wrapper nor an original exception has one' do
+      described_class.new(message: 'boom', source: :taric)
+
+      expect(Rails.logger).to have_received(:error).with(satisfy { |message| !message.include?('Backtrace:') })
+    end
```
HWallenberg: This test currently locks in the backtrace bug flagged on `error.rb` as if it were intended behaviour. Once that's fixed (setting the backtrace from `caller` at construction time), `self.backtrace` is never `nil` any more — even a bare `.new()` call has a call stack behind it — so "neither the wrapper nor an original exception has one" can't happen, and this expectation would need to flip. Suggesting the replacement test below, which asserts a backtrace section is now always present when there's no `original`.

```suggestion
    it 'includes a backtrace from the construction site when there is no original exception' do
      described_class.new(message: 'boom', source: :taric)

      expect(Rails.logger).to have_received(:error).with(include('Backtrace:'))
    end
```

app/lib/tariff_synchronizer/import/error.rb
punitcse: Done in [2256651](https://github.com/trade-tariff/trade-tariff-backend/pull/3615/commits/22566513efd2255989b866395d113d92753d0b58)

HWallenberg (approved):

PR #3619, opened 2026-08-04
AI-1066: Add evaluation experiment/run/result persistence

HWallenberg:
## What:

This PR adds the database tables and models the backend needs to own AI classification evaluation data — experiments, runs, and results — so that a separate app (currently talking directly to its own Postgres database) can eventually go through the backend instead. This PR is persistence only: no API endpoints and no Admin UI yet, those are separate follow-up tickets.

- **`EvaluationExperiment`** — a named, reusable experiment definition (e.g. "baseline-gpt4o"), storing only the settings that differ from the backend's classification defaults
- **`EvaluationRun`** — one execution of an experiment: its status (`queued`/`running`/`completed`/`partially_failed`/`failed`/`cancelled`), the fully resolved configuration that was actually used, and a reproducibility digest
- **`EvaluationResult`** — the per-input outcome and diagnostic trace for one run
- **`EvaluationConfiguration::Merger`/`DigestCalculator`/`AllowlistValidator`** — small, standalone Ruby modules that combine a baseline configuration with sparse overrides, produce a deterministic fingerprint for that combination, and reject any override that isn't on an approved list
- **`EvaluationRun.start!`** and **`EvaluationResult.ingest!`** — the two entry points a future API will call: one to kick off a run (validating and resolving its configuration first), one to record a result (safely re-runnable without creating duplicates)

Three new migrations: `20260803000001_create_evaluation_experiments.rb`, `20260803000002_create_evaluation_runs.rb`, `20260803000003_create_evaluation_results.rb`. All additive — nothing existing is altered.

## Why:

Right now a separate evaluation app owns and writes to its own copy of this data directly. The plan (see Jira epic AI-1063) is to make the backend the single owner of this data, with the eval app going through backend APIs instead of its own database. This PR is step one of that: get the tables and models in place so the next ticket (AI-1065) can build the API layer on top.

- Runs are linked to experiments with a real database foreign key that **blocks deletion** of an experiment that still has runs (`ON DELETE RESTRICT`) — history must never be lost. Results are linked to runs with the opposite behaviour (`ON DELETE CASCADE`) — a run and its own results are one unit.
- Every run stores the **exact resolved configuration and a digest of it** at the moment it started, not a reference to "whatever the current settings are" — so if the backend's defaults change later, old runs still mean what they meant when they ran.
- Only a fixed, closed list of configuration overrides is accepted (e.g. `max_rounds`, `simulator_model`) — anything else is rejected before a run can even start, so the eval app can't smuggle in arbitrary settings.

## Ticket:

[AI-1066](https://transformuk.atlassian.net/browse/AI-1066) (epic: [AI-1063](https://transformuk.atlassian.net/browse/AI-1063))

## Risk:
**Risk level:** 🟢

**Reason for rating:** Additive-only migrations (three brand-new tables, nothing existing altered), no API or Admin UI wiring yet so nothing currently consumes this code, and it doesn't touch commodity codes, measures, quotas, duty calculation, or any existing endpoint. Full model/lib spec coverage (51 examples) plus the repo-wide factory-validity spec, all green; rubocop clean.

## Have you?

## Deployment risks
- None — draft PR opened to check the model/schema layer before starting the API work in AI-1065.


app/models/evaluation_experiment.rb
```diff
@@ -0,0 +1,13 @@
+class EvaluationExperiment < Sequel::Model(Sequel[:evaluation_experiments].qualify(:uk))
+  one_to_many :evaluation_runs, key: :experiment_id
+
+  def validate
+    super
+    errors.add(:name, 'cannot be blank') if name.nil? || name.strip.empty?
+    return if name.nil?
```
punitcse: Should not this also checks for empty string? You can also replace with `blank?` method which checks both empty string and `nil`?

```suggestion
    return if name.blank?
```

app/models/evaluation_result.rb
```diff
@@ -0,0 +1,17 @@
...
+  def self.ingest!(run:, source_type:, source_id:, persona:, attrs: {})
+    payload = attrs.merge(run_id: run.id, source_type:, source_id:, persona:)
+    values = new(payload).values
+    update_payload = values.reject { |key, _| IDENTITY_COLUMNS.include?(key) }
+
+    dataset
+      .insert_conflict(target: IDENTITY_COLUMNS, update: update_payload)
+      .insert(values)
+
+    first(run_id: run.id, source_id:, persona:)
```
punitcse: Nitpick but there could be RACE condition between these two calls if this record is deleted between these two calls. May be you can just call `first` when inserting?
```ruby
 dataset
      .insert_conflict(target: IDENTITY_COLUMNS, update: update_payload)
      .returning
      .insert(values) 
      .first

File: app/models/evaluation_run.rb [hunk 3]
```diff
@@ -0,0 +1,42 @@
+class EvaluationRun < Sequel::Model(Sequel[:evaluation_runs].qualify(:uk))
+  STATUSES = %w[queued running completed partially_failed failed cancelled].freeze
+
+  many_to_one :evaluation_experiment, key: :experiment_id
+  one_to_many :evaluation_results, key: :run_id
+
+  def validate
+    super
+    errors.add(:status, "must be one of #{STATUSES.join(', ')}") unless STATUSES.include?(status)
+    errors.add(:experiment_id, 'is required') if experiment_id.nil?
```
punitcse: Is there any reason why you can not use built in validation helpers like `validates_presence/validates_includes ` ?

spec/lib/evaluation_configuration/merger_spec.rb
```diff
@@ -0,0 +1,45 @@
+# spec/lib/evaluation_configuration/merger_spec.rb
```
punitcse, line 1: unnecessary comment, may want to remove it.

app/models/evaluation_run.rb
HWallenberg: Agreed replicated the pattern for other validators.

punitcse (approved):
🥮

PR #3620, opened 2026-08-05
Remove V1 API code

neilmiddleton:
## What:
Remove all V1 API code now that V1 access has been disabled.

## Why:
V1 was disabled last week (routes now return 404 with a deprecation message). This cleans up the dead code: controllers, views, the `V1Api` engine, rabl views, the `rabl` gem, and the V1 error serialisation service.

## Ticket:
N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:** Pure dead-code removal. V1 was already returning 404 before this PR — no behaviour change for any live traffic.


spencerldixon (dismissed):
LGTM get in the bin v1 👍

spencerldixon (approved):

PR #3621, opened 2026-08-05
HMRC-2500: Replace global importer notification collectors with a per-import OperationTracker

neilmiddleton:
## What:
Replaces the global `ActiveSupport::Notifications` subscriptions that `TaricImporter` and `CdsImporter` used to calculate import result counts/durations with a `TariffSynchronizer::Import::OperationTracker` created per import and injected at the points where persistence happens (the shared `BatchRecordInserter` for TARIC, `CdsImporter::RecordInserter` for CDS).

- New `TariffSynchronizer::Import::OperationMetrics` — the per-import accumulator (same aggregation shape as the old subscriber: per-operation/entity-class counts and durations, `mapping_path`, skipped-record `records` array).
- New `TariffSynchronizer::Import::OperationTracker` — times each persistence block with a monotonic clock, records via `OperationMetrics`, and republishes the *same* notification payloads as before for external observability (nothing subscribes to them to build the result anymore).
- New `TariffSynchronizer::Import::Result` — a `Hash` subclass, so `TaricUpdate`/`CdsUpdate` (which do `[:total_count]`, `#fetch`, `#to_json`) needed no changes.
- `TaricImporter`/`CdsImporter` each build one tracker per import and return `tracker.result`; the `subscribe`/`unsubscribe` machinery is gone entirely.

## Why:
TARIC and CDS importers subscribed globally to `ActiveSupport::Notifications` to calculate operation counts and durations. These subscriptions were never removed for CDS, and only unsubscribed after building the whole result for TARIC. Every import left another live subscriber holding a reference to its importer state — risking retained memory over time and, for concurrent imports, one import's events being processed by another's leftover subscriber. Each import now owns an isolated tracker instead.

## Ticket:
https://transformuk.atlassian.net/browse/HMRC-2500

## Risk:
**Risk level:** 🟠

**Reason for rating:** Behaviour-preserving refactor with full test coverage (result hash structure, oplog writes, CDS batching, staging promotion and TARIC transaction behaviour are all unchanged, verified by the existing integration suite), but it touches the core TARIC/CDS import pipeline directly, so flagging as amber rather than green.

### DIFF
--- a/app/lib/tariff_synchronizer/import/result.rb
+++ b/app/lib/tariff_synchronizer/import/result.rb
@@ -1,0 +1,10 @@
+module TariffSynchronizer
+  module Import
+    class Result < Hash
+      def initialize(operations:, total_count:, total_duration:)
+        super()
+        merge!(operations:, total_count:, total_duration:)
+      end
+    end
+  end
+end
--- a/app/lib/cds_importer.rb
+++ b/app/lib/cds_importer.rb
@@ -23,35 +23,26 @@ class CdsImporter
     CdsImporter::ExcelWriter,
   ].freeze
 
+  OPERATION_KEYS = %i[create update destroy destroy_missing skipped].freeze
+
   def initialize(cds_update, handler_classes: DEFAULT_HANDLER_CLASSES, staging_manager: nil)
     @cds_update = cds_update
     @handler_classes = handler_classes
     @staging_manager = staging_manager
-    @oplog_inserts = {
-      operations: {
-        create: { count: 0, duration: 0 },
-        update: { count: 0, duration: 0 },
-        destroy: { count: 0, duration: 0 },
-        destroy_missing: { count: 0, duration: 0 },
-        skipped: { count: 0, duration: 0 },
-      },
-      total_count: 0,
-      total_duration: 0,
-    }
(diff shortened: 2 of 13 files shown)


app/lib/tariff_synchronizer/import/result.rb
```diff
@@ -0,0 +1,10 @@
+module TariffSynchronizer
+  module Import
+    class Result < Hash
```
punitcse, line 3: A lot of other places in code uses `Data` classes to define result objects. Would it be better to use it here as it will provide immutability and prevent the result from corrupting?

neilmiddleton: Thanks for the suggestion — the `Data.define` pattern is usually my preference too. The constraint here is that `TaricUpdate` and `CdsUpdate` access the result via `[:total_count]`, `#fetch`, and `#to_json` (the last one serialises to the DB). `Data.define` doesn't support subscript access or a matching `to_json` shape without extra methods, so changing to it would mean touching the callers and potentially the stored JSON format — expanding scope for what's meant to be a behaviour-preserving refactor. The `Result` is only ever constructed once per import and immediately handed off, so the immutability concern is low in practice.

punitcse: Make sense!

punitcse (approved):

PR #3628, opened 2026-08-06
AI-1065a: Add evaluation experiment/run/result CRUD APIs

HWallenberg:
## What:

This PR adds the first slice of admin-facing HTTP APIs for the search evaluation feature. AI-1066 (already merged) added the database tables and Sequel models for evaluation experiments, runs, and results — this PR adds the controllers, serializers, and routes that let the admin app actually create and manage them over HTTP.

I've added:

- **`Api::Admin::Search::Evaluation::ExperimentsController`** — index/show/create/update for experiments (a named, reusable set of search configuration overrides)
- **`Api::Admin::Search::Evaluation::RunsController`** — index/show/create/update for runs. `create` kicks off a run via `EvaluationRun.start!`, which merges the experiment's overrides with any run-time overrides and validates them against an allowlist. `update` triggers aggregate reconciliation (result counts, costs, latency) when a run moves into a terminal status (`completed`, `partially_failed`, `failed`, `cancelled`)
- **`Api::Admin::Search::Evaluation::ResultsController`** — bulk `create` for ingesting individual search results against a run, idempotent on `(run_id, source_id, persona)` so re-running the same ingest is safe. Returns per-item success/failure so a partial batch failure doesn't sink the whole request
- Matching JSON:API serializers for all three resources
- A `rescue_from` in `Api::ApiController` mapping `EvaluationConfiguration::OverrideValidationError` (raised when a caller tries to override a search parameter that isn't on the allowlist) to a 422 response, consistent with the other generic error handling already in that controller
- Routes added under `Api::Admin::Search::Evaluation`, following the same nested-module pattern already used for `Api::Admin::GreenLanes`

New request specs cover all three controllers (66 examples).

## Why:

This is part of AI-1063/AI-1065 (search evaluation tooling), which lets us run repeatable experiments against different search configurations and compare results. These APIs are the CRUD layer the admin app will use to define experiments, start runs, and record results.

- **Namespace**: these live under `Api::Admin::Search::Evaluation`, not `Api::Internal` as originally sketched in the design doc. `Api::Admin` already has no application-level auth (same as `Api::Internal` would have had), and putting this alongside the existing `Api::Admin::GreenLanes` module keeps admin-only, non-public API surface area in one consistent place rather than spreading it across two namespaces.
- **Result ingestion is bulk and idempotent**: the eval tooling will be posting results as an experiment run progresses, potentially retrying on failure, so `EvaluationResult.ingest!` upserts on `(run_id, source_id, persona)` rather than erroring on duplicates.
- **Scope**: this PR is Tasks 1-3 of an 11-task plan. It deliberately stops here — the next slice (AI-1065b) adds a real `BaselineProvider` and a configuration-preview endpoint, and the slice after that (AI-1065c) is contingent on a still-open architecture question about whether the eval tooling or this backend owns the run-orchestration loop. None of that affects this PR; these are self-contained CRUD endpoints over already-merged persistence.

## Ticket:

## Risk:
**Risk level:** 🟢

**Reason for rating:** Purely additive — new routes, new controllers, new serializers, no changes to any existing endpoint's behaviour. No new authentication is introduced (matches the existing `Api::Admin` posture). Full request spec coverage (66 examples, all green) and rubocop clean.

## Have you?

## Deployment risks

- None

### DIFF
--- a/app/controllers/api/admin/search/evaluation/experiments_controller.rb
+++ b/app/controllers/api/admin/search/evaluation/experiments_controller.rb
@@ -12,7 +12,7 @@ module Api
           end
 
           def create
-            experiment = EvaluationExperiment.new(experiment_params)
+            experiment = EvaluationExperiment.new(experiment_params.merge(created_by: TradeTariffRequest.whodunnit))
 
             if experiment.valid? && experiment.save
               render json: serialize(experiment), status: :created
@@ -45,7 +45,7 @@ module Api
 
           def experiment_params
             params.require(:data).require(:attributes).permit(
-              :name, :description, :enabled, :created_by,
+              :name, :description, :enabled,
               configuration_overrides: {}, default_scope: {}
             ).to_h
           end
--- a/app/controllers/api/admin/search/evaluation/experiments_controller.rb
+++ b/app/controllers/api/admin/search/evaluation/experiments_controller.rb
@@ -1,0 +1,67 @@
+module Api
+  module Admin
+    module Search
+      module Evaluation
+        class ExperimentsController < AdminController
+          def index
+            render json: serialize(experiments.all, is_collection: true, meta: pagination_meta)
+          end
+
+          def show
+            render json: serialize(experiment)
+          end
+
+          def create
+            experiment = EvaluationExperiment.new(experiment_params)
+
+            if experiment.valid? && experiment.save
+              render json: serialize(experiment), status: :created
+            else
+              render json: serialize_errors(experiment), status: :unprocessable_content
+            end
+          end
+
+          def update
+            experiment.set(update_params)
+
+            if experiment.valid? && experiment.save
+              render json: serialize(experiment), status: :ok
+            else
+              render json: serialize_errors(experiment), status: :unprocessable_content
+            end
+          end
+
+        private
+
+          def serializer_class = Api::Admin::Search::Evaluation::ExperimentSerializer
+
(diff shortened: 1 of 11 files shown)


app/controllers/api/admin/search/evaluation/experiments_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+            @experiments ||= EvaluationExperiment.order(:name).paginate(current_page, per_page)
+          end
+
+          def experiment
+            @experiment ||= EvaluationExperiment.with_pk!(params[:id])
+          end
+
+          def experiment_params
+            params.require(:data).require(:attributes).permit(
+              :name, :description, :enabled, :created_by,
```
punitcse: I could be missing something here, but would `created_by` is not the logged in admin user?

app/controllers/api/admin/search/evaluation/runs_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+          end
+
+          def show
+            render json: serialize(run)
+          end
+
+          def create
+            experiment = EvaluationExperiment.with_pk!(create_params[:experiment_id])
+
+            created = EvaluationRun.start!(
```
punitcse: naming is slightly confusing as what start returns, can we call this `run` or  `evaluation_run`?

punitcse (approved):

PR #3629, opened 2026-08-07
AI-1065b: Add classification configuration discovery endpoint

HWallenberg:
### Jira link

### What?

This is the second of three small, independent PRs building the internal APIs described in AI-1065 (the first, experiment/run/result CRUD, is already merged). This one replaces a placeholder with a real implementation, and adds one new read-only endpoint.

Here's what's changed:

- **`EvaluationConfiguration::BaselineProvider`** — until now this just returned an empty hash as a placeholder. It now reads the real classification settings (things like which AI model is in use, vector search thresholds, and so on) from `AdminConfiguration` and returns them together as a single "baseline" hash. This is what an evaluation run compares itself against whenever someone overrides a setting for a single experiment.
- **`GET /search/evaluation/configuration`** — a new read-only admin endpoint that returns that baseline hash, plus the list of setting names an experiment is allowed to override. This lets the admin app show "here's what's currently configured" before someone sets up an experiment, instead of guessing or hardcoding values on the client side.

### Why?

AI-1065's evaluation experiments work by taking the *current* real settings and layering a few overrides on top, rather than requiring every experiment to specify every setting from scratch. Up to now there was nowhere to fetch what those current settings actually are — this PR builds that.

A couple of things worth knowing:

- Two of the eleven settings an experiment is allowed to override (`simulator_model` and `filter_prefixes`) are deliberately left out of the baseline hash entirely, rather than being set to `nil`. Neither has a real production equivalent today — `simulator_model` is the AI role that pretends to be a trader during an evaluation, and in production a real human fills that role, so there's nothing to read from settings. They stay "override-only".
- We hit a small autoloading gotcha while building the endpoint. A first pass added a manual `require` to work around it, which review caught as an anti-pattern for this codebase's autoloader (Zeitwerk) — it can leave dev-mode code reloading out of sync. Fixed by triggering the autoload the normal way instead, so no workaround is needed.
- While reviewing this, we noticed a separate, smaller gap: an evaluation run's "configuration digest" (used to check whether two runs used the same settings) doesn't currently account for a setting called reasoning effort. That's real, but genuinely out of scope for this PR — raised separately as [AI-1148](https://transformuk.atlassian.net/browse/AI-1148) so it isn't lost.

### Ticket:

### Risk:
**Risk level:** 🟢

**Reason for rating:** Purely additive — one new read-only endpoint, no existing endpoint changes behaviour. No new authentication is introduced (matches the rest of `Api::Admin`, which has no app-level auth today). 77 examples covering the evaluation namespace all pass, rubocop clean.

### Have you?

### Deployment risks

- None

### DIFF
--- a/app/lib/evaluation_configuration/baseline_provider.rb
+++ b/app/lib/evaluation_configuration/baseline_provider.rb
@@ -1,11 +1,20 @@
 module EvaluationConfiguration
-  # Deliberately minimal for AI-1066. AI-1065 owns sourcing the real
-  # classification baseline from AdminConfiguration and replaces this
-  # implementation; nothing else in EvaluationRun.start! needs to change
-  # when that happens, since it only depends on .call returning a Hash.
   class BaselineProvider
     def self.call
-      {}
+      # Baseline includes 9 of 11 ALLOWED_OVERRIDE_KEYS. simulator_model is omitted
+      # because production uses real humans (no backend model config; orchestration choice).
+      # filter_prefixes is omitted because it's already a plain request param (override-only).
+      {
+        'rrf_k' => AdminConfiguration.integer_value('rrf_k'),
+        'vector_score_threshold' => AdminConfiguration.integer_value('vector_score_threshold'),
+        'vector_ef_search' => AdminConfiguration.integer_value('vector_ef_search'),
+        'search_non_declarables' => AdminConfiguration.enabled?('search_non_declarables'),
+        'search_compressed_notes_enabled' => AdminConfiguration.enabled?('search_compressed_notes_enabled'),
+        'search_general_rules_enabled' => AdminConfiguration.enabled?('search_general_rules_enabled'),
+        'candidate_limit' => AdminConfiguration.integer_value('opensearch_result_limit'),
+        'max_rounds' => AdminConfiguration.integer_value('interactive_search_max_questions'),
+        'question_model' => AdminConfiguration.nested_options_value('search_model')[:selected],
+      }
     end
   end
 end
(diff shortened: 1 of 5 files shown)


spencerldixon (approved):
LGTM

app/lib/evaluation_configuration/baseline_provider.rb
```diff
@@ -1,11 +1,20 @@
...
+      {
+        'rrf_k' => AdminConfiguration.integer_value('rrf_k'),
+        'vector_score_threshold' => AdminConfiguration.integer_value('vector_score_threshold'),
+        'vector_ef_search' => AdminConfiguration.integer_value('vector_ef_search'),
+        'search_non_declarables' => AdminConfiguration.enabled?('search_non_declarables'),
+        'search_compressed_notes_enabled' => AdminConfiguration.enabled?('search_compressed_notes_enabled'),
+        'search_general_rules_enabled' => AdminConfiguration.enabled?('search_general_rules_enabled'),
+        'candidate_limit' => AdminConfiguration.integer_value('opensearch_result_limit'),
+        'max_rounds' => AdminConfiguration.integer_value('interactive_search_max_questions'),
+        'question_model' => AdminConfiguration.nested_options_value('search_model')[:selected],
```
punitcse, line 16: Wondering if it is better to move this method to AdminConfiguration as all the values were derived from it?

HWallenberg: We'll keep it as is — BaselineProvider bundles **evaluation**-specific choices (which settings, renames, omissions), and adding that into the generic AdminConfiguration would make a shared class care about one feature's business rules. Better for the small feature to depend on the generic service than the other way round.

PR #3633, opened 2026-08-10
feat(HMRC-2609): alert on GOV.UK Notify delivery failure rate

neilmiddleton:
## What:
Emit a CloudWatch custom metric from GOV.UK Notify delivery status check workers, and centralise both the `my_ott` and `appendix5a` pipelines through `Notifications::DeliveryStatusChecker` so they share the same observability path.

- Adds `aws-sdk-cloudwatch` gem
- `DeliveryStatusChecker` emits a `DeliveryFailures` count to the `TradeTariff/Notify` namespace for `permanent-failure` and `technical-failure` statuses; `temporary-failure` is excluded as it may self-resolve. Each metric carries `Environment` and `Pipeline` dimensions.
- Refactors `GovukNotifierStatusCheckWorker` to delegate through `DeliveryStatusChecker` (pipeline: `my_ott`) instead of calling `GovukNotifier` directly — it was previously invisible to Slack alerting, instrumentation, and CloudWatch metrics. User invalidation on `permanent-failure` remains in the worker.
- Also fixes a pre-existing broken spec from #3628 (`json_response['error']` → `json_response.dig('errors', 0, 'detail')`).
- Companion Terraform PR: [trade-tariff-platform-aws-terraform#705](https://github.com/trade-tariff/trade-tariff-platform-aws-terraform/pull/705)

## Why:
MyOTT subscription emails and Appendix 5a notifications could fail at scale with no operational visibility. This adds a CloudWatch metric that a companion Terraform alarm will monitor, firing to `#production-alerts` on any permanent or technical failure.

## Ticket:
https://transformuk.atlassian.net/browse/HMRC-2609

## Risk:
**Risk level:** 🟢

**Reason for rating:**
Additive observability only — no change to email sending, subscription logic, or public API surface. The ECS task role already grants `cloudwatch:PutMetricData`. The CloudWatch call is fully wrapped in a rescue so any AWS errors are logged, never raised.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem 'sequel-rails'
 # File uploads and AWS
 gem 'aws-actionmailer-ses'
 gem 'aws-sdk-cloudfront'
+gem 'aws-sdk-cloudwatch'
 gem 'aws-sdk-cloudwatchlogs'
 gem 'aws-sdk-rails'
 gem 'aws-sdk-s3'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -97,6 +97,9 @@ GEM
     aws-sdk-cloudfront (1.151.0)
       aws-sdk-core (~> 3, >= 3.254.0)
       aws-sigv4 (~> 1.5)
+    aws-sdk-cloudwatch (1.145.0)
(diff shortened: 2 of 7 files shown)


HWallenberg (approved):
cool! 🚀

PR #3636, opened 2026-08-11
AI-1065c: Allow evaluation searches to override retrieval configuration

HWallenberg:
### Jira link

### What?

This is the third and final PR of the three building the internal APIs described in AI-1065 (the first two — experiment/run/result CRUD, and the configuration-discovery endpoint — are already merged). This one adds the actual search endpoint: a way to run a real search through the same pipeline real users hit, but with a specific experiment's settings swapped in instead of the live production ones.

Here's what's changed:

- **`Api::Admin::Search::Evaluation::SearchesController`** — new endpoint, `POST /search/evaluation/searches`. Validates the overrides it's given against the same allowlist used elsewhere in AI-1065, then runs a search through `Api::Internal::SearchService` — the exact same service class real user searches go through.
- **`Api::Internal::SearchService`** — now accepts an optional `configuration_overrides` hash and an optional `search_type`. Neither is passed by the real production search route, so its behaviour is unchanged for every existing caller — this is additive only.
- **`HybridRetrievalService`, `VectorRetrievalService`, `InteractiveSearchService`** — each now accepts the handful of override values relevant to it (things like which AI model asks the questions, how many search results to fetch, vector search thresholds) instead of always reading them from the live global settings.
- **`EvaluationConfiguration::AllowlistValidator`** — dropped `filter_prefixes` from the list of overridable settings. While building this we confirmed nothing in the search pipeline actually reads it as an override — it's a real parameter elsewhere (description-intercept filtering), just not one an experiment can usefully tweak. Leaving it in the allowlist would have let someone "set" a value that silently did nothing.
- Folded in a one-line fix to `spec/requests/api/admin/search/evaluation/runs_controller_spec.rb` — a pre-existing bug from the first PR in this trio, where a spec asserted the wrong shape for a validation error response. Caught while verifying error handling for this PR; fixing it here rather than opening a separate PR for one line.

### Why?

An evaluation experiment is only useful if it can actually run a search with its own settings rather than whatever the live site is currently configured with. Up to now there was no way to do that — this PR is what makes an experiment "real" rather than just a stored configuration.

The one thing designed deliberately here: every override is threaded through as an explicit parameter with a safe default, rather than the search service reaching for global settings itself when an override is present. That's what keeps real user search completely untouched — proven by the fact every existing `Api::Internal::SearchController` request spec still passes unmodified, and by new specs confirming that omitting `configuration_overrides` entirely reproduces today's exact behaviour.

A couple of things worth knowing before this gets wired up to the eval app (AI-1073):

- **Instrumentation segregation is partial, not complete.** The main "did a search happen, how long did it take, did it fail" events (`search_started`/`search_completed`/`search_failed`) are correctly tagged with `search_type` now, so dashboards built on those will correctly separate evaluation traffic from real traffic. But several finer-grained events fired during a search — query refinement, description-intercept checks, note-evidence evaluation, question/answer emission, and a second, separate `search_failed` call used for API/embedding-level failures — still hardcode `search_type: 'interactive'` regardless of who actually triggered the search. Anyone building dashboards or alerts off those finer events specifically will currently double-count evaluation traffic as real traffic. Worth its own follow-up ticket before evaluation volume is high enough to skew those numbers.
- **AI spend/cost dashboards can't tell evaluation and production searches apart at all.** Cost tracking (`AiUsage::Instrumentation`, which records OpenAI and embedding call spend) has no `search_type` concept today, and this PR's override mechanism doesn't reach it — it's a separate logging path entirely. Evaluation runs will show up in cost reporting exactly like real user activity. Also worth a follow-up ticket, ideally before evaluation runs happen at any real volume.
- Two smaller things for AI-1073: `retrieval_method` itself (hybrid vs. vector vs. opensearch) still isn't overridable — several of the ten allowed override keys only take effect if the live site's retrieval method already happens to be `hybrid` or `vector`. And `simulator_model` is in the allowlist but isn't read anywhere in this repo — it's purely for the eval app's own use (deciding which model plays "the trader"), confirmed by checking every service that consumes overrides.

### Ticket:

### Risk:
**Risk level:** 🟠

**Reason for rating:** This one touches the core search services that live user traffic runs through (`Api::Internal::SearchService`, `HybridRetrievalService`, `VectorRetrievalService`, `InteractiveSearchService`), and adds a new endpoint meant to be consumed by another service (the eval app, once AI-1073 wires it up) — both match this repo's own criteria for an Amber change. Mitigated by explicit regression coverage: every existing production search spec passes unmodified, and dedicated specs prove that omitting overrides reproduces today's behaviour byte-for-byte. 345 examples across the evaluation namespace, search services, and production search controller all pass; rubocop clean.

### Have you?

### Deployment risks

- Modifies shared search services that also serve real production user searches — mitigated by the regression coverage described above, but worth a careful look given the blast radius if that coverage ever has a gap.
- No new authentication is introduced on the new endpoint, matching every other endpoint in `Api::Admin` today. Provisioning real network access for the eval app's own service to reach this endpoint is separate infrastructure work, tracked outside this PR (AI-1064, already done) and outside AI-1065's three PRs generally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,8 +52,3 @@ repos:
     rev: v1.7.12
     hooks:
       - id: actionlint-docker
-
-  - repo: https://github.com/trade-tariff/trade-tariff-tools
-    rev: main
-    hooks:
-      - id: debride
--- a/app/controllers/api/admin/search/evaluation/searches_controller.rb
+++ b/app/controllers/api/admin/search/evaluation/searches_controller.rb
@@ -1,0 +1,31 @@
+module Api
+  module Admin
+    module Search
+      module Evaluation
+        class SearchesController < AdminController
+          def create
+            EvaluationConfiguration::AllowlistValidator.call(configuration_overrides)
+
+            result = Api::Internal::SearchService.new(search_params.to_h.merge(configuration_overrides:, search_type: 'evaluation')).call
+
+            if result.is_a?(Hash) && result[:errors]
+              render json: result, status: :unprocessable_content
+            else
+              render json: result
+            end
+          end
+
+        private
+
+          def configuration_overrides
+            params.permit(configuration_overrides: {})[:configuration_overrides].to_h
+          end
+
+          def search_params
+            params.permit(:q, :as_of, :request_id, :expanded_query, :skip_question, answers: %i[question answer options])
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/engines/admin_api.rb
+++ b/app/engines/admin_api.rb
@@ -163,6 +163,7 @@ AdminApi.routes.draw do
           resources :runs, only: %i[index show create update]
           resources :results, only: %i[index show create]
           resource :configuration, only: [:show]
+          resources :searches, only: [:create]
         end
       end
     end
--- a/app/lib/evaluation_configuration/allowlist_validator.rb
+++ b/app/lib/evaluation_configuration/allowlist_validator.rb
@@ -13,11 +13,75 @@ module EvaluationConfiguration
   ].freeze
 
   class AllowlistValidator
+    MODEL_KEYS = %w[question_model simulator_model].freeze
+    BOOLEAN_KEYS = %w[
+      search_non_declarables
+      search_compressed_notes_enabled
+      search_general_rules_enabled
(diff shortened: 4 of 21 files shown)


anupamc (approved):
Looks all good now :+1:

PR #3637, opened 2026-08-11
[HMRC-2476] - Add Categorisation API routes that will be routed through API gateway

punitcse:
## What:

Add Categorisation API routes that will be routed through API gateway. It does not affect any green lanes routes currently in the system.

## Why:

So the categorisation API users have higher  usage limit.

## Ticket:

https://transformuk.atlassian.net/jira/software/c/projects/HMRC/boards/155?selectedIssue=HMRC-2476

## Risk:
**Risk level:** 🟢 

**Reason for rating:**

These are new routes and would not affect existing system.

### DIFF
--- a/app/controllers/api/v2/categorisation/faq_feedback_controller.rb
+++ b/app/controllers/api/v2/categorisation/faq_feedback_controller.rb
@@ -1,8 +1,9 @@
 module Api
   module V2
     module Categorisation
-      # See GoodsNomenclaturesController — reachable via API Gateway (tariff/categorisation
-      # scope) instead of the legacy static API key/token.
+      # Same behaviour as GreenLanes::FaqFeedbackController, but reachable through
+      # API Gateway under the tariff/categorisation OAuth scope instead of the legacy
+      # static API key/token. The two routes run in parallel during the migration.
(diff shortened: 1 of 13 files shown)


spencerldixon:
Can `/xi/api/categorisation` also be reached through the normal CloudFront/ALB route, bypassing API Gateway? If so, do we need to block that route or keep a Rails authentication check?

punitcse:
> Can `/xi/api/categorisation` also be reached through the normal CloudFront/ALB route, bypassing API Gateway? If so, do we need to block that route or keep a Rails authentication check?

I have added an ALB rule to block public API accessing categorisation API here https://github.com/trade-tariff/trade-tariff-platform-aws-terraform/pull/713

spencerldixon (approved):
Looks good to me, the only thing I noticed is that there `FaqFeedbackController` doesn't have a show method. This is left out of the normal green lanes controller too: https://github.com/trade-tariff/trade-tariff-backend/blob/main/app/controllers/api/v2/green_lanes/faq_feedback_controller.rb so I think it was removed at some point and the route never tidied up. We can do that in another PR though. Happy to start testing this on dev 👍

PR #3638, opened 2026-08-11
AI-1143: Commodity context

anupamc:
## What:
- Extends the VAT guidance context graph with six example commodities across tariff Chapters 20 and 84.
- Links each commodity to explicitly selected supporting sections in VAT Notices 701/14 and 744C.
- Adds VAT Notice 744C as a root guidance source.
- Captures official GOV.UK documents referenced by root guidance, with bounded traversal, redirect validation and failed-source reporting.
- Improves statutory, cross-document and aliased-path reference handling.
- Regenerates the deterministic context-graph artifact and adds integrity tests for graph hashes, unique identities, valid edge endpoints and resolved commodity evidence.

## Why:
The base context graph structures HMRC guidance, but does not demonstrate how tariff commodities can be connected to the specific guidance evidence relevant to VAT treatment.
This change proves that commodity context can be represented through explicit, auditable links rather than inferred from broad commodity-code prefixes. Chapter 20 prepared foods and Chapter 84 aviation goods provide different scenarios for testing food-related and qualifying-aircraft VAT conditions.

## Ticket:
[AI-1143](https://transformuk.atlassian.net/browse/AI-1143)

## Risk:
**Risk level:** 🟢

**Reason for rating:**
This remains an additive, offline prototype capability. It does not change existing APIs, database schemas, tariff calculations or user-facing journeys.


app/services/vat_guidance/context_graph_source.rb
```diff
@@ -1,41 +1,122 @@
...
-        read_timeout: 30,
+        open_timeout: 5,
+        read_timeout: 15,
       ) { |http| http.get(uri.request_uri, 'Accept' => 'application/json') }
-      raise "GET #{uri} failed: HTTP #{response.code}" unless response.is_a?(Net::HTTPSuccess)
+
+      if response.is_a?(Net::HTTPRedirection)
+        raise FetchError, "GET #{uri} exceeded the redirect limit" if redirects_remaining.zero?
+
+        redirected_uri = URI(response.fetch('location'))
```
HWallenberg: **Relative redirect `Location` header crashes the build.**

`URI(response.fetch('location'))` treats the redirect target as always absolute. A relative Location (legal per RFC 7231, e.g. `Location: /api/content/vat-record-keeping`) parses with `host == nil`, so the check on the next line misreads it as "redirected outside the GOV.UK Content API" and raises. Since the four root paths are fetched at `depth == 0`, `fetch_source_graph` re-raises instead of recording it as a failure — so `rake vat_guidance:context_graph` crashes on what's actually a normal redirect. Only the absolute-Location case is covered by the current spec.

Resolving the Location against the request URI (equivalent to `URI.join`) handles both the absolute and relative cases:

```suggestion
        redirected_uri = uri.merge(response.fetch('location'))
```

HWallenberg (commented):
@anupamc Suggested fixes for the bugs found in review, as suggested changes so you can accept/tweak/discuss inline. Happy to talk through any of these if the reasoning isn't clear.

app/services/vat_guidance/context_graph_builder.rb
```diff
@@ -166,22 +264,133 @@ def add_link_edge(section, link)
...
-      text.scan(SECTION_REFERENCE_PATTERN).flatten.each do |expression|
         expand_section_numbers(expression, section.node.fetch('document_id')).each do |number|
           target = resolve_section_number(section.node.fetch('document_id'), number)
           reference_text = "section #{number}"
-          next if linked_text.include?(reference_text)
 
           add_edge(section.node, target, 'prose_section_reference', reference_text, nil)
         end
       end
     end
```
HWallenberg, line 296: Two issues in this method, same fix:

1. **Suppression runs before statutory/notice detection.** The `next if expression_numbers.all? { ... }` check fires first and applies to the whole match, so if an unrelated link elsewhere in the section happens to contain the same digits (e.g. a footnote link with '105' in it), a real statutory citation like "Section 105(1) of the Civil Aviation Act 1982" gets silently dropped before `statutory_reference?` is ever checked.
2. **Partial-link expressions get spuriously duplicated.** For "sections 12, 13 and 14" where only 12 is hyperlinked, `expression_numbers.all?` is false, so the *whole* expression falls through and `expand_section_numbers` adds a `prose_section_reference` edge for 12 as well — duplicating the existing `hyperlink` edge to the same target, since `finalise_edges`'s dedup key includes `reference_kind` (so the two edges aren't deduped).

Moving statutory/notice detection first (so it isn't gated on nearby links) and filtering per-number in the final loop (instead of an all-or-nothing check) fixes both:

```suggestion
    def add_prose_edges(section, linked_section_numbers)
      fragment = Nokogiri::HTML.fragment(section.node.fetch('content_html'))
      fragment.css('a').each { |link| link.replace(" #{link.text} ") }
      text = normalise_text(fragment.text)

      text.to_enum(:scan, SECTION_REFERENCE_PATTERN).each do
        match = Regexp.last_match
        expression = match[1]

        if statutory_reference?(text, match)
          add_statutory_edge(section, text, match)
          next
        end

        referenced_notice = referenced_notice_number(text, match)
        if referenced_notice
          add_notice_reference_edges(section, expression, referenced_notice)
          next
        end

        expand_section_numbers(expression, section.node.fetch('document_id')).each do |number|
          next if linked_section_numbers.include?(number)

          target = resolve_section_number(section.node.fetch('document_id'), number)
          reference_text = "section #{number}"

          add_edge(section.node, target, 'prose_section_reference', reference_text, nil)
        end
      end
    end
```

app/services/vat_guidance/context_graph_builder.rb
```diff
@@ -6,22 +6,39 @@ module VatGuidance
   class ContextGraphBuilder
     SCHEMA_VERSION = 1
     HEADING_SELECTOR = (2..6).map { |level| "h#{level}" }.join(',').freeze
-    SECTION_REFERENCE_PATTERN = /\b(?:paragraph|section)s?\s+((?:\d+(?:\.\d+)*(?:\s*(?:,|and|to|-)\s*\d+(?:\.\d+)*)*))/i
-    NOTICE_NUMBER_PATTERN = /\b(\d{3})\s*\/\s*(\d+[A-Z]?)\b/i
+    SECTION_REFERENCE_PATTERN = %r{
+      \b(?:paragraph|section)s?\s+
+      ((?:\d+(?:\.\d+)*[A-Z]?(?:\(\d+\))*(?:\s*(?:,|and|or|to|-)\s*\d+(?:\.\d+)*[A-Z]?(?:\(\d+\))*)*))
+    }ix
+    NOTICE_NUMBER_PATTERN = /\b(?:(\d{3})\s*\/\s*(\d+[A-Z]?)|(\d{3}[A-Z]))\b/i
```
HWallenberg: The bare `\d{3}[A-Z]` alternative has no anchor to an actual "Notice" mention, so it matches any 3-digit-plus-letter token in nearby prose — e.g. "regulation 105A" — not just genuine notice numbers like "744C" in "Notice 744C". Since `referenced_notice_number` scans a 120-char window of preceding text for this pattern, an unrelated section number can hijack notice detection and produce a bogus `notice_reference` node.

Anchoring the bare form to a preceding `Notice` keyword restricts it to real notice numbers, while leaving `notice_number` callers on titles/link text like "Notice 744c" unaffected:

```suggestion
    NOTICE_NUMBER_PATTERN = /\b(?:(\d{3})\s*\/\s*(\d+[A-Z]?)|(?<=Notice\s)(\d{3}[A-Z]))\b/i
```

app/services/vat_guidance/context_graph_builder.rb
```diff
@@ -286,6 +503,10 @@ def expand_section_numbers(expression, document_id)
       available[start_index..end_index]
     end
 
+    def referenced_section_numbers(text)
+      text.to_s.scan(/\d+(?:\.\d+)*[A-Z]?(?:\(\d+\))*/i)
```
HWallenberg: **Case mismatch between heading and prose section numbers.** This regex is case-insensitive (`/i`) but preserves the source's original case in the captured group, while `resolve_section_number` (line ~419) compares numbers with plain `==`. A heading rendered as "29a." and prose referencing "section 29A" therefore never match, and the reference silently falls through to `unresolved_target` even though the section exists.

Normalising case here (and in `heading_number` below) fixes every downstream comparison at the source, instead of patching each comparison site separately:

```suggestion
    def referenced_section_numbers(text)
      text.to_s.scan(/\d+(?:\.\d+)*[A-Z]?(?:\(\d+\))*/i).map(&:upcase)
    end
```

app/services/vat_guidance/context_graph_builder.rb
```diff
@@ -304,18 +525,28 @@ def document_id(canonical_path)
...
       match = text.to_s.match(NOTICE_NUMBER_PATTERN)
-      match && "#{match[1]}/#{match[2].upcase}"
+      return match[3]&.upcase || "#{match[1]}/#{match[2].upcase}" if match
+
+      text.to_s.match(BARE_NOTICE_NUMBER_PATTERN)&.[](1)
     end
 
     def heading_number(text)
-      text.to_s.match(/\A\s*(\d+(?:\.\d+)*)\b/)&.[](1)
+      text.to_s.match(/\A\s*(\d+(?:\.\d+)*[A-Z]?(?:\(\d+\))*)\b/i)&.[](1)
```
HWallenberg: Companion fix to the `referenced_section_numbers` comment above — same case-mismatch issue applies to headings.

```suggestion
    def heading_number(text)
      text.to_s.match(/\A\s*(\d+(?:\.\d+)*[A-Z]?(?:\(\d+\))*)\b/i)&.[](1)&.upcase
    end
```

app/services/vat_guidance/context_graph_builder.rb
```diff
@@ -150,10 +174,84 @@ def build_reference_edges
...
+                   unresolved_target("#{evidence.fetch('guide_key')}##{evidence.fetch('section_key')}")
+                 end
+      add_edge(
+        source,
+        resolved,
+        'guidance_evidence',
+        "#{evidence.fetch('guide_key')}##{evidence.fetch('section_key')}",
+        target&.fetch('source_url', nil),
+      )
+    end
```
HWallenberg, line 252: **Curated catalog entries can drift silently.** `CommodityContextCatalog` evidence pairs are hand-written, hardcoded `guide_key`/`section_key` values, not free text — so an unresolved lookup here always means something is actually broken (a GOV.UK heading rename, a typo in the catalog), not a legitimate "reference not found" case like prose/hyperlink resolution can be. Right now it degrades to `resolution: 'unresolved'` exactly like those legitimate cases, so drift stays invisible until someone happens to regenerate the artifact and notices `context_graph_artifact_spec.rb`'s aggregate check fail.

Raising here instead surfaces drift immediately, at build time:

```suggestion
    def add_evidence_edge(source, evidence)
      target = @nodes.find do |node|
        node['node_type'] == 'section' &&
          node['guide_key'] == evidence.fetch('guide_key') &&
          node['section_key'] == evidence.fetch('section_key')
      end
      unless target
        raise "Unresolved commodity evidence #{evidence.fetch('guide_key')}##{evidence.fetch('section_key')} " \\
              "for #{source.fetch('id')}"
      end

      add_edge(
        source,
        resolved_target(target),
        'guidance_evidence',
        "#{evidence.fetch('guide_key')}##{evidence.fetch('section_key')}",
        target.fetch('source_url'),
      )
    end
```

HWallenberg (approved):
lgtm! 🙌

neilmiddleton (changes requested):
As I understand it, this work is a spike. Why is this therefore going anywhere near production?

PR #3648, opened 2026-08-13
AI-1156: Add evaluation gold set generation pipeline

HWallenberg:
### Jira link

[AI-1156](https://transformuk.atlassian.net/browse/AI-1156)

### What?

This PR adds the first working pipeline for building a "gold set" of search evaluation data — a curated list of realistic search phrases paired with the commodity code they *should* return, so we can later measure how good our search results actually are.

Here's what's been added:

- **`evaluation_gold_queries` table + `EvaluationGoldQuery` model** — stores each gold query: the search phrase, which commodity code it should match, and where it came from
- **`Evaluation::GoldQueryGenerator`** — calls an AI model (OpenAI, `gpt-5-mini`) with an ATaR ruling (a real historical tariff classification decision) and asks it to write three search phrases a non-expert trader might realistically type, at three levels of detail (generic/ordinary/specific). It filters out anything that looks too much like tariff jargon, retries up to 3 times if the AI's answer doesn't pass that filter, and only saves a result once all three phrases pass
- **`tariff:evaluation:generate_gold_queries` rake task** — the thing you actually run to populate the table. It's additive by default (safe to re-run, skips ATaRs that already have queries), with a `RESET=true` flag to wipe and start over, and a `LIMIT=n` flag to cap a run (handy for a quick smoke test before letting it loose on everything)
- **Read-only internal API** (`GET /uk/internal/evaluation_gold_queries` and `/:id`) — lets other tools (or a person) browse what's been generated, filterable by `persona`, `source_type`, and `source_ids`, paginated

### Why?

We can't reliably tell if a change to search actually made results better or worse without a fixed, trustworthy set of "known good" query → answer pairs to test against. This PR builds the pipeline that generates and stores that set — nothing consumes it yet, that's follow-up work.

- The AI prompt and the "is this query acceptable" filter logic are deliberately ported closely from an existing Python service (`ai-search-evaluation-suite`) rather than redesigned from scratch, so results stay comparable with what that tool already produces
- The rake task calls out to OpenAI, so it has real cost — it's wrapped in this repo's existing `AiUsage::Instrumentation` so spend is tracked, and `LIMIT` exists specifically so you don't accidentally kick off an unbounded run against the whole ATaR corpus on your first try

### Try it out

Once the migration's run and `OPENAI_API_KEY` is set, generate a small batch:

```bash
LIMIT=3 bundle exec rake tariff:evaluation:generate_gold_queries
```

Then browse what got created:

```bash
curl http://localhost:3000/uk/internal/evaluation_gold_queries
curl "http://localhost:3000/uk/internal/evaluation_gold_queries?persona=emu_generic"
```

### Have you?

- [x] Added documentation for new APIs

### Deployment risks

- The migration only adds a new table (`evaluation_gold_queries`) — no changes to existing tables, nothing destructive
- The new API route is read-only and only mounted for UK, so it can't affect XI or write any data
- The rake task is opt-in (not scheduled) and calls an external AI API, so it does have a real dollar cost per run — that's why `LIMIT` exists, to let whoever runs it try a small batch first

### Risk

**Risk level:** 🟠 Amber

**Reason for rating:** This adds a brand new API endpoint and a new external-AI call path with a real dollar cost per run, plus a small extraction from an existing shared file (`CachedCommodityDescriptionService`). None of that touches measures, duties, classification logic, or upstream CDS/HMRC sync, and the migration is purely additive — but it's new enough surface area, and new enough spend, that it's worth a quick look from the team rather than treating it as routine.


app/controllers/api/internal/evaluation_gold_queries_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+        render json: serializer_class.new(
+          paginated_dataset.all,
+          is_collection: true,
+          meta: pagination_meta,
+        ).serializable_hash
+      end
+
+      def show
+        id = Integer(params[:id], exception: false)
+        raise Sequel::NoMatchingRow unless id
```
punitcse, line 19: Would it be better to return 404 if the id is not found? This will unnecessary noise in newrelic, which is not fixable.

app/engines/internal_api.rb
```diff
@@ -8,7 +8,10 @@ class InternalApi < ::Rails::Engine
       get 'search' => 'search#search'
       get 'search_suggestions' => 'search#suggestions'
 
-      resources :atars, only: %i[index show], param: :ref if TradeTariffBackend.uk?
+      if TradeTariffBackend.uk?
```
punitcse, line 13: Why only UK? Can you add a comment to reason that please?

app/models/evaluation_gold_query.rb
```diff
@@ -0,0 +1,10 @@
+class EvaluationGoldQuery < Sequel::Model(Sequel[:evaluation_gold_queries].qualify(:uk))
+  IDENTITY_COLUMNS = %i[source_type source_id persona].freeze
```
punitcse, line 2: Suggestion: Would it be a good idea to add a model level validation for uniqueness so it fails early with nice message?

app/services/evaluation/gold_query_generator.rb
```diff
@@ -0,0 +1,172 @@
...
+      'specific' => 'emu_specific',
+    }.freeze
+
+    # Ported verbatim from ai-search-evaluation-suite's apps/product/backend/intercepts.py
+    # (_TIERED_SYSTEM) — deliberately not rewritten, see design doc.
+    SYSTEM_PROMPT = <<~PROMPT.freeze
+      You are simulating a non-expert UK trader using HMRC's tariff search.
+      Given a commodity description, write THREE search phrases at different specificity levels.
+
+      GENERIC tier: 1-3 words, vague noun phrase a trader might try first
```
punitcse, line 28: Generic tier accept one word query but `acceptable?` rejects anything < 2 words, so you may want to either change the system prompt or validation.

app/services/evaluation/gold_query_generator.rb
```diff
@@ -0,0 +1,172 @@
+module Evaluation
+  class GoldQueryGenerator
+    include HtmlToPlainText
+
+    MODEL = 'gpt-5-mini-2025-08-07'.freeze
```
punitcse, line 5: You may want to move this to configuration or env variable so without changing code we can change the AI model we are using?

lib/tasks/tariff.rake
```diff
@@ -79,6 +79,61 @@ module_function
...
+    # Ordered by ref (the natural key) so a LIMIT-capped run always picks the same
+    # subset of ATaRs across repeated invocations, rather than an arbitrary subset
+    # at Postgres's discretion.
+    rulings = TariffKnowledge::PublicAtarRuling.exclude(ref: complete_refs).order(:ref)
+    rulings = rulings.limit(limit) if limit
+
+    processed = generated = failed = 0
+    rulings.each do |ruling|
+      processed += 1
+      tiers = Evaluation::GoldQueryGenerator.call(ruling)
```
punitcse, line 121: What happened if the query to Open AI takes too long or fail. Wondering if we need a retry mechanism or may be run this in background?

app/engines/internal_api.rb
HWallenberg: ATARs are relevant to UK tariff only

app/services/evaluation/gold_query_generator.rb
HWallenberg: I think this is okay to be hardcode for now, it is just to generate the queries for the eval tool.  If it becomes an issue we can make it more configurable but it's another configuration we would need to track.

lib/tasks/tariff.rake
HWallenberg: Failures and timeouts are handled by `OpenaiClient` and in `GoldQueryGenerator#generate_tiers` so not necessary just for a rake task.

app/controllers/api/internal/evaluation_gold_queries_controller.rb
HWallenberg: It already returns a clean 404 — `Sequel::NoMatchingRow` is rescued centrally in `ApiController`. The rescued-exception noise you're seeing is a pre-existing pattern (`AtarsController#show` does the same), so fixing it here would just be inconsistent with the rest of api/internal.

app/models/evaluation_gold_query.rb
HWallenberg: Not required as the only writer (`GoldQueryGenerator#persist`) inserts via `dataset.insert_conflict(...)`, which bypasses model validations entirely.

anupamc (dismissed):

anupamc (dismissed):

anupamc (approved):
All good now - thank you :-)

PR #3650, opened 2026-08-14
[ci skip] Add docs for a new route for categorisation API

punitcse:
## What:
Add docs for a new route for categorisation API

## Why:

So we have a document of reason for changing and constraints like XI only.

## Ticket:
https://transformuk.atlassian.net/browse/HMRC-2476

## Risk:
**Risk level:** 🟢

**Reason for rating:**

Updated only docs so no real risk.


AigulDj (approved):

spencerldixon (approved):
LGTM 👍

amberstarlight (approved):

PR #3651, opened 2026-08-17
AI-1184: Make evaluation run creation retry-idempotent

HWallenberg:
## What:

This PR stops the "start an evaluation run" API from accidentally creating duplicate runs when a request gets retried.

Some background for anyone new to this part of the codebase: an "evaluation run" is one execution of a test experiment against our search — it's created by calling `POST /evaluation_runs`. Before this change, every single call to that endpoint created a brand new row in the database, no questions asked. That's a problem if the caller doesn't get a clean response back (say, the connection drops, or the request times out) — because the caller can't tell the difference between "my request never reached the server" and "it worked but I didn't hear back," the only safe thing for it to do is retry. And retrying meant creating a second, duplicate run for the exact same thing.

Changes:

- **New migration** — adds a `idempotency_key` column (nullable, with a uniqueness rule) to the `evaluation_runs` table.
- **`EvaluationRun.start!`** — now requires an `idempotency_key` argument. Before creating anything, it checks: "has a run already been made with this exact key?" If yes, it just returns that existing run instead of making a new one. There's also a small extra safety net for the rare case where two retries land on the server at almost exactly the same moment — whichever one loses that race gets handed back the winner's run instead of erroring out.
- **`RunsController#create`** — now requires the caller to send an `Idempotency-Key` header on the request. If it's missing, the API replies with a clear `400 Bad Request` immediately, rather than quietly accepting the request and going back to the old duplicate-creating behaviour.
- **`RunSerializer`** — the `idempotency_key` is now included in the API response, mainly so it's visible for debugging.

## Why:

**What's an "idempotency key" and why does the caller send it, not the server?** It's just a unique token (in practice, a UUID) that the *caller* generates once, right before it makes its first attempt to create a run. If it has to retry, it sends that exact same token again. This only works if the caller does the generating, because only the caller actually knows whether a given request is "the first attempt at this" or "a retry of something I already tried" — the server has no way to tell those apart on its own, since each incoming request looks the same to it in isolation.

We considered having the server try to spot duplicates itself (e.g. "if two requests come in for the same experiment with identical settings, treat the second one as a duplicate"), but rejected that — a person might genuinely want to run the exact same experiment twice in a row, and that approach would silently merge those into one, which is wrong. The idempotency-key approach doesn't have that problem, because it only matches requests that are *literally* retries of each other, not just similar-looking ones.

This was originally supposed to be covered when the evaluation API was first built (ticket AI-1065's own requirements said "API retries cannot create duplicate runs/results"), but it was missed at the time. It came up again during external review of PR #3648 (a different, unrelated piece of work) — confirmed as a real gap, but not caused by that PR, so it's been pulled out into its own ticket (this one) to fix properly.

One deliberate choice worth flagging: the header is *required*, not optional. That means if whoever eventually builds the client for this endpoint (tracked separately under AI-1073, not started yet) forgets to send it, they'll get an obvious, immediate error instead of the endpoint silently working in the old, duplicate-prone way. I've left a comment on that ticket explaining the contract so it isn't missed.

## Ticket:

[AI-1184](https://transformuk.atlassian.net/browse/AI-1184)

## Risk:
**Risk level:** 🟠

**Reason for rating:**
This changes the request contract of an existing API endpoint (`POST /evaluation_runs`) by making a new header mandatory — that's called out explicitly as an amber-risk change in our risk guide, since other services could consume this endpoint. In practice there's no live caller of this endpoint yet (the eval app client that will use it, AI-1073, hasn't been built), so nothing in production breaks today — but I'm rating it amber rather than green because it's still a real API contract change, not just an internal refactor.

### DIFF
--- a/app/controllers/api/admin/search/evaluation/runs_controller.rb
+++ b/app/controllers/api/admin/search/evaluation/runs_controller.rb
@@ -19,6 +19,7 @@ module Api
             evaluation_run = EvaluationRun.start!(
               experiment:,
               triggered_by: create_params[:triggered_by],
+              idempotency_key: idempotency_key!,
               run_time_overrides: create_params[:configuration_overrides] || {},
             )
 
@@ -55,6 +56,16 @@ module Api
 
           def run
             @run ||= EvaluationRun.with_pk!(params[:id])
+          end
+
+          # Required so retrying a create request (timeout, dropped response) can be
+          # told apart from a deliberate second run of the same experiment — see
+          # EvaluationRun.start!.
+          def idempotency_key!
+            key = request.headers['Idempotency-Key']
+            raise ActionController::BadRequest, 'Idempotency-Key header is required' if key.blank?
+
+            key
           end
 
           def create_params
--- a/app/controllers/api/admin/search/evaluation/runs_controller.rb
+++ b/app/controllers/api/admin/search/evaluation/runs_controller.rb
@@ -4,6 +4,13 @@ module Api
       module Evaluation
         class RunsController < AdminController
           RECONCILING_STATUSES = %w[completed partially_failed failed cancelled].freeze
+          # evaluation_runs.idempotency_key is an unbounded `text` column with a unique
+          # btree index on it — Postgres btree entries are capped at roughly a third of a
+          # page (~2.7KB on the default 8KB page size), so a sufficiently large header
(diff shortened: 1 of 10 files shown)


anupamc (approved):
LGTM :slightly_smiling_face:

PR #3652, opened 2026-08-18
AI-1145: generate VAT question journeys

anupamc:
## What:
  This PR adds the AI-1145 VAT question-journey generation and validation pipeline:

  - Defines a strict journey contract with only four trader-facing treatments: standard, reduced, zero, and exempt.
  - Adds an LLM generator and explicit cost-confirmed generation sweep for the context packets produced by AI-1144.
  - Adds deterministic validation for evidence provenance, scope, graph reachability, cycles, outcomes, exclusions, internal fall-through, and standard-by-default rules.
  - Prevents numerical VAT rates being stored in journeys; rated treatments are resolved from tariff measures at runtime.
  - Adds hash-bound candidate, review, and final artifacts with atomic, fail-closed generation tasks.
  - Commits 11 validated spike journeys covering all three target notices and six curated commodity codes, plus the local-versus-reference-expanded catering comparison.
  - Records review coverage for all 148 context packets and captures cases where no contract-safe journey could be identified.
  - Adds supporting Rake tasks, documentation, and comprehensive test coverage.

  The generated artifacts remain explicitly marked human_review_required: true and runtime_approved: false. Nothing in this PR loads them into a production route, model, database, or decision service.

## Why:
  AI-generated regulatory journeys must be mechanically constrained and auditable rather than trusted as free-form model output. This work demonstrates whether the AI-1144 context packets contain enough attributable evidence to produce usable VAT question trees, while rejecting unsupported outcomes, invented rates, unsafe defaults, and ambiguous or mixed-treatment cases.
 It also provides the feasibility evidence required by AI-1145, including a comparison showing which additional catering question becomes possible when referenced guidance is included.

## Ticket:
[AI-1145](https://transformuk.atlassian.net/browse/AI-1145)

## Risk:
**Risk level:** 🟢 

**Reason for rating:**
  This is an additive, offline spike with no production API, database, scheduled job, or trader-facing behaviour change. The artifacts require human review, are explicitly not runtime-approved, and fail closed through deterministic validation and atomic writes. The paid OpenAI generation task is opt-in and requires an explicit CONFIRM_AI_COST=true acknowledgement.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@ vendor/*
 data/*/*
 !data/vat_guidance/context_graph.json
 !data/vat_guidance/context_packets.json
+!data/vat_guidance/question_journey_candidates.json
+!data/vat_guidance/question_journeys.json
 data/failed-measures-report*
 
 # Environment files
--- a/app/services/vat_guidance/question_journey_artifact_builder.rb
+++ b/app/services/vat_guidance/question_journey_artifact_builder.rb
@@ -1,0 +1,414 @@
+require 'digest'
+
+module VatGuidance
+  class QuestionJourneyArtifactBuilder
+    BuildError = Class.new(StandardError)
+
+    TARGET_NOTICE_BY_GUIDE = {
+      'vat-notice-701-23' => '701/23',
+      'vat-notice-701-14' => '701/14',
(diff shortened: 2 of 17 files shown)


HWallenberg (approved):
lgtm

PR #3653, opened 2026-08-18
fix: validate news item date attributes

amberstarlight:
## What:
- Adds custom validator to apply bounds checking to the start dates of news items.
- Adds a bounds check for end dates of news items so that they are on or after the start date.

## Why:
Currently the model accepts any valid Sequel Date object, so you could write a start date with a year of 20026 or another malformed thing and it would be accepted. This PR adds a `MAX_YEARS_IN_FUTURE` constant to the model that sets the maximum date in the future, so we aren't letting typos or something through.

The end date was also not checked to be on or after the start date, so I have added this check too.

I've intentionally not added maximum bounds checking to the end date; looking through the `news_items` database table I did find at least one entry that had an end date over 4 years from the start date, but in theory a news item might have an end date even longer than that. A further PR in the admin repo will limit the UI to only accept dates in the format `YYYY-MM-DD`, so for our admin tool users that check is probably enough.

## Ticket:
[HMRC-2635](https://transformuk.atlassian.net/browse/HMRC-2635)

## Risk:
**Risk level:** 🟠 

**Reason for rating:**

- Adds date validation to the news item model
- Making it Amber because I don't know if 2 years is right - this will affect end users writing news stories.

### DIFF
--- a/app/models/news/item.rb
+++ b/app/models/news/item.rb
@@ -176,6 +176,9 @@ module News
     def valid_date?(date)
       max_future_date = (date + MAX_YEARS_IN_FUTURE.years).to_date
 
+      puts "date: #{date}"
+      puts "max_future_date: #{max_future_date}"
+
       date <= max_future_date
     end
   end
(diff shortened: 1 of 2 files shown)


HWallenberg (approved):
lgtm

PR #3662, opened 2026-08-20
AI-1146: Connect answer paths to commodities

anupamc:
## What:
  Implemented the AI-1146 offline proof of concept that connects validated VAT guidance answer paths to applicable tariff measures and composes commodity-specific journeys.

  The Spike:

  - Enumerates every terminal answer path and exclusion from AI-1145.
  - Creates evidence-backed path × measure proposals.
  - Supports VATZ and VATR outcomes without storing VAT percentages.
  - Chains rule fallbacks in order and permits standard treatment only after complete rule and measure exhaustion.
  - Captures and validates complete inherited commodity cohorts.
  - Independently verifies scoped commodity VAT measures against the UK Trade Tariff API.
  - Uses synthetic, hash-bound review decisions to exercise the approval workflow.
  - Produces deterministic JSON and browsable HTML artifacts covering 11 commodities and 72 composed routes.
  - Remains explicitly offline, non-production and subject to authorised tax/content review.

## Why:
  AI-1146 needs to demonstrate that the outputs from the earlier AI-1069 Spike stages can be connected end to end:

  guidance graph → context packets → question journeys → answer paths → tariff measures → commodity journeys

  This implementation tests whether:

  - Guidance rules can be associated with the correct tariff measures and inherited commodity scope.
  - Incorrect relief can be challenged using negative evidence.
  - Every relief connection can be explicitly reviewed and cryptographically bound to its evidence.
  - Commodity journeys can safely continue between rules.
  - Standard VAT treatment is only reached after all applicable relief rules have declined.
  - The same contract could later accept authorised human decisions without redesigning the workflow.

## Ticket:
[AI-1146](https://transformuk.atlassian.net/browse/AI-1146)

## Risk:
**Risk level:** 🟢

**Reason for rating:**
This is an additive, offline Spike and does not affect production tariff behaviour.


app/services/vat_guidance/spike_tariff_snapshot_refresher.rb
```diff
@@ -0,0 +1,148 @@
...
+        'cohort_verified_from_origin_response' => true,
+        'connection_commodity_codes' => CONNECTION_SCOPE.fetch(measure.fetch('measure_id')).sort,
+      )
+    end
+
+    def verify_origin_cohort!(measure, body)
+      response = JSON.parse(body)
+      expected = measure.fetch('declarable_commodity_codes').sort
+      observed = if measure.fetch('source_url').include?('/v2/measures/')
+                   measure.fetch('connection_commodity_codes').sort
```
HWallenberg: This branch compares `connection_commodity_codes` against `declarable_commodity_codes` — both come from the seed data, not from `response` (the parsed API body), so it never actually checks the live `/v2/measures/` result; the spec's `next '{}' if url.include?('/v2/measures/')` stub still passes today, which confirms nothing here reads the response.

```suggestion
                   response.fetch('included')
                           .select { |item| item['type'] == 'goods_nomenclature' }
                           .map { |item| item.dig('attributes', 'goods_nomenclature_item_id') }
                           .compact.sort
```

anupamc: Hi @HWallenberg - The latest push addresses the review comments. Kindly have a look when you can. Thanks

HWallenberg (dismissed):

PR #3668, opened 2026-08-25
Populate evaluation run cost, latency and provider-call totals from retrieval/question-asking

HWallenberg:
### What:

Today, a completed evaluation run's `total_cost_usd`, `total_latency_seconds`, and
`total_provider_calls` always read `0`/`null` — not because anything is broken, but because
nothing has ever populated the data they're summed from. This PR wires that up, plus fixes
three related bugs found while debugging real runs from Rails console once the plumbing was in
place.

Here's what's been added:

- **`EvaluationResult.provider_calls`** — new column (migration
  `20260824120000_add_provider_calls_to_evaluation_results.rb`, `Integer, null: false, default: 0`),
  permitted on the bulk results endpoint, and summed into `EvaluationRun#reconcile_aggregates!`
  alongside the existing `total_cost_usd`/`total_latency_seconds` sums.
- **`SearchesController#create`** (eval-only, `app/controllers/api/admin/search/evaluation/searches_controller.rb`) —
  subscribes to the `api_call_completed.search` and `embedding_api_call_completed.ai_usage`
  events that `InteractiveSearchService` and `VectorRetrievalService` already publish on every
  LLM/embedding call, real and eval alike, and merges the summed cost/token/duration data into
  the response as `meta.usage`. **Zero lines change in `InteractiveSearchService`,
  `Api::Internal::SearchService`, or `Search::Instrumentation`** — this taps an existing
  publish/subscribe seam rather than threading new data through the shared search code path.
- **`EvaluationRun#before_save`** — stamps `started_at` the moment a run's `status` transitions
  to `"running"`, and re-stamps it if an already-finished run is re-executed (so a stale
  `started_at` doesn't sit next to a fresh `completed_at`).
- **`EvaluationRun.start!`** — normalizes `run_time_overrides` keys to strings before merging.
  Without this, a symbol-keyed override (the natural way to write one from a Rails console, e.g.
  `{ question_model: 'gpt-5.6' }`) either silently failed to apply (the override just sat next
  to the baseline value, unused, since `'question_model'` and `:question_model` are different
  hash keys) or crashed with `ArgumentError: comparison of String with :question_model failed`
  once the merged hash's mixed key types hit `DigestCalculator`'s `keys.sort`.

### Why:

Without real cost/latency/provider-call data, the evaluation app can't do the one thing it
exists to do: benchmark the classification pipeline's real-world cost and latency, not just its
accuracy. This PR closes that gap for the Rails side (the retrieval/question-asking cost) —
simulator-side cost (the Python "trader" LLM calls) is a separate, linked ticket (AI-1226) in
`ai-search-evaluation-suite`, since it needs its own small pricing table in Python.

A few decisions worth calling out:

- **Why tap notifications instead of threading data through the search services directly:** an
  earlier draft of this design passed usage data through `InteractiveSearchService::Result`,
  `#parse_model_response`, and `Api::Internal::SearchService#build_meta` — all three sit in the
  exact code path real traders hit on the live commodity search (`Api::Internal::SearchService`
  is called from both the eval endpoint and `Api::Internal::SearchController`, real trader
  traffic). A mistake there risks breaking real search, not just eval telemetry. Rails already
  **publishes** the exact data needed via `ActiveSupport::Notifications` on every call (real and
  eval); `Search::Logger` already subscribes to one of these permanently for JSON logging. This
  PR adds a second, eval-only subscriber rather than touching the shared publishers.
- **Why the collector filters by `request_id`:** `ActiveSupport::Notifications.subscribed`
  subscribes on the single, process-wide `Fanout` notifier for the duration of the block — it is
  **not** scoped to the current thread or request. `AdminApi` (this eval endpoint) and
  `InternalApi` (real trader search) run in the same multi-threaded Puma process, so without
  filtering, this collector would also pick up cost/token data from any other search completing
  concurrently in that process. The collector only accepts events whose `payload[:request_id]`
  matches this request's own `service.request_id`, resolved before the subscription opens. This
  is exactly the concurrency bug the whole-branch review caught and fixed before this PR was
  opened — see `spec/requests/.../searches_controller_spec.rb`'s cross-request-leak regression
  test.
- **Why `started_at` is server-stamped, not client-sent:** unlike `completed_at` (which the
  Python eval client computes and sends explicitly), `started_at` is derived from the status
  transition itself — this covers every way a run can reach `"running"` without each caller
  needing its own clock, and correctly re-stamps if a finished run gets re-executed with the
  same `run_id`.

### Ticket:

[AI-1225](https://transformuk.atlassian.net/browse/AI-1225)

### Risk:

**Risk level:** 🟠 

**Reason for rating:** No production trader-facing behaviour changes — the only endpoint with
new logic (`SearchesController#create`) is eval-admin-only, not exposed to traders or consumed
by other services. `InteractiveSearchService`, `Api::Internal::SearchService`, and
`Search::Instrumentation` (the real search code path) have zero behaviour changes; the new
subscriber only reads events those files already publish, filtered to its own request. The new
migration is additive with a safe default (`not null, default 0`). Full test suite (976 spec
files) and rubocop pass clean on this branch.
Repo guidance classifies changes to a consumed API response.


anupamc:
@HWallenberg - Looks amazing as a whole 👍 
Can we please change low-risk to medium-risk as the repo guidance classifies changes to a consumed API response as medium risk. Thanks

anupamc (approved):
Looks all good now 👍

PR #3699, opened 2026-09-04
Update .env.test at localhost so specs can run outside Docker  

punitcse:
## What                                                    
                                                                         
 Swapped host.docker.internal to localhost in `.env.test` for:            
 - PGHOST                                                              
 - REDIS_URL                                                           
 - ELASTICSEARCH_URL                                                   
 - FRONTEND_HOST                                                       
 - TARIFF_QUERY_SEARCH_PARSER_URL                                      
                                                                       
This lets the RSpec suite run natively on the host against the backing services (Postgres, Redis, OpenSearch/Elasticsearch, search-parser) — whichever way those are started (natively, or via the dev-stack's published ports) — without  depending on host.docker.internal. It also brings .env.test in line with how CI already resolves these hosts (CI sets PGHOST: localhost etc. directly as job env vars).                                                             

## Why                                                                   
Our Docker images (both the production image and the local dev-stack images) strip the test bundler group and exclude spec/ .rspec/.env.test from the build  context — they're built to be lean runtime images, not test runners. That means 
bundle exec rspec currently only works in CI (via GitHub Actions service containers) or against a fully native host setup. .env.test was still configured for the container-network case (host.docker.internal), which doesn't resolve when specs run directly on the host.                                                                            

## Ticket:
N/A
## Risk:
**Risk level:** 🟢 

**Reason for rating:**
Test only change


amberstarlight (approved):
Nice 👍🏻 I think I've had these overwritten like this for a while so good to have it in the repo

willfish:
I think @neilmiddleton might use this as the host? I've replaced that hostname in `/etc/hosts`

punitcse:
> I think @neilmiddleton might use this as the host? I've replaced that hostname in `/etc/hosts`

Spoke with @neilmiddleton over slack and he is happy it to merge.

PR #3701, opened 2026-09-07
[HMRC-2664] -  Retry XICN import worker and refactor open AI and embedded services with retry helper

punitcse:
## What:
Adds retry behaviour to the XICN import worker so transient failures (network, timeouts, temporary downstream errors) are retried with exponential backoff and jitter. The change includes:

- [x] Retry only for classified transient errors (network errors, 5xx responses, DB timeouts/locks).
- [x] Remove slack alert on every time the job fails
- [x] Send slack alert only when job exhausted all retries Using sidekiq `SidekiqDeathHandler`
- [x] Prevent concurrent race surface as a UniqueConstraintViolation error causing retries. 
- [x] Refactors existing OpenAI and embedding clients to use the shared retry implementation while retaining deadline, rate-limit, and usage-instrumentation behavior.
- [x] Expands test coverage around retry exhaustion, retry delays, jitter, HTTP classification, operation deadlines, Sidekiq configuration, duplicate notifications, and unique-constraint handling.
## Why:
Intermittent transient errors during XICN imports currently cause import jobs to fail permanently and require manual reprocessing. Adding controlled retries reduces failed imports, lowers manual intervention, and improves observability of remaining failures while preserving correctness via idempotency checks.

## Ticket:

HMRC-2664

## Risk:
**Risk level:** 🟠 

**Reason for rating:**
Changes XI CN import retry semantics (SPARQL HTTP retries/backoff and increased Sidekiq retries). Although the DB race is handled by rescuing Sequel::UniqueConstraintViolation, retries can still cause duplicate side-effects, amplify Sidekiq traffic/error noise, or delay alerting for persistent upstream failures.
~Refactor for open AI and embedding services so it increases the blast radius further.~


app/lib/xi_cn_importer/document_fetcher.rb
```diff
@@ -82,25 +112,92 @@ def call
...
+          response
+        end
       end
 
       raise "SPARQL request failed: HTTP #{response.code}" unless response.is_a?(Net::HTTPSuccess)
 
       JSON.parse(response.body).dig('results', 'bindings') || []
     end
 
+    def with_retries(url: SPARQL_ENDPOINT, max_retries: MAX_RETRIES, base_delay: BASE_DELAY, jitter_range: JITTER_RANGE)
```
willfish: Can we consolidate this retry loop with the existing implementations in [`EmbeddingService#with_retry`](https://github.com/trade-tariff/trade-tariff-backend/blob/074a750dac87bee22e47f26de53f2bb01498f1d7/app/lib/embedding_service.rb#L93) and [`OpenaiClient#with_retry`](https://github.com/trade-tariff/trade-tariff-backend/blob/074a750dac87bee22e47f26de53f2bb01498f1d7/app/lib/openai_client.rb#L126)? Both already handle bounded retries with exponential backoff. A small shared helper could keep the retryable errors, attempt limits and instrumentation configurable, while preserving XICN's jitter and the OpenAI client's deadline and `Retry-After` handling.

willfish (approved):
All of this isn't live yet so I'd argue its low-risk

punitcse:
> All of this isn't live yet so I'd argue its low-risk

I marked this as amber before these refactoring changes with open ai and embed service. XICN report worker retries so this is just avoid any side effects of these changes.

PR #3708, opened 2026-09-07
HMRC-2677: enable read-only root filesystem

AigulDj:
## What:

Enables `readonly_root_filesystem` across all five `ecs-service` instances (backend-uk, backend-xi, backend-job, worker-uk, worker-xi) and pins the container uid/gid to `1000` in the Dockerfile. Shared `writable_paths` in `locals.tf`:

- all: `["/tmp", "/app/tmp", "/app/log"]`
- workers also: `["/app/data/cds_updates"]`

## Why:

Security Hub control ECS.5.

- `/app/tmp` — bootsnap, or the app fails to boot
- `/app/log` — New Relic agent log
- `/tmp` — `Tempfile` in `NotesExtractor`, general scratch
- `/app/data/cds_updates` — `CdsImporter::ExcelWriter` writes `.xlsx` files here with **no `Rails.env` guard**, driven by `CdsUpdateNotificationWorker` (workers only)

`/app/data` itself is **not** mounted — it ships tracked runtime files (`guides.csv`, `preference_codes.json`, …) that a volume would mask, so only the `cds_updates` subdirectory is mounted. `DUMP_CDS_DATA_AS_JSON=true` would also need `/app/data/cds_dumps`; it defaults to `false`.

## Ticket:

HMRC-2677 — module PR: trade-tariff/trade-tariff-platform-terraform-modules#107

## Risk:

**Risk level:** 🟠

**Reason for rating:**
Infrastructure change across five services — each gets a new task definition revision with a read-only root filesystem and an init container (deploy-order dependency). Riskiest consumer of the module.

**Validated in development** (all five instances, 2026-09-08):

- All 4 long-running services reached steady state; init container exits 0; 0 restarts
- Read-only enforced (`touch /nope` → `Read-only file system`); `/app/tmp`, `/app/log`, `/tmp` and (workers) `/app/data/cds_updates` all owned by `tariff` and writable as `tariff`
- `/app/data` siblings not masked; bootsnap cache present
- backend-uk serving live API traffic; ALB target healthy; Sidekiq running
- No `EROFS` / `EACCES` in logs
- `aws ecs execute-command` works

The `/app/data/cds_updates` write was verified directly (`touch` as `tariff`); a real CDS sync exercising `ExcelWriter` still hasn't run, but the path and permissions are confirmed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,8 +82,10 @@ COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
 
 RUN bundle config set without 'development test'
 
-RUN addgroup -S tariff && \
-  adduser -S tariff -G tariff && \
+# Pin uid/gid so the ecs-service module's writable-volume permissions init container
+# can chown the read-only-root-filesystem mounts to a known id (container_user).
+RUN addgroup -S -g 1000 tariff && \
+  adduser -S -u 1000 -G tariff tariff && \
   chown -R tariff:tariff /app && \
   chown -R tariff:tariff /usr/local/bundle
 
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -21,9 +21,9 @@ Terraform to deploy the service into AWS.
 | Name | Source | Version |
 | ---- | ------ | ------- |
 | <a name="module_ai_costs_dashboard"></a> [ai\_costs\_dashboard](#module\_ai\_costs\_dashboard) | ./modules/ai_costs_dashboard | n/a |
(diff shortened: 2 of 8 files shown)


neilmiddleton (approved):
Only nitpick: there's a rare edge case where a custom writable path could clash with one of the built-in ECS Exec paths and cause a confusing error instead of a clean validation message. Not a real risk, just worth a heads-up. not worth blocking on.

AigulDj:
> Only nitpick: there's a rare edge case where a custom writable path could clash with one of the built-in ECS Exec paths and cause a confusing error instead of a clean validation message. Not a real risk, just worth a heads-up. not worth blocking on.

Resolved here https://github.com/trade-tariff/trade-tariff-platform-terraform-modules/pull/109

PR #3710, opened 2026-09-08
[HMRC-2224] Exposes semantic roles on measure types

spencerldixon:
## What:
Exposes `semantic_roles` on measure-type endpoints and embedded measure types. 
Adds regression coverage, update Swagger documentation, and bump commodity, heading and subheading cache versions.

## Why:
Centralise measure-type classifications in the API so the frontend can remove its hardcoded role mappings in a follow-up PR.

## Ticket:
[HMRC-2224>](https://transformuk.atlassian.net/browse/HMRC-2224)

## Risk:
🟠 

**Reason for rating:**
Adds an API attribute and invalidates cached responses. Must be deployed and verify refreshed responses before merging the frontend follow-up PR.

### DIFF
--- a/app/models/measure_type.rb
+++ b/app/models/measure_type.rb
@@ -55,6 +55,17 @@ class MeasureType < Sequel::Model
     2 => 'both',
   }.freeze
 
+  SEMANTIC_ROLE_TYPES = {
+    'supplementary' => SUPPLEMENTARY_TYPES,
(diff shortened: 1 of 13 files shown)


neilmiddleton (approved):
Looks solid, nice clean addition. Good coverage on the semantic_roles method itself plus the serializer and cache key bumps for commodity, subheading and heading. No issues to flag.

PR #3719, opened 2026-09-10
HMRC-2694: look up subscriber emails over the private identity route

AigulDj:
## What:
Two changes to how the backend looks up subscriber emails:

1. `IdentityApiClient.get_email` raises `LookupError` on a non-2xx instead of returning `nil`. 404 still returns `nil` — that genuinely means no such user.
2. The client trusts the self-signed `*.tariff.internal` certificate via a new `INTERNAL_CA_PEM`, wired into `worker-uk` in Terraform, so `IDENTITY_API_HOST` can point at the private Cloud Map name.

Depends on trade-tariff/identity#349, which creates that name.

## Why:
`get_email` returned `nil` for every non-2xx, so a rate-limited lookup was indistinguishable from a subscriber with no address. `StopPressEmailWorker:16`, `MyCommoditiesEmailWorker:15` and `WatchListInvitationEmailWorker:13` all bail on a blank email, so large sends dropped 235-326 recipients each with no error anywhere — ongoing since late April. The ticket names the first two; the third has the same pattern and is fixed by the same change.

Raising makes it a Sidekiq retry and a visible failure. Every path into `get_email` is a worker via `PublicUsers::User#email`; the request path takes the email from the JWT payload and never calls the client, so no user-facing behaviour changes.

The private route removes the cause: internal calls stop leaving the VPC through NAT and CloudFront, where the WAF rate limit is keyed on the single shared NAT IP.

## Ticket:
https://transformuk.atlassian.net/browse/HMRC-2694

## Risk:
**Risk level:** 🟠

**Reason for rating:** Changes the failure behaviour of live email workers — previously silent failures now raise and retry. Terraform is additive (one env var, no resource recreation; the dev apply updated all four ECS services in place).

Notes for the reviewer:
- `terraform/locals.tf` and `worker_uk.tf` will conflict with HMRC-2677 — both add a `worker_uk_service_env_vars` local. Trivial to resolve, whichever lands second `concat`s both lists.
- `ssl_options` calls `set_default_paths` before adding the internal certificate, so the store trusts the public CAs *and* the internal one. This makes the change order independent: it can merge and sit safely on the public host until identity#349 is deployed and `IDENTITY_API_HOST` is flipped. Without that call the store would trust only the internal cert and every lookup would fail TLS on the public host.
- Activating the private route needs `IDENTITY_API_HOST=https://identity.tariff.internal:8443` in `backend-uk-worker-configuration`, plus a redeploy — the secret is read at terraform apply time and baked into the task definition, so a secret edit alone has no effect.
- Out of scope, worth its own ticket: `delete_user` still returns `response.success?`, so a non-2xx makes `ExternalUserDeletionWorker:21-23` skip the deletion and report success, leaving `external_id` set. Same bug class as `get_email`, with right-to-erasure rather than email consequences.

### DIFF
--- a/app/lib/identity_api_client.rb
+++ b/app/lib/identity_api_client.rb
@@ -1,11 +1,15 @@
 module IdentityApiClient
+  class LookupError < StandardError; end
+
   def self.get_email(username)
     return nil unless username
 
     response = user_request(:get, username)
-    if response.success?
-      JSON.parse(response.body)['user']['email']
-    end
+
+    return JSON.parse(response.body)['user']['email'] if response.success?
+    return nil if response.status == 404
+
+    raise LookupError, "Identity lookup for #{username} returned HTTP #{response.status}"
   end
 
   def self.delete_user(username)
@@ -17,9 +21,32 @@ module IdentityApiClient
 
   def self.user_request(method, username)
     url = URI.join(TradeTariffBackend.identity_api_host, '/api/users/', username).to_s
-    Faraday.public_send(method, url) do |req|
+
+    connection.public_send(method, url) do |req|
       req.headers['Authorization'] = "Token #{TradeTariffBackend.identity_api_key}"
       req.headers['Content-Type'] = 'application/json'
     end
+  end
+
+  def self.connection
+    pem = TradeTariffBackend.internal_ca_pem
+
+    if @connection.nil? || @connection_pem != pem
+      @connection_pem = pem
+      @connection = Faraday.new(ssl: ssl_options)
+    end
+
+    @connection
+  end
+
+  def self.ssl_options
+    pem = TradeTariffBackend.internal_ca_pem
+    return {} if pem.blank?
+
+    store = OpenSSL::X509::Store.new
+    store.set_default_paths
+    store.add_cert(OpenSSL::X509::Certificate.new(pem))
+
+    { cert_store: store }
   end
 end
--- a/app/lib/trade_tariff_backend/config/authentication.rb
+++ b/app/lib/trade_tariff_backend/config/authentication.rb
@@ -16,6 +16,10 @@ module TradeTariffBackend
       def identity_api_key
         ENV['IDENTITY_API_KEY']
       end
+
+      def internal_ca_pem
+        ENV['INTERNAL_CA_PEM'].presence&.gsub('\n', "\n")
+      end
     end
   end
 end
--- a/spec/lib/identity_api_client_spec.rb
+++ b/spec/lib/identity_api_client_spec.rb
@@ -27,13 +27,31 @@ RSpec.describe IdentityApiClient do
 
       it { is_expected.to eq 'email@example.com' }
 
+      context 'when the user does not exist' do
+        before do
+          stub_request(:get, "#{host}/api/users/#{username}")
+            .to_return(status: 404)
+        end
+
+        it { is_expected.to be_nil }
+      end
+
       context 'when api errors' do
         before do
           stub_request(:get, "#{host}/api/users/#{username}")
             .to_return(status: 500)
         end
 
-        it { is_expected.to be_nil }
+        it { expect { client }.to raise_error(IdentityApiClient::LookupError, /HTTP 500/) }
+      end
+
+      context 'when the request is rate limited' do
+        before do
+          stub_request(:get, "#{host}/api/users/#{username}")
+            .to_return(status: 429)
+        end
+
+        it { expect { client }.to raise_error(IdentityApiClient::LookupError, /HTTP 429/) }
       end
     end
   end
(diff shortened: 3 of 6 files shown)


app/lib/identity_api_client.rb
```diff
@@ -17,9 +21,24 @@ def self.delete_user(username)
...
+
+  def self.connection
+    Faraday.new(ssl: ssl_options)
+  end
+
+  def self.ssl_options
+    pem = TradeTariffBackend.internal_ca_pem
+    return {} if pem.blank?
+
+    store = OpenSSL::X509::Store.new
```
neilmiddleton, line 46: This store trusts the internal certificate and nothing else. `OpenSSL::X509::Store.new` starts empty, and the system CA bundle is only loaded by `set_default_paths`, so as soon as `INTERNAL_CA_PEM` is set every request is verified against a store containing one self-signed cert.

That matters because the terraform in this PR sets `INTERNAL_CA_PEM` on worker-uk, while `IDENTITY_API_HOST` still points at `https://id.trade-tariff.service.gov.uk`, which presents a publicly trusted certificate. So on deploy, and before the secret is flipped, every lookup fails TLS verification and raises.

I checked rather than assumed, building the store exactly as written here and verifying a real public chain against it:

```
PR's store (add_cert only)   -> public host verifies? false
with set_default_paths added -> public host verifies? true
internal cert trusted in A?  true
internal cert trusted in B?  true
```

So the note in the description, that merging before the secret is flipped is safe and just surfaces the existing 429s, is inverted: it would take out every stop press, my commodities and watch list invitation email, not only the ones currently being rate limited. Louder than today's silent drop, but a much wider blast radius.

One line keeps both trusted and makes the change genuinely order independent, which I think is what you intended:

```ruby
store = OpenSSL::X509::Store.new
store.set_default_paths
store.add_cert(OpenSSL::X509::Certificate.new(pem))
```

With that, this can merge and sit safely on the public host until identity#349 is deployed and the private name is verified from a worker task.

Worth a spec for it as well, since the existing `ssl_options` tests only assert the store's type and would pass either way. Something asserting a publicly trusted host still verifies against the returned store would have caught this.

neilmiddleton (changes requested):
Approach looks right, and the diagnosis matches what I traced independently: only workers reach this client, since `PublicUsers::User` has `attr_writer :email` and `Api::User::UserService` assigns from the JWT payload, so the request path never calls it. The `404 -> nil` distinction is a good refinement.

One blocking issue on the cert store, and two smaller notes.

**Blocking: the deployment order described in the PR is the unsafe one.** Detail inline on `ssl_options`. Short version: a fresh `OpenSSL::X509::Store` contains no CAs at all, so once `INTERNAL_CA_PEM` is deployed while `IDENTITY_API_HOST` is still the public host, TLS verification fails on every lookup. That turns a partial silent failure into total failure of all subscriber email until the secret is flipped. One line fixes it.

**`delete_user` is still silent.** It remains `response.success?`, so a 429 returns `false` and `ExternalUserDeletionWorker:21-23` then does nothing and reports success, leaving `external_id` set and the identity record undeleted. Same bug class as `get_email`, but with right-to-erasure consequences rather than a missed email. Fine to scope out of this PR, but worth a follow-up ticket rather than leaving it implicit.

**A new connection and cert store per request.** `connection` calls `ssl_options` on every call, so a 1,108 recipient send re-parses the PEM and allocates a store 1,108 times. Not a regression, since the old `Faraday.public_send` had no reuse either, but memoising is trivial and this is exactly the bulk path being optimised.

Worth adding to the description too: `WatchListInvitationEmailWorker:13` has the same `return if user&.email.blank?` pattern. The `get_email` change covers it automatically, so it is a documentation gap rather than a code one, but anyone reading the ticket will think only two workers are affected.

On the dependency: identity#349's plan came back as an in-place update to `aws_ecs_service` with `service_registries` added, and no replacement, so that one is safer than its amber rating suggests. The `forces replacement` lines in its plan are task definition `container_definitions`, which is normal on every deploy.

neilmiddleton (approved):
`set_default_paths` is in, so the store trusts the public CAs alongside the internal cert and the deploy ordering described in the description now holds: this can merge and sit safely on the public host until identity#349 is deployed and the secret is flipped. Memoising the connection with a PEM guard covers the other point too.

Non-blocking, but worth doing at some point: the `ssl_options` specs still only assert the store's type, so they pass with or without `set_default_paths`. A test that verifies a publicly trusted chain against the returned store is what would stop this coming back, and it is about four lines.

On the conflict with #3721: resolved from my side, no action needed here. Both branches were inserting an error class on the line straight after `module IdentityApiClient`, which git will not auto-merge even though the additions are unrelated. #3721 now defines `DeletionError` next to `delete_user` instead. `git merge-tree` against this branch's head comes back clean, and the merged file has both error classes and both methods intact, so the two can land in either order.

One cosmetic consequence: in the merged result `LookupError` sits at the top of the module and `DeletionError` sits mid-file. Once both are on main it would read better with the two together at the top. Happy to do that tidy-up afterwards rather than have either of us churn a branch for it now.

app/lib/identity_api_client.rb
```diff
@@ -1,11 +1,15 @@
...
   def self.get_email(username)
     return nil unless username
 
     response = user_request(:get, username)
-    if response.success?
-      JSON.parse(response.body)['user']['email']
-    end
+
+    return JSON.parse(response.body)['user']['email'] if response.success?
+    return nil if response.status == 404
```
neilmiddleton, line 10: A 404 here does not only mean "no such user". In `identity`'s `app/models/user.rb`, `User.find` returns nil in three cases and `app/controllers/api/users_controller.rb:8` renders all of them as 404:

- `UserNotFoundException` - genuinely no such user, which is what this branch assumes
- `ResourceNotFoundException` - the Cognito **user pool** is missing or misconfigured
- `return nil unless in_group` - the user exists but is not in the group carried by the API token

The last two are configuration faults that would affect **every** lookup, not one subscriber. If `COGNITO_USER_POOL_ID` is wrong on the identity task, or the token-to-group mapping drifts, every call 404s, `get_email` returns nil for everyone, and all three workers bail on `user.email.blank?` exactly as they do today. That is the same silent whole-send drop this PR exists to remove, just reached by a different route, and `LookupError` would never fire.

I am not sure the right answer is to raise on 404, since a genuinely deleted user is a normal case and raising there would be noisy. But the branch should at least be observable, so a bad pool config shows up as something rather than nothing:

```ruby
if response.status == 404
  Rails.logger.warn("Identity lookup for #{username} returned 404")
  return nil
end
```

Even a counter would do. The thing worth avoiding is a config error that silently drops an entire send while the PR's own error class stays quiet.

neilmiddleton (commented):
Went back over the whole diff rather than just the two points I raised first time, since my earlier approval only really covered those. Not re-blocking, and the approval stands. One of these is worth a decision before merge though, the rest are cheap.

**The 404 branch is the one I would think about** (inline on line 10). `identity`'s `User.find` returns nil for three different reasons and the controller renders all three as 404, and two of them are configuration faults that would hit every lookup, not one user. Same silent whole-send drop this PR exists to remove.

**Rollout, not code:** the description says this can merge and sit safely on the public host until identity#349 lands. Worth noting that all three callers are `sidekiq_options retry: 3`, so in that window each rate-limited lookup costs up to four requests against the same 500 rpm budget that is already the bottleneck. During a large send that makes the 429 storm worse, and can push failures onto recipients who previously got through. Not an argument against the change, just an argument for deploying it together with the `IDENTITY_API_HOST` flip rather than ahead of it.

**`worker_xi` does not get `INTERNAL_CA_PEM`.** `terraform/worker_xi.tf:34` still uses `worker_xi_secret_env_vars`, and worker-xi runs the same image and the same `ExternalUserDeletionWorker` through the same connection. Nothing breaks today, but the obvious follow-up is pointing `backend-xi-worker-configuration` at the private name, and that would fail TLS with no CA present. Adding the same concat now costs nothing.

**Nit, genuinely theoretical:** `OpenSSL::X509::Certificate.new(pem)` on line 48 parses only the first PEM block, so a chain would silently contribute just the leaf. I checked and `ecs-tls.tf` uses `tls_self_signed_cert`, which emits a single certificate, so there is no chain today and this is fine as written. Only worth changing to `OpenSSL::X509::Certificate.load(pem).each { |c| store.add_cert(c) }` if you think that secret might ever hold a real CA-issued chain.

On the earlier points: `set_default_paths` and the memoised connection both look right, and I confirmed the merged result with #3721 is clean and correct in both orders.

PR #3727, opened 2026-09-11
fix: raise when an OpenSearch bulk response reports item failures

neilmiddleton:
## What:
Adds `TradeTariffBackend::BulkResponse.check!` and calls it from both places that issue an OpenSearch `_bulk` request, so a bulk whose items failed now raises instead of returning quietly.

- `app/workers/build_index_page_worker.rb`
- `app/services/tariff_knowledge/public_atar_search_refresh.rb`

Two error classes, both descending from `BulkError`:

- `BulkRejectedError` when every failure is `es_rejected_execution_exception`, which is transient backpressure from a full write queue
- `BulkIndexingError` for anything else, for example a mapping error, which will fail identically on every retry

The error message is bounded on purpose. A bulk of 500 rejected items must not produce 500 error objects in a log line or an unreadable Slack attachment, so it reports the failure count out of the batch size, a count by error type, and the first five failures with their document ids, each reason truncated to 200 characters.

Transport level failures in `BuildIndexPageWorker` still become `IndexingError` exactly as before. Only the new bulk errors pass through unwrapped, because they already carry the index, the page number and the failed document ids, and their class is what tells you whether a retry would help.

## Why:
OpenSearch answers a `_bulk` request with HTTP 200 even when individual documents in the batch were rejected. The per item outcome lives only in the response body, as an `"errors": true` flag plus an `"error"` object on each failed item. Both call sites discarded that body, so a partially or entirely failed bulk looked exactly like a clean one.

The scenario this hides in production: `ReindexModelsWorker` (`retry: false`) drops the index and enqueues a `BuildIndexPageWorker` per page. A full reindex is when the cluster is under its heaviest write load, which is precisely when it sheds work with `es_rejected_execution_exception`. A page whose documents were all rejected returned success, and because `BuildIndexPageWorker` is also `retry: false` there was no retry, no dead job, no Slack alert and no New Relic error. The index was left permanently short of commodities and search returned nothing for them, with no signal anywhere.

`PublicAtarSearchRefresh` had the same hole with an extra consequence: it went on to queue `ScoreLabelBatchWorker` for documents that never made it into the index.

**Note for reviewers, this may surface failures that have been invisible until now.** That is the point of the change, but expect it: the first reindex after deploy could fail loudly where it previously passed quietly. If it does, the error message names the index, the page and the failing document ids, and the error class tells you whether it is transient (`BulkRejectedError`, retry) or permanent (`BulkIndexingError`, fix the document or the mapping).

**A recommendation, not changed here.** `BuildIndexPageWorker` is `retry: false`. That was survivable while failures were silent, but now that queue rejections raise, a single burst of backpressure during a reindex will fail a page with no automatic recovery. `BulkRejectedError` is exactly the case that a bounded retry with backoff would clear on its own. I would suggest a follow up giving the worker something like `retry: 3` with `sidekiq_retry_in` backing off, or retrying only on `BulkRejectedError`, but that changes the reindex failure behaviour more broadly than this fix should, so I have left it for a separate conversation.

Alerting is unaffected by `retry: false` and works today. Verified against sidekiq 8.1.7 `lib/sidekiq/job_retry.rb`: `local` hits `raise e unless msg["retry"]`, which propagates to `global`, whose else branch runs the configured death handlers. `SidekiqDeathHandler` posts to Slack unless the job sets `slack_alerts: false`, and neither `BuildIndexPageWorker` nor `ReindexModelsWorker` sets it, so both already alert. New Relic collects the raised error as well.

Written red/green: the specs stub a 200 response carrying `"errors": true` and assert the raise, the error class, the bounded message and the truncation.

## Ticket:
N/A

## Risk:
**Risk level:** 🟠

**Reason for rating:**
It touches the search indexing pipeline and turns a class of previously silent failure into a raised error, so a reindex that has been quietly dropping documents will now fail visibly and alert. No change to what is indexed or how documents are serialised, only to whether a failed bulk is noticed.

### DIFF
--- a/app/lib/trade_tariff_backend/bulk_response.rb
+++ b/app/lib/trade_tariff_backend/bulk_response.rb
@@ -1,0 +1,107 @@
+module TradeTariffBackend
+  # OpenSearch answers a `_bulk` request with HTTP 200 even when individual
+  # items in that request failed. The per item outcome only shows up in the
+  # response body, as an `"errors" => true` flag plus an `"error"` object on
+  # each failed item. A caller that ignores the body therefore treats a
+  # partially (or entirely) rejected bulk as a success.
+  #
+  # `check!` inspects that body and raises with enough detail to diagnose the
+  # failure, while keeping the message small enough for a log line or a Slack
+  # alert. A bulk of 500 rejected items must not produce 500 error objects of
+  # output, so we report the first few failures and a count by error type.
+  module BulkResponse
+    BulkError = Class.new(StandardError)
+
+    # A failure that will fail identically on every retry, such as a mapping
+    # error or a malformed document.
+    BulkIndexingError = Class.new(BulkError)
+
+    # Transient backpressure: the cluster's write queue was full and shed the
+    # work. The same documents would very likely succeed later.
+    BulkRejectedError = Class.new(BulkError)
+
+    REJECTED_ERROR_TYPE = 'es_rejected_execution_exception'.freeze
+    UNKNOWN_ERROR_TYPE = 'unknown'.freeze
+
+    # Bounds on the error message. Five failures is enough to spot a pattern,
+    # and the type counts cover the rest.
+    MAX_REPORTED_FAILURES = 5
+    MAX_REASON_LENGTH = 200
+
+    # Returns the response unchanged when every item succeeded, otherwise
+    # raises BulkRejectedError (all failures were queue rejections) or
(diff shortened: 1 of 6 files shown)


punitcse (approved):

willfish (approved):
Looks right. `_bulk` 200s with per-item failures now raise, and both call sites inspect the body.

PR #3731, opened 2026-09-11
fix: stop silently dropping acknowledged notifications and enquiries

neilmiddleton:
## What:

This PR corrects three silent failures. Each failure has the same shape. The job does not do the work that we promised to a caller. The job then reports success. Each failure has its own commit and its own red/green test.

**1. The worker loses enquiry form submissions when the Redis cache does not have the data** (`EnquiryForm::SendSubmissionEmailWorker`)

`Api::V2::EnquiryForm::SubmissionsController` writes the body of the submission to Redis. Redis keeps the body for one hour only. When the body is not in Redis, the worker wrote a log entry and stopped. Sidekiq then recorded a success. The worker now writes the same log entry, but it also raises `MissingSubmissionDataError`.

**2. The worker loses third-party notifications in the same conditions** (`NotificationsWorker`)

When `notification_data` was nil, the worker wrote a log entry and continued to a usual end. The nil branch now writes the log entry and raises `MissingNotificationDataError`. The outer `rescue ... raise` was already correct, thus this PR does not change it. The `else` branch becomes a guard clause, because the nil path does not continue.

**3. Two callers get `failed_items` and then discard it** (`Notifications::EnqueueService`)

After `MAX_ATTEMPTS`, the service instruments `enqueue_failed`, sends a message to Slack, and returns a `Result`. `Appendix5aPopulatorService` and `CustomsTariffUpdateNotifierService` ignored this `Result`. Thus `RefreshAppendix5aGuidanceWorker` and the customs tariff import completed in the usual way, but the system put no notification in a queue. The two callers now raise `Notifications::EnqueueService::EnqueueFailedError` when `result.failed_items` is not empty.

The check is at the call site, and not in the service. Only the caller knows if a notification that stays out of the queue must fail the related job. `Result` now also carries the name of the pipeline and a `failure_message`. Thus the caller and the Slack alert use the same words, and we do not write the words two times.

This PR also changes `SlackNotifierService`. When the system asks the service to send a message, but no notifier is configured, the service now writes a log entry at error level. Before this change, the service used `notifier.presence&.ping(message)`. `config/initializers/slack_notifier.rb` makes a notifier in production only, and only when a webhook is set. Thus a lost alert was also silent. The other behaviour does not change. The service still returns and does not raise. This PR does not change the spec that holds `EnqueueService` to rescue a Slack failure.

## Why:

Bugs 1 and 2 are about requests that **we already acknowledged to an external caller**. The enquiry form API returned 201 and a reference number to a member of the public. The notifications API returned 202 Accepted to a third party. After we say yes, a request that we cannot complete is a failure. The only honest result is a result that persons can see. When the worker wrote a log entry and stopped, the system counted the job as completed work. Nothing went to Sidekiq, to the death handler, or to New Relic.

The risk is real, and not only theoretical. The system puts `SendTradeTariffSubmissionEmailWorker` in a queue with `perform_in(5.minutes)`, but Redis keeps the data for one hour. This margin is small if the default queue becomes full. Also, a Redis eviction or failover removes the key immediately.

Bug 3 has the same shape, but one layer higher. The only signals were a log entry and a Slack message. Slack is weaker than it looks, for the reason above. Thus a run that notified nobody looked the same as a successful run.

A raise is a true signal here. I checked these facts, and did not assume them:

- In sidekiq 8.1.7, `lib/sidekiq/job_retry.rb` has `raise e unless msg["retry"]` in `local`. The error goes to `global`. The else branch of `global` runs `@config[:death_handlers]`. Thus `retry: false` does **not** stop the alert.
- `config/initializers/sidekiq.rb` registers `SidekiqDeathHandler`. This handler sends a message to Slack, unless the job sets `slack_alerts: false`.
- New Relic APM runs in the Sidekiq workers in production. Thus New Relic collects an error that a worker raises.

This PR does not change the retry counts. The two workers that read the cache already have a `retry` limit of 3. This limit is correct, because Redis keeps the data for one hour only. The default Sidekiq limit of 25 attempts continues for many hours. Those attempts continue after the key expires, thus they cannot be successful. A quick failure is more honest. Three attempts occur in the one-hour period, thus each retry can be successful.

Please look at this point in the review. For the customs tariff path, `ImportCustomsTariffDocumentWorker` and `ImportXiCnDocumentWorker` put the call to `CustomsTariffUpdateNotifierService` in a `rescue StandardError` for each result. That rescue writes a `customs_tariff_update_notifier_failed` log entry, sends a message to Slack, and continues. Thus those workers catch the new raise, and the import job is still successful. The gain is a different error log that you can find with grep, and a New Relic notice. The gain is not a failed job. A different result needs changes to those importer workers, and those changes are out of the scope of this PR.

## Ticket:

N/A

## Risk:

**Risk level:** 🟠

**Reason for rating:**
This PR changes behaviour on live notification paths. Jobs that completed before, when the cache had no data or the enqueue failed, now fail and retry. Thus expect new Sidekiq retries, new Slack alerts from the death handler, and new New Relic errors. This is the purpose of the change, but it is a change that persons can see in production. Also, `SlackNotifierService` now writes an error log entry each time it sends a message with no configured notifier. There are no changes to data, to the schema, or to the API contract.

### DIFF
--- a/app/services/slack_notifier_service.rb
+++ b/app/services/slack_notifier_service.rb
@@ -1,7 +1,16 @@
 class SlackNotifierService
   class << self
     def call(message)
-      notifier.presence&.ping(message)
+      if notifier.blank?
+        # The notifier is only built in production with a webhook configured,
+        # so an absent one used to make this a silent no-op. Callers treat a
+        # Slack ping as their alerting channel, so say when one is dropped.
+        Rails.logger.error('slack_notifier_unconfigured: dropped a Slack message because no notifier is configured')
+
+        return nil
+      end
+
+      notifier.ping(message)
     end
 
   private
--- a/app/services/appendix_5a_populator_service.rb
+++ b/app/services/appendix_5a_populator_service.rb
@@ -102,9 +102,13 @@ private
   end
 
   def enqueue_notifications(pending, new_count, changed_count, removed_count)
-    Notifications::EnqueueService.new(pending, pipeline: 'appendix5a') { |recipient_index|
+    result = Notifications::EnqueueService.new(pending, pipeline: 'appendix5a') { |recipient_index|
       send_to_recipient(recipient_index, new_count, changed_count, removed_count)
     }.call
+
+    return if result.failed_items.empty?
+
+    raise Notifications::EnqueueService::EnqueueFailedError, result.failure_message
   end
 
   def send_to_recipient(recipient_index, new_count, changed_count, removed_count)
--- a/app/services/customs_tariff_update_notifier_service.rb
+++ b/app/services/customs_tariff_update_notifier_service.rb
@@ -27,10 +27,14 @@ class CustomsTariffUpdateNotifierService
     changes = CustomsTariffUpdateChangeSummary.new(update).call
     personalisation = build_personalisation(update, changes)
 
-    Notifications::EnqueueService.new([@version], pipeline: PIPELINE) { |version|
+    result = Notifications::EnqueueService.new([@version], pipeline: PIPELINE) { |version|
       status_check_args = [version]
       Notifications::EmailWorker.perform_async(email, TEMPLATE_ID, personalisation, 'CustomsTariffUpdateNotificationStatusCheckWorker', status_check_args)
     }.call
+
+    return if result.failed_items.empty?
+
+    raise Notifications::EnqueueService::EnqueueFailedError, result.failure_message
   end
 
 private
--- a/app/services/notifications/enqueue_service.rb
+++ b/app/services/notifications/enqueue_service.rb
@@ -3,10 +3,19 @@ require_relative '../../lib/notifications/logger'
 
 module Notifications
   class EnqueueService
+    # Raised by callers when every attempt to enqueue was exhausted. The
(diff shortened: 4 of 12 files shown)


app/services/slack_notifier_service.rb
```diff
@@ -1,7 +1,16 @@
 class SlackNotifierService
   class << self
     def call(message)
-      notifier.presence&.ping(message)
+      if notifier.blank?
+        # The notifier is only built in production with a webhook configured,
+        # so an absent one used to make this a silent no-op. Callers treat a
+        # Slack ping as their alerting channel, so say when one is dropped.
+        Rails.logger.error('slack_notifier_unconfigured: dropped a Slack message because no notifier is configured')
```
punitcse: Would this create unnecessary noise for non production environments? Can we log this error only for production?

punitcse (dismissed):
one minor comment but looks good otherwise

willfish (dismissed):
Looks right. Cache misses now fail the job instead of returning, and the enqueue call sites raise when `failed_items` is non-empty.

willfish (approved):
LGTM. I have asked copilot 🤖 to also have a look 🤷

PR #3733, opened 2026-09-11
Preserve false values in search diagnostics

willfish:
## What:
- [x] Retain boolean `false` when extracting search diagnostic fields from CloudWatch log messages.
- [x] Include disabled switches in the configuration snapshot logged for internal search.
- [x] Keep other blank-value filtering and structured nested details unchanged.

## Why:
`compact_blank` discarded explicit false values, making disabled configuration and negative diagnostic flags indistinguishable from missing data. Preserve those values so operators can tell what configuration was used and whether a recorded condition was false.

This changes diagnostic emission and extraction only, not search execution or result selection. Consumers of the admin diagnostics response may now receive boolean `false` for fields that were previously absent; projected CloudWatch string values retain their existing representation. Historical configuration values that were never logged cannot be recovered.

## Ticket:
N/A

## Risk:
**Risk level:** Green (low-risk)

**Reason for rating:** The affected feature is unused and has not yet gone live. This is a diagnostics-only correction to preserve explicit false values, with no changes to search decisions, public tariff endpoints, credentials or infrastructure.

### DIFF
--- a/app/services/api/internal/search_service.rb
+++ b/app/services/api/internal/search_service.rb
@@ -447,7 +447,7 @@ module Api
           search_model: diagnostics_search_model,
           expand_model: model_configuration('expand_model'),
           filter_prefixes: filter_prefixes,
-        }.compact_blank.tap do |configuration|
+        }.reject { |_key, value| value != false && value.blank? }.tap do |configuration|
           configuration[:hybrid_query_guardrail_enabled] = AdminConfiguration.enabled?('hybrid_query_guardrail_enabled')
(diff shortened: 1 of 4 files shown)


neilmiddleton (approved):
Logic checks out, false and 0 are preserved while blanks still get stripped. Good test coverage on the edge cases.

trade-tariff-infrastructure-continuity (approved):

trade-tariff-infrastructure-continuity (approved):

PR #3738, opened 2026-09-14
HMRC-2655: Introduce TARIC importer, downloader services to refactor TARIC Update model

rasikasri:
## What:
Introduce Taric importer and downloader services to refactor Taric Update model

## Why:
Today, CdsUpdate and TaricUpdate (Sequel models on tariff_updates) mix three concerns:
Data persistence — state, metadata, queries (pending, applied, etc.)
Download tariff file — sync loops, date ranges, HMRC/TARIC API interaction
Import tariff data — importer wiring, staging/transaction, marking applied

Workers (CdsUpdatesSynchronizerWorker, TaricUpdatesSynchronizerWorker) and synchronizers (CdsSynchronizer, TaricSynchronizer) call into these models, which makes the sync pipeline hard to follow, test, and change safely.

This pr refactor Taric synchronisation so TaricUpdate is persistence/query models only, with download and import logic moved into dedicated service classes (TaricUpdateDownloader, TaricUpdateImporter).

## Ticket:
[HMRC-2640](https://transformuk.atlassian.net/browse/HMRC-2640)

## Risk:
**Risk level:** 🟠 

**Reason for rating:**
As data sync is a critical part of the system, this needs to be tested and verified that nothing is broken.


app/lib/tariff_synchronizer/taric_update_importer.rb
```diff
@@ -0,0 +1,52 @@
...
+      @oplog_inserts = TaricImporter.new(@taric_update, staging_manager:).import
+      # Atomically promote all staged rows into the real oplog tables.
+      # This transaction is short: the data is already on disk in the UNLOGGED
+      # staging tables, so it is just a bulk INSERT … SELECT per table.
+      staging_manager.promote!
+
+      check_oplog_inserts
+      @taric_update.mark_as_applied
+      store_oplog_inserts
+
```
willfish, line 23: Previously, insert statistics were assigned before `mark_as_applied` saved both together. Now they’re saved afterwards, leaving a crash window where an applied update has no statistics. Keep the original ordering and drop the extra `save`.

willfish (approved):

PR #3739, opened 2026-09-14
fix: Differences report worker by avoiding production filesystem writ…

punitcse:
## What:
Update `Reporting::Differences.generate` so it only passes a filename to `FastExcel` in development, and uses in-memory workbook generation in production.

- **Before:** always called `new(File.basename(object_key))`, which tried to write a local file in the container working directory.
- **After:** calls `new(nil)` in production, avoiding explicit filesystem writes on read-only root filesystems.

## Why:
Production containers only allow writes to `/tmp`, so writing the workbook to the app working directory caused the differences report worker to fail.  Does not change workbook serialisation behaviour in production (`workbook.read_string` for upload/attachment); it only removes the requirement to first write a local file path.
This change aligns `Differences` with existing reporting patterns: local file convenience in development, in-memory generation in production. It preserves report output behavior (S3 upload + email attachment) while removing an environment-specific failure mode.

## Ticket:

https://transformuk.atlassian.net/browse/HMRC-2697

## Risk:
**Risk level:** 🟢

**Reason for rating:**
Small, targeted change in report workbook initialization with test coverage for both environment branches. The production path now avoids a brittle local disk dependency and continues using the existing serialized upload flow.


neilmiddleton (approved):
Good fix. The nil-filename branch already existed in `initialize` and routes straight into the in-memory `FastExcel.open` path used elsewhere, so this isn't introducing new behaviour, just avoiding the disk write on the read-only prod filesystem. Test coverage for both env branches looks solid too.

PR #3745, opened 2026-09-15
AI-1287: Add daily query result storage foundation

willfish:
## What:
- [x] Add a table and model for one successful result per service, reporting day and query name.
- [x] Add a model helper that reuses matching results and replaces changed or explicitly forced results.
- [x] Leave the existing analytics collector, schedules and API unchanged. No production caller uses the new helper in this change.

## Why:
Provide the persistence foundation for daily query reuse, without introducing collection-attempt history or CloudWatch billing accounting. The helper stores each successful query independently so callers can retain completed work when another query fails. Production collection is deliberately not activated by this storage-only change.

## Ticket:

## Risk:
**Risk level:** 🟢

**Reason for rating:** Additive, unused storage only. No change to existing collection, scheduling, APIs or trader behaviour. The migration uses the active schema and runs independently for UK and XI.


willfish:
`Unknown` deliberately leaves `cost_complete` false. It stops polling, but does not establish that the scan statistics are final. The known subtotal remains visible while the query ID awaits reconciliation. Added an explicit comment and coverage for Unknown alongside the other uncertain states; treating it as reconciled would weaken the ledger’s safety contract.

trade-tariff-infrastructure-continuity (approved):

PR #3749, opened 2026-09-16
AI-1287: Read and aggregate stored daily analytics

willfish:
## What:
- [x] Read complete daily query groups with matching definition fingerprints from the existing result table.
- [x] Adapt the working dashboard calculations to those stored results, deduplicating frontend journey IDs and summing all their recorded AI calls and costs.
- [x] Combine latency observations and zero-result counts across days; retain hourly detail for single-day views and expose missing-day coverage.
- [x] Add opt-in jobs per service/day/query, with three database-coordinated execution lanes and no automatic Sidekiq retries. Queue only missing or changed results unless force is explicit.
- [x] Leave current API and scheduling wiring unchanged. The new reader and fan-out are not activated by this change; reads never collect data.

## Why:
The dashboard needs accurate range figures from locally stored daily results without adding publication tables or collection-attempt bookkeeping. Independent query jobs retain successful results when another query fails, while bounded concurrency controls simultaneous scans. Missing or incompatible query groups remain coverage gaps, not zero traffic. Existing request-based rate denominators remain separate from the distinct-journey headline.

## Ticket:

## Risk:
**Risk level:** 🟢

**Reason for rating:** Additive read-only aggregation and opt-in query jobs with no active API or scheduled caller. No schema changes or changes to existing rolling snapshot behaviour. AWS submission retries remain disabled; successful stored results are reused.


willfish:
The trend now retains request buckets with zero frontend journeys as well as journey-only buckets. Costs intentionally remain inside the selected UTC reporting dates: the agreed metric is period activity, not lifetime cost of journeys starting in that period. Added an explicit boundary example showing later collected calls for the same ID stay outside a single-day selection; calls across midnight are included when both dates are selected.

trade-tariff-infrastructure-continuity (approved):

PR #3765, opened 2026-09-17
Load chemicals in one query when searching chemicals by name

spencerldixon:
## What:
  This PR makes the chemical name search load its results with a fixed number of database queries. Before, the search ran one extra query for each name that matched.

  - The name search in `ChemicalSearchService` now uses Sequel eager loading (`.eager(:chemical)`) to load the chemical for every matching name in one query.
  - The search results, the sort order, and the API response do not change.
  - A new service spec checks that each matching chemical is returned once and that only one query reads the `chemicals` table.

  ## Why:
  The Trade Tariff service lets traders find the commodity code for the goods they import or export. Some goods are chemicals. The backend has a public API endpoint, `GET /api/v2/chemicals/search`, that finds
  chemicals by a CAS number (a unique registry number for a chemical) or by name. The frontend calls this endpoint when a user searches by chemical name.

  Each chemical can have many names (for example, "citric acid" and "citric acid monohydrate"). The name search first finds all names that contain the search text. It then loads the chemical for each name.
  
  Before this change, it loaded each chemical with a separate query. A search for "acid" that matched 500 names ran 500 extra queries. This pattern is known as an N+1 query problem, and it makes the search slow and adds load to the database.

  After this change, the search loads all the chemicals for the matching names in one query. A search now runs two queries in total, whatever the number of matching names. The search result is cached for 24 hours, so the first search for each term gets the benefit.

  ## Ticket:
  N/A

  ## Risk:
  **Risk level:** 🟢 Green

  **Reason for rating:** This is a refactor with no behaviour change to a public endpoint: the query results, ordering, and cached objects are the same, and the new spec asserts both the returned chemicals and the single `chemicals` query.


neilmiddleton (approved):
Nice fix. Eager loading with the spec asserting a single query for the chemicals table is exactly what I'd want to see for an N+1 fix like this.

PR #3769, opened 2026-09-17
AI-1287: Build versioned search analytics read models

willfish:
## What:
- [x] Add derived journey observations, daily and hourly rollups, and source revision metadata.
- [x] Add an explicit SQL rebuild task with per-date processing and bounded database resource settings.
- [x] Publish each generation atomically. Keep the previous generation if a rebuild fails.
- [x] Preserve cross-day identities and reject narrower replacement ranges that would remove existing coverage.

## Why:
Search analytics reads currently expand large identifier collections in Ruby. Compact derived data lets the dashboard read counts without loading those collections. Stored query results remain the source of truth and are not changed by a rebuild.

Run `bundle exec rake search_analytics:rebuild_read_model` in the selected backend service after migration. The database role must be allowed to set `temp_file_limit`. This change does not activate API reads or schedule rebuilds.

## Ticket:

## Risk:
**Risk level:** 🟠

**Reason for rating:** An explicit rebuild writes derived data on the shared database. Per-date processing and SQL limits bound the work, but database capacity needs team review before a production rebuild. The migration only adds new objects. Tariff data, original query results and permissions are unchanged.


willfish:
Superseded by #3772 and #3773. The replacement uses ordinary SQL views and indexed materialized views, following the existing repository idiom. It removes the custom generation tables and polling design. Closing this proposal while retaining its review history.

PR #3770, opened 2026-09-17
AI-1287: Reduce memory use in search analytics reads

willfish:
## What:
- [x] Read compact journey counts and outcomes from compatible SQL generations.
- [x] Aggregate and rank search terms in PostgreSQL before applying the display limit.
- [x] Pin source revisions and the selected generation in one read-only transaction.
- [x] Preserve the response contract, date coverage, cross-day journeys and recorded cost attribution.
- [x] Use the existing reader when derived data is absent or outdated.

## Why:
The dashboard needs counts and totals, not millions of identifiers in the web process. This read path avoids loading identifier and term-history blobs into Ruby. It does not add Redis caching or worker caching.

Build a current generation before expecting faster reads. The fallback retains the previous latency and memory cost until a rebuild completes. Request-local database memory settings do not change the connection pool defaults.

## Ticket:

## Risk:
**Risk level:** 🟢

**Reason for rating:** Covered read-only optimisation for a dashboard that is not in active use. Metric definitions and the API response stay the same. Incompatible derived data is not served. Collection, tariff data and permissions are unchanged.


willfish:
Superseded by #3773, which reads native materialized views instead of custom generation tables. The replacement retains the metric contract and source freshness checks.

PR #3771, opened 2026-09-17
AI-1287: Maintain bootstrapped analytics read models

willfish:
## What:
- [x] Check bootstrapped read models every fifteen minutes on the batch queue.
- [x] Skip services without a generation and source revisions that have not changed.
- [x] Leave an existing rebuild alone when its advisory lock is held.
- [x] Report other failures through normal job handling, without automatic job retries.

## Why:
New daily results and forced refreshes can make derived data outdated. Periodic maintenance restores the fast read path without adding work to dashboard requests or CloudWatch collection jobs.

An operator must build the first generation explicitly. After that bootstrap, scheduled checks can rebuild the available data in the rolling 366-day window. The worker uses SQL generation building, not an in-memory result cache.

## Ticket:

## Risk:
**Risk level:** 🟠

**Reason for rating:** After bootstrap, this adds bounded derived-data writes to the shared database. Review database capacity and the worker role's permission to set the required SQL resource limits before production use. Queue priority does not isolate database resources. Tariff data and original query results are unchanged.


willfish:
Superseded by #3773. Refresh now follows collection-completion signals after explicit bootstrap, rather than a fifteen-minute polling schedule.

PR #3772, opened 2026-09-18
AI-1287: Add efficient search analytics views

willfish:
## What:
- [x] Create ordinary SQL views and indexed materialized views through versioned Sequel migrations.
- [x] Keep the existing successful query-result rows as the source of truth.
- [x] Add an explicit refresh task with atomic, dependency-ordered refresh and bounded SQL settings.
- [x] Support concurrent refresh after initial population and retain source revisions for freshness checks.
- [x] Populate test materialized views in dependency order, including dependencies through ordinary views.

## Why:
The dashboard needs compact counts rather than millions of identifiers in Ruby. PostgreSQL views express the transformations and materialized views retain expensive results. This follows the repository's read-side idiom without an operation log, custom generation tables or polling.

The migration creates unpopulated views. Run `bundle exec rake search_analytics:refresh_views` in the selected service to populate them. The database role must own the views and be allowed to set `temp_file_limit`. Refresh recomputes whole relations and needs database headroom; its memory setting is per operation, not a total process cap. This change does not activate API reads or automatic refresh.

## Ticket:

## Risk:
**Risk level:** 🟢

**Reason for rating:** Additive storage for an unused internal feature, with explicit population and safe unpopulated defaults. Existing source data, collection and API behaviour are unchanged. No permissions are granted and no existing tariff objects are replaced.


docs/architecture/search-analytics-materialized-views.md
```diff
@@ -0,0 +1,71 @@
...
+- `search_analytics_journey_rollup_rows` composes daily/hourly statistics for
+  journeys confined to one date.
+- `search_analytics_journey_rollup_totals` materializes those statistics.
+- `search_analytics_source_revisions` materializes source identity, fingerprints
+  and collection timestamps. It is current-state metadata, not a change history.
+
+The API combines additive rollups with exact reconciliation of repeated
+identities. It does not sum daily unique counts. It aggregates term rankings in
+PostgreSQL and does not transfer identifier collections or full term histories
+to Ruby.
```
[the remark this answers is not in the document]
willfish: Fixed on e13fdd32b. The architecture document now describes only the storage and refresh added here. It no longer presents API snapshot reads, a faster API path, or PostgreSQL term ranking as current behaviour. The suppressed comment on the snapshot paragraph is covered by the same change.

neilmiddleton (approved):
Looks solid. Additive views/migration with no data population baked in, and the test coverage on lock contention, concurrent refresh and rollback gives me confidence this is safe to merge.

app/services/search_analytics/materialized_views.rb
```diff
@@ -0,0 +1,134 @@
...
+      db.with_advisory_lock(lock_id, wait:) do
+        populated = populated?
+        next false if only_if_populated && !populated
+
+        db.transaction(isolation: :repeatable) do
+          if !force && populated && source_revisions_match_live?
+            false
+          else
+            apply_local_settings
+            refresh_matviews(concurrently: concurrently && populated)
```
punitcse, line 51: I think you can not refresh materialised view inside a DB transaction in Postgres. This will raise an error.

willfish: Thanks for checking this, Punit. I ran a small PostgreSQL 18.6 reproduction to confirm: both ordinary refresh and REFRESH MATERIALIZED VIEW CONCURRENTLY succeed inside BEGIN ISOLATION LEVEL REPEATABLE READ, and ROLLBACK restores the previous view contents.

The transaction restriction applies to CREATE INDEX CONCURRENTLY rather than concurrent materialized-view refresh. The latter requires a populated view and a qualifying unique index, which this implementation provides. Keeping these refreshes in one transaction lets us publish the related views together or roll them all back on failure. The materialized-view specs also cover transactional concurrent refresh and rollback.

Resolving this thread on that basis. Thanks for raising it.

PR #3783, opened 2026-09-21
Fix Difference report worker by memoising the workbook data

punitcse:
## What:
Update report generation to avoid double-reading `FastExcel` temp files and to stop relying on a writable app filesystem in production.

- Ensure the differences report workbook is serialized exactly once and consumers only read cached bytes.
- Add `serialize_workbook` (calls `FastExcel#read_string` once) and make `workbook_data` an attr reader.
- Call `serialize_workbook` immediately after rendering the workbook then upload/email using cached bytes.
- Update mailer to use report.workbook_data.
  
## Why:
`FastExcel` creates a temporary folder under /tmp when opened without a filename and deletes it after `read_string`. The code previously read the workbook twice (once for upload, once for the mailer), so the second read attempted to access the deleted tmp folder and raised Errno::ENOENT in production read-only containers.

Memoizing the serialised bytes:
- Prevents the double-read/delete race that caused `ENOENT`.
- Preserves using FastExcel's tmpdir (no need to attempt writing under` /app` or changing container mounts).
- Keeps upload (S3) and email attachment flows unchanged functionally.

## Ticket:
https://transformuk.atlassian.net/browse/HMRC-2697

## Risk:
**Risk level:** 🟢

**Reason for rating:**
Small, localised change with test coverage. No infra or API changes. Behaviour is preserved reports continue to be uploaded and attached, but now avoids a `race/ENOENT` in read-only containers. Tests include unit and worker, mailer assertions and use verifying doubles.


app/services/reporting/backfill_differences_reports.rb
```diff
@@ -73,6 +73,10 @@ def initialize(as_of:, workbook_data:)
         @workbook = WorkbookWrapper.new(workbook_data)
       end
 
+      def workbook_data
```
willfish: Backfill still only asserts `workbook.read_string`. The mailer now needs `workbook_data`. If this method goes missing, backfill emails raise `TypeError` and that spec still passes. Assert `workbook_data` on the report you pass to the mailer.

spec/workers/differences_report_worker_spec.rb
```diff
@@ -14,15 +14,32 @@
...
+    let(:report_date) { Time.zone.today.iso8601 }
     let(:differences) { Reporting::Differences.new }
 
     context 'when delivering email' do
       before { worker.perform }
 
       it { expect(Reporting::Differences).to have_received(:generate) }
       it { expect(ActionMailer::Base.deliveries.count).to eq(1) }
+
+      it 'serializes the workbook once and attaches those bytes' do
```
willfish: This doesn't prove we only serialise once. `generate` is stubbed, so `serialize_workbook` never runs. You only check the mailer reads the stubbed `workbook_data`. The example in `differences_spec` that calls real `read_string` once is the one that pins the ENOENT. Keep that. Drop the serialise-once claim here.

willfish (commented):
The FastExcel `read_string` deletes the tmp dir in `ensure`. Reading once in `serialize_workbook` and attaching `workbook_data` is the right fix.

Two test holes below.

willfish (approved):

willfish (approved):

PR #3784, opened 2026-09-21
feat(admin-api): add show endpoint for a single version

neilmiddleton:
## What:
This change adds `GET /uk/admin/versions/:id` to the admin API.

The action returns one version. It uses the same serializer as the index action. It returns 404 if the version does not exist. No existing behaviour changes.

## Why:
The Admin Portal restore action authorises against the `Version` class, not against the record that the version belongs to. `VersionPolicy#restore?` permits a TECHNICAL_OPERATOR. `AdminConfigurationPolicy#update?` permits only a SUPERADMIN. `AdminConfiguration` is a restorable type. A TECHNICAL_OPERATOR can therefore restore a configuration version and set a model, a prompt or a threshold to a value that the record held before. A direct edit of the same value reserves that action to SUPERADMIN.

The Access Controls and User Roles Validation page records this as gap 8.

To close the gap, the Admin Portal must read the item type of the version before it restores the version. It can then apply the policy of that record type. The admin API lists versions and restores them, but it cannot return one version, so the portal has no way to read the item type.

This is the first of two changes. The second change is in `trade-tariff-admin`. It uses this endpoint and applies the correct policy.

## Ticket:
AI-1274 (Access Controls Validated). Gap 8 in the Access Controls and User Roles Validation page:
https://transformuk.atlassian.net/wiki/spaces/HO/pages/23305715716/Access+Controls+User+Roles+Validation+-+AI+Guided+Search+Classification

## Risk:
**Risk level:** 🟠

**Reason for rating:**
The change adds a new admin API endpoint that the Admin Portal will consume. The endpoint is read-only and additive, and no existing route or response changes. The amber rating is for the new consumed endpoint only.


willfish (approved):
The show action matches restore's lookup and the index serializer. 404 is wired through the existing Sequel::RecordNotFound handler. That is what the admin restore change needs.

PR #3785, opened 2026-09-21
Stop versioning compressed notes

willfish:
## What:
- [x] Stop paper trail on compressed notes
- [x] Delete unused compressed note version rows in the UK schema
- [x] Keep the versions endpoint for leftover rows until that delete runs

## Why:
Compressed note generation writes frequent updates. Paper trail stored a full copy of each note. That made the SQL dump grow. This version history is unused.

## Ticket:
N/A

## Risk:
**Risk level:** 🟢

**Reason for rating:** Compressed note version history is unused. The change stops writing unused paper trail rows and deletes those unused rows.


neilmiddleton (approved):
Looks good. Clean way to unwind the paper_trail dependency on compressed notes, the data migration is idempotent and correctly guarded to the uk schema, and the spec changes properly replace the reliance on has_paper_trail with direct Version records.

